PR #7, opened 2017-12-08
Adds kubernetes deployment and configmap manifests.

aurcioli-handy:
Signed-off-by: Alexander Urcioli <alexurc@gmail.com>

### DIFF
--- a/scripts/kubernetes/config-map.yml
+++ b/scripts/kubernetes/config-map.yml
@@ -1,0 +1,179 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: localanalyzer-config
+  namespace: anchore
+data:
+  config.yaml: |
+    # Anchore configuration file
+
+    #
+    # anchore_data_dir: <default is $HOME/.anchore/>
+    #
+    # anchore_db_driver: 'anchore_image_db_fs'
+    # anchore_db_driver_config:
+    #   # only applicable to the 'anchore_image_db_fs' driver (default driver), path is relative to anchore_data_dir unless it is set to an abs path
+    #   db_dir: 'data/'
+    #
+    # tmpdir: '/tmp'
+    # log_file: '/var/log/anchore.log'
+    # debug_log_file: '/var/log/anchore_debug.log'
+    #
+    # image_data_store: 'data/'  # deprecated but still works - should use the anchore_db_driver_config + db_dir settings above to control location of anchore data
+    #
+    docker_conn: 'tcp://localhost:2375'
+    # docker_conn_timeout: '120'
+    #
+    # anchore_auth_conn_timeout: '5'
+    # anchore_auth_max_retries: '3'
+    # anchore_client_url: 'https://ancho.re/v1/account/users'
+    # anchore_token_url: 'https://ancho.re/oauth/token'
+    #
+    # feeds_url: 'https://ancho.re/v1/service/feeds'
+    # feeds_conn_timeout: '10'
+    # feeds_max_retries: '3'
+    #
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: anchore-engine-config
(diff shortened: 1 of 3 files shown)


nurmi:
Thank you for the contribution @aurcioli-handy! Kubernetes deployment options are a valuable addition to users, and we ( myself and @zhill ) will be able to review over the coming days. 

-Dan

aurcioli-handy:
Having some problems with larger containers (700Mb) or so. Kernel is OOMKilling the process even after removing memory limits. When that is done memory usage spikes all the way up to 15 Gigabytes (absurd).

At first I bumped the limit from 1Gi to 2Gi:
```
[427182.833623] Memory cgroup out of memory: Kill process 24876 (twistd) score 1087 or sacrifice child
[427182.838637] Killed process 24876 (twistd) total-vm:683036kB, anon-rss:84348kB, file-rss:11852kB
```

Then removed limit altogether and thats where you see the 15 Gigabyte usage spikes. 
```
[427182.833623] Memory cgroup out of memory: Kill process 24876 (twistd) score 1087 or sacrifice child
[427182.838637] Killed process 24876 (twistd) total-vm:683036kB, anon-rss:84348kB, file-rss:11852kB

```

Memory spikes:
screen shot 2017-12-09 at 4 09 10 pm

zhill:
@aurcioli-handy thanks for submitting this PR! You might want to check out [anchore_deploy](https://github.com/anchore/anchore_deploy) where we have some similar (but not identical) k8s configs. There is a config there that will deploy the analyzer separately from the core engine services to help isolate resource usage a bit if you'd like to try that as well.

But first, lets figure out what's going on with the memory issue. It definitely should not be using anywhere near 15GB of RAM. We typically recommend allocating something like 2GB if you're running everything in the same container.

I'll take a look at your configs and see if I can replicate the memory issue to see if it is directly related to dind or something in anchore-engine itself. We haven't really tested with dind, since its still not really recommended by Docker for usage other than dev. But, I certainly understand the desire to not require mounting the docker socket into the container. To that end, we're also currently working on modifying the anchore analyzer to not require any docker access at all, but that is likely to be a little ways out for testing etc.

Is the container you're using 700MB compressed or uncompressed on the disk? I'll try to use a similar sized container for testing.

I'll update once I have info on the mem usage. Again, thanks for the PR and the feedback.

aurcioli-handy:
Hi,

After changing the default storage driver from `vfs` to `overlay2` the memory issues have gone away and are now within the normal ranges (sub 4Gi)

zhill:
@aurcioli-handy, glad to see that the storage driver resolved the memory issue.

WRT this PR, I'd like to restructure where we put deployment-related stuff, so I'll move some things around and then ping you with an update on the new location and resubmit the PR with that location. The reason for this is to differentiate between scripts for the development process and deploying the system for production use. Appreciate your patience and really appreciate your help and work on this. Updates coming shortly.

aurcioli-handy:
@zhill Awesome sounds good. Also totally missed that anchore deployment repo! Hope this isn't too redundant.

zhill:
Hi @aurcioli-handy, apologies for the delay on updates for this. What I'd like to do is close this PR and have you open one against [anchore_deploy](https://github.com/anchore/anchore_deployment) instead. I have added a CONTRIBUTE.rst and the LICENSE as well as DCO so we can take contributions there cleanly. The primary differentiators between your approach and what is already in that repo are: DinD and the direct configmap in yaml instead of instructions to create the configmap from the config file.

To start with, please put the configs in the _contrib/_ directory in that repo, with a subdirectory like _kubernetes/dind/_ to differentiate it since we already have a version that uses the docker.sock instead of Dind in there. Also happy to accept PRs for the existing k8s configs in that repo if you have suggestions or improvements. Thanks!

aurcioli-handy:
Okay great will do.

PR #10, opened 2017-12-13
Add thread management

aurcioli-handy:
This code just ensures that the service threads are supervised by the parent process and restarted if they die.

### DIFF
--- a/anchore_engine/conf/config.yaml.example
+++ b/anchore_engine/conf/config.yaml.example
@@ -13,6 +13,7 @@
 # internal_ssl_verify: True
 # image_analyze_timeout_seconds: '36000'
 cleanup_images: True
+auto_restart_services: True
 #
 # docker_conn: 'unix://var/run/docker.sock'
 # docker_conn_timeout: 600
--- a/anchore_engine/configuration/localconfig.py
+++ b/anchore_engine/configuration/localconfig.py
@@ -16,6 +16,7 @@ DEFAULT_CONFIG = {
     'image_analyze_timeout_seconds': '36000',
     'cleanup_images': False,
     'internal_ssl_verify': True,
+    'auto_restart_services': True,
     'services': {},
     'credentials': {},
     'webhooks': {},
--- a/scripts/anchore-engine
+++ b/scripts/anchore-engine
@@ -7,6 +7,7 @@ import time
 import threading
 import subprocess
 import watchdog
+import anchore_engine.configuration.localconfig
 from watchdog.observers import Observer  
 from watchdog.events import RegexMatchingEventHandler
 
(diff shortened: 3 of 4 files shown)


zhill:
Hi @aurcioli-handy, this is an issue we're discussing right now, so it may take us a bit to decide what to do here. Your approach is certainly one that @nurmi and I have discussed. Another is mapping a single process to each container and running a bunch of small containers. In K8s that isn't really a problem just a pod-spec but with docker-compose it's a bit more unpleasant. Another option is to use an external manager like supervisord to handle these transitions but it has logging implications that would require work to resolve. If we do use an auto-restart it needs to be very clear an evident when a restart has occurred since it can make debugging very difficult with a cycling service process.

What specifically prompted this change for you? Were you seeing bootstrap errors in deployment on k8s or something else like OOM killer interruptions?

Would you be opposed to a requirement that anchore is executed as a set of containers rather than just one + postgres? Feedback appreciated.

aurcioli-handy:
Hey @zhill , expected as much with regards to your response. I think having each of the Anchore services as its own container probably makes the most sense. 

This code was primarily just a bandaid for the kernel OOM Killing a service every now and then. Maybe its better as an optional configuration? 

Either approach seems fine to me in the end as long as there is some kind of way to make sure these services remain available.

zhill:
@aurcioli-handy Thanks. Ok let's take the initial approach of configurable restarts in the anchore-engine script. We can get more advanced later if needed. Can you add a config value in the config.yaml at the top level (not nested in a service section) that can control whether anchore-engine script will restart killed processes. Then I'll test it and we can merge the PR if everything looks good.

Thanks again for the contribution. Really appreciate your efforts and interest. Let me know if there is anything I can help you with.

aurcioli-handy:
@zhill Sounds good, LMK if the changes are OK.  And this is a really great project! I am in the process of integrating it into our workflow here at Handy.

zhill:
I took a look at the code, and the overall approach looks good but I think you can do without the ServiceThread class definition. It looks like the primary purpose of that class is to track the thread name, but that can be set in the Thread constructor call using the 'name' parameter so the extra class can be removed to simplify the code.

Also, procedurally, please make sure you sign-off on every commit so we have a specific record after the merge.

Otherwise I think it looks good and should be able to merge.

aurcioli-handy:
I think it will be a little messier without the class abstraction due to tracking thread target and arguments and re-instantiating the thread since you can't simply restart them.

zhill:
@aurcioli-handy yep, you're right and a class is simpler than tracking other ways. Should be good to merge. Can you either rebase and squash or sign both commits? I see a signature on the 2nd commit but not the first, and we need the sign-off on all commits. Once that's ready I'm happy to merge.

aurcioli-handy:
Cool! done and done :)

zhill:
Thanks @aurcioli-handy ! All merged!

PR #24, opened 2018-01-16
Fixes #12 - Adds 'registryIds' to AWS ECR get_authorization_token call

### LINKED ISSUE #12: Cannot analyze tags from ECR repositories in another account using IAM_AUTO mode
When the engine is configured to use IAM_AUTO mode (allow_awsecr_iam_auto = True in config.yaml) and a registry is added with 'awsauto' as username, if the registry is in another account from the account the role exists in the token is not configured correctly because the registryID (account ID of the registry) is not set in the get_authorization_token() call to get credentials.

To reproduce:
* Grant access to the engine account in the ECR repository policy in AWS, read-only.
* Run anchore-engine in aws iam auto mode and add a registry in a different account from the one the engine is running in.
* Try to analyze a tag in that repository. Will fail with auth issue.

To fix:
add _registryIds=[aid]_ in the params the get_authorization_token() call for https://github.com/anchore/anchore-engine/blob/master/anchore_engine/auth/aws_ecr.py#L26

cmattoon:
Signed-off-by: Curtis Mattoon <cmattoon@cmattoon.com>

### DIFF
--- a/anchore_engine/auth/aws_ecr.py
+++ b/anchore_engine/auth/aws_ecr.py
@@ -6,14 +6,37 @@ import time
 import boto3
 import datetime
 
+
 from anchore_engine.subsys import logger
 import anchore_engine.configuration.localconfig
+from urlparse import urlparse
+
+
+def parse_registry_url(registry_url):
+    """
+    Given an AWS ECR registry URL, parses out the AWS Account ID and Region
+
+    >>> aid, region = parse_registry_url('http://12345.dkr.ecr.us-west-2.amazonaws.com')
+    ('12345', 'us-west-2')
+
+    >>> aid, region = parse_registry_url('12345.dkr.ecr.us-west-2.amazonaws.com')
+    ('12345', 'us-west-2')
+
+    Args:
+      - registry_url (str): The AWS ECR Registry URL
+
+    Returns a tuple of strings (aid, region)
+    """
+    parsed = urlparse(registry_url)
+    host = parsed.hostname if parsed.hostname is not None else parsed.path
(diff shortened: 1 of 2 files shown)


zhill:
Thanks @cmattoon. This looks good. We'll give it a quick test to confirm everything and then merge it.

zhill:
@cmattoon to make this a little safer in cases where the user enters a registry as _http://accountid.dkr.ecr.region.amazonaws.com_ instead of just _accountid.dkr.ecr.region.amazonaws.com_, please add a quick url parse and use only the host section for the split to get the account id. In the current form it doesn't matter if _aid_ gets a weird value including the url schema because it wasn't used. But it won't be valid if the user enters a schema on the registry name.

e.g.
```
host = urlparse.urlparse(registry).hostname
(aid, dkr, ecr, region, azn, com) = host.split('.')
```

Once that's in I will merge it up. Thanks again for the PR!

cmattoon:
I ended up splitting out that logic to a separate method, and ran some manual tests. I wasn't able to get the test suite to run, but the function passes the test cases. 

`000.dkr.ecr.region.amazonaws.com` results in `urlparse(url).hostname` being `None`, and the relevant part ends up in the `path` attribute instead.

zhill:
@cmattoon lgtm, will merge. Before I can do that I need you to Sign-off on that second commit. We need a sign-off on each commit. Once done I'm merging. Thanks, really appreciate your contribution!

zhill:
bump @cmattoon

cmattoon:
Hey, I missed the notification for this... should be fixed now!

zhill:
All merged. Thanks @cmattoon !

Issue #48, opened 2018-05-30
Add Java/Maven metadata support


jvz:
I was browsing through the codebase and found [this feature](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/analyzers/modules/32_java_packages.py) where `manifest.mf` files are scanned for data. Not all Java libraries add these metadata. However, any library packaged with Apache Maven tends to include a file named `pom.properties` (generally inside `META-INF/maven/groupId/artifactId/pom.properties`) which has the name and version info data. This is also generally more reliable than the manifest file due to, once again, typical build tooling usage. This can be very helpful when libraries are shaded together; the manifest files may get destroyed in the process, but all the `pom.properties` files will safely-ish combine together (other than version number conflicts, but that's also handled by shading).

zhill:
@jvz thanks for filing this. We'll definitely take a look and update this ticket with progress. Anything that helps the fidelity of the detection process is much appreciated.

jvz:
I'd love to help out where I can with this. I'm investigating integrating Anchore with Jenkins's Jenkins build, and if I can use the same tool for both Docker images as well as the Java libraries within, that would be fantastic. I'm not exactly sure how these data overlap with the CVE feeds, but that would be my goal at least.

nurmi:
Hi @jvz - that is a great observation!  There are effectively two 'sides' to the anchore process, where the first is 'gather the data' and the second is 'parse, store and use' the data for policy evaluation.  The analyzer module (32_java_packages.py) represents the first phase, where I'd suggest that we would add the pom.properties file as an additional 'metadata' element, if found.  The second part is largely implemented [here in the policy engine](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/engine/loaders.py#L590), where candidates for the name/version are discovered from the analyzer artifacts, constructed into a set of potential CPEs, and then matched against NVD data (and used as content output when java package information is requested for an image).  

I thought I'd provide a bit more context about the two 'sides' of the process when considering adding more analysis capabilities for java packages!

jvz:
Great, that's exactly what I was wondering about! Are there any sample unit or integration tests to go off of? Or perhaps a desired way of doing so if there aren't any already.

zhill:
[...]

PR #60, opened 2018-06-07
Add support for scanning Jenkins plugins

jvz:
This adds the file extensions ".hpi" and ".jpi" to the list of recognized Java
library filenames.

This adds Jenkins plugin file extensions to the list of Java file extensions to scan.

**Special notes**:

Related to #48 somewhat.

### DIFF
--- a/anchore_engine/analyzers/modules/32_java_packages.py
+++ b/anchore_engine/analyzers/modules/32_java_packages.py
@@ -22,6 +22,8 @@ import anchore_engine.analyzers.utils
 
 analyzer_name = "package_list"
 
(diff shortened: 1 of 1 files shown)


jvz:
I'm almost tempted to add `.rar` files since those are technically another Java EE/Jakarta EE file extension, but that could be confused with real rar files (the Java EE rar file is still a zip). Plus, I've literally never seen anyone use them in the wild before. Realistically, you'd only find them inside a `.ear` file or an exploded ear directory which makes it a little more complicated to find properly.

zhill:
@jvz Thanks, much appreciated! We'll give this a review.

Issue #81, opened 2018-08-13
low performance on image evaluation


armstrongli:
**Is this a request for help?**:
* No
---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):
* bug report

**Version of Anchore Engine and Anchore CLI if applicable**:
All the versions of anchore
 
**What happened**:
The kubernetes image policy is very low performance

**What did you expect to happen**:
The evaluation has been done by calling the eval, and a evaluation result is stored in table *policy_eval* . The evaluation should be extremely fast as it is only a one-row query. 

**Any relevant log output from /var/log/anchore**:

**What docker images are you using**:

**How to reproduce the issue**:

**Anything else we need to know**:
The call stack of the kubernetes webhook is:
* anchore_engine/services/kubernetes_webhook/api/controllers/default_controller.py
   * `imagepolicywebhook(bodycontent)`
   * result = catalog.get_eval_latest(request_user_auth, tag=fulltag, imageDigest=imageDigest, policyId=requestPolicyId)
* anchore_engine/clients/catalog.py
   * get_eval_latest(userId, policyId=None, imageDigest=None, tag=None, evalId=None)
   * eval_records = get_evals(userId, policyId=policyId, imageDigest=imageDigest, tag=tag, evalId=evalId, newest_only=True)
   * get_evals(userId, policyId=None, imageDigest=None, tag=None, evalId=None, newest_only=False)
   * ret = http.anchy_get(url, params=params, auth=auth, headers=headers, verify=localconfig['internal_ssl_verify'])
* anchore_engine/services/catalog/api/controllers/policy_evaluations.py
   * get_evals(policyId=None, imageDigest=None, tag=None, evalId=None, newest_only=False)
   * **list_evals_impl(session, userId=user_id, policyId=policyId, imageDigest=imageDigest, tag=tag, evalId=evalId)** . Down here, all the calls are going through the same logic as the `newest_only` option is dropped and not used anywhere.
   * list_evals_impl(dbsession, userId, policyId=None, imageDigest=None, tag=None, evalId=None, newest_only=False)
   * rc = catalog_impl.perform_policy_evaluation(userId, imageDigest, dbsession, evaltag=evaltag, policyId=policyId)

The whole stack show that all the requests are hitting the policy engine and does the whole evaluation every time when there's a evaluation call.

armstrongli:
PR for this issue on performance fix: https://github.com/anchore/anchore-engine/pull/82

[...]

PR #82, opened 2018-08-13
get the latest evaluation result from database if the newest_only is …

armstrongli:
…true

Fix one issue on kubernetes webhook

* https://github.com/anchore/anchore-engine/issues/81

**Special notes**:

The logic change part is to check the existing evaluation result first, then do evaluation if there're no evaluation records in the database.

### DIFF
--- a/anchore_engine/services/catalog/api/controllers/policy_evaluations.py
+++ b/anchore_engine/services/catalog/api/controllers/policy_evaluations.py
@@ -27,11 +27,8 @@ def get_evals(policyId=None, imageDigest=None, tag=None, evalId=None, newest_onl
         user_id = request_inputs['userId']
 
         with db.session_scope() as session:
-            if newest_only:
-                evals = list_evals_impl(session, userId=user_id, policyId=policyId, imageDigest=imageDigest, tag=tag,
-                                        evalId=evalId)
-            else:
-                evals = list_evals_impl(session, userId=user_id, policyId=policyId, imageDigest=imageDigest, tag=tag, evalId=evalId)
+            evals = list_evals_impl(session, userId=user_id, policyId=policyId, imageDigest=imageDigest, tag=tag,
+                                    evalId=evalId, newest_only=newest_only)
 
             if not evals:
                 httpcode = 404
@@ -141,6 +138,10 @@ def list_evals_impl(dbsession, userId, policyId=None, imageDigest=None, tag=None
     if evalId is not None:
         dbfilter['evalId'] = evalId
 
+    if newest_only:
+        records = db_policyeval.tsget_byfilter(userId, session=dbsession, **dbfilter)
+        if len(records) > 0:
+            return records
 
     # perform an interactive eval to get/install the latest
     try:


zhill:
@armstrongli Thanks for digging in and submitting this PR, we definitely appreciate the work and contribution, and we'll definitely work to solve the performance issue with you.

However, in this PR it looks like you're significantly changing the semantics of the 'newest_only' flag in the internal catalog API from it's original intent and in ways that have impacts to other parts of the API. The original intent was to control the number of results returned, not short-circuit the evaluation. I understand the desire to achieve that short-circuit, but implementing it there has implications on the rest of the system and the overall accuracy of the policy evaluation results.

Specifically, with the proposed change, any user call to 'anchore-cli evaluate check ' would only return the first policy evaluation ever done unless a user also has the policy_evaluation subscription active for the tag, and then only for the newest image mapped to that tag, so untagged images would never have their policy evaluations updated to reflect new CVEs, policy changes, or other external updates that impact the policy result.

There is definitely a bug in the 'newest_only' handling, and a fix for that will reduce some db traffic, but this PR is making a significant semantic change to how the engine behaves, so as-is I can't accept it, but I am happy to merge a PR for fixing the newest_only bug that does not change the semantics of the call:

Fix the invocation to pass the param -- https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/catalog/api/controllers/policy_evaluations.py#L31

Fix the lookup (add a 'limit 1' on the query itself in the db_policyeval.tsget_byfilter via a new parameter like 'limit=None' in function declaration) -- https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/catalog/api/controllers/policy_evaluations.py#L165

I definitely want to help fix your core issue: slow k8s webhook performance. Probably accomplished by introducing a proper cache with things like ttl control in the handler or as an external client, and I'm happy to follow up on this with you if you're interested in tackling it or helping. I'm going to be working on something here in the next few days to add support for Validating and Mutating webhooks from k8s as well.

armstrongli:
@zhill 
Agree with you that this PR is changing the meaning of the `newest_only` flag. The flag should be a flag to change the set of  queries from anchore eval list. 

>  Probably accomplished by introducing a proper cache with things like ttl control in the handler or as an external client,

I think it's not the best option to set up a external cache for the evaluation result as the cache has no idea what is happening in the backend, and don't know when to refresh the cache. By anchore engine, the valuation result will never change if 
1. there're no CVE changes
2. there're no image updates
3. there're no policy updates

And anchore engine itself has the full context on the image and know when to re-evaluate the image and when to refresh the result in the table. 
There're subscriptions set up in anchore engine - vuln_update, analysis_update, tag_update, etc. - and it is available for anchore engine to detect which images should be re-evaluate and when to update the result. 
A cache server is an extremely thin layer without any context on the backend service. and it is better to have a fast sync between cache layer and service layer.

zhill:
@armstrongli fundamentally, your description of the update condition is correct (except that any feed update can cause a policy eval change, not just CVEs), but that is not currently implemented in the system and those existing subscriptions only cover the currently tagged image, which is not the same scope as the policy evaluation code, which can evaluate any image based on digest whether currently tagged or not. What you're describing is definitely what we'll get to, but only at that point is it safe for the rest of the system to change the semantics of the _newest_only_ parameter. Before those other changes are made this change can lead to very stale evaluations that may never get updated in some cases.

The most direct solution to your issue, performance in the webhook, can be accomplished with a very simple cache that just remembers the (image digest, tag, bundle id, result) tuple for say 30 or 60 seconds and reduces hits to the policy engine. That should get you over the performance issue while we work to implement the more sophisticated update logic you described, which is certainly the better long-term solution and has been in our plans for a while.

I've also run into this with other webhook handlers (ValidatingWebhooks for anchore in other projects), so I'll prioritize the full state check work for the next release and will keep #82 updated. Once that is in place the newest_only flag shouldn't be necessary as the api call for getting a policy evaluation should work as efficiently as possible in any invocation. If you're interested in contributing this work that would be great and we can discuss where/how it should work (check for feed sync since last eval, bundle update, if any has changed, re-run the evaluation, else return the last eval).

If you want to fix the newest_only bug as I described in my initial response, or file a new PR to implement the design you and I discussed around checking image, policy, and feed sync state, I'm happy to review those as well.

zhill:
This should be no longer an issue with the interactive evaluation and changes in 0.3.1 of anchore-engine to support policy evalutions only when either the image, feeds, or policy has changed. See: eec019d6533e3bc7e51ddd91993094e0c4be127e (there are a couple of follow-up bug fixes, but that commit is the primary feature).

PR #98, opened 2018-10-04
Prevent zombie processes/containers

### LINKED ISSUE #97: anchore-engine in Docker Swarm gives zombie containers
I've deployed `anchore-engine` to a Docker Swarm cluster using the provided `docker-compose.yml` in the `master` branch of this repo. This deploys a stack with 2 containers as expected (database and engine).

When redeploying the stack Docker Swarm will first issue a SIGTERM to the running engine container, wait for it to stop, and when stopped start a new container. If the container does not stop within a specific time (10 seconds), it will issue a SIGKILL.

Because `shell` form of `CMD` is used, the `anchore-engine` process is started from `/bin/sh -c`. Normally this is not a problem, however `anchore-engine` spawns subprocesses, and because of the main process being in a shell, the SIGTERM and SIGKILL signals are not properly propagated into the child processes. This causes `anchore-engine` zombie containers to be left on the Docker host, which can only be removed by restarting the Docker daemon.

The `anchore-engine` Dockerfiles should be fixed to:
- use `exec` form instead of `shell` form
- add an `init` like process that properly propagates the system SIGTERM and SIGKILL signals (e.g. `tini`)

For more background info see:
https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/
and:
https://docs.docker.com/engine/reference/builder/#cmd

## Steps to Reproduce:
1. use docker-compose installation instructions from README
2. instead of `docker-compose up` use `docker stack deploy`
3. use `docker service scale` to first set `anchore-engine` to 0 replicas
4. and then scale `anchore-engine` back to 1 replica
5. with `docker service ls` it is observed that `anchore-engine` now has `2/1` in the replica column

I will make a PR with suggested changes.

bjwschaap:
Fixes #97 

- rewrite CMD to `exec` form
- add `tini`

Signed-off-by: Bastiaan Schaap <bastiaan.schaap@siteminds.nl>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,22 @@ FROM centos:7
 ENV LANG=en_US.UTF-8
 EXPOSE 8228 8338 8087 8082
 
-RUN yum -y update && yum -y install epel-release && yum -y install skopeo python-pip dpkg gcc python-devel openssl-devel psmisc && yum clean all -y
+RUN yum -y update && yum -y install epel-release && yum -y install skopeo python-pip dpkg gcc python-devel openssl-devel psmisc && yum clean all -y && rm -rf /var/cache/yum
+
+ENV TINI_VERSION=v0.18.0
+RUN cd /tmp && \
+  gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 && \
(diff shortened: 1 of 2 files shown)


zhill:
Closing this PR as it's no longer in line with how the containers are run or their entrypoint. If the zombie issue persists on the latest codebase please update #97 and we'll look into it further and ensure the problem is resolved. thanks!

PR #101, opened 2018-10-11
fix the issue that the lays of images doesn't have full image info.

armstrongli:
some of the history has body like:
```json
    {
      "v1Compatibility":
        "{\"id\":\"38c5916c333...\",\"parent\":\"bac7f0439852...\"}"
    }
```

The analyze function break directly.


nurmi:
Hi @armstrongli - thank you for yet another improvement contribution, we really appreciate it!

We'll need for you to re-commit after signing off on the commit (the process hasn't changed since you're last contributions, although we have added the DCO check module that should be showing in this thread as part of the PR to help act as a reminder)

Once signed off, the PR looks good to me.

armstrongli:
@nurmi sign-off added. PTAL

nurmi:
Thanks @armstrongli !

PR #102, opened 2018-10-11
Improve Java properties and manifest file parsing support

jvz:
This is a small cleanup for the Java analyzer. In preparation for adding
better manifest file support, I've moved some Java-related utility code
into its own file along with unit tests.

This now includes said manifest parsing improvements. Later on, I can
add a Jenkins plugin artifact type which will use additional manifest
data from the jar.

@reviewbybees


nurmi:
thanks @jvz !  Really nice cleanup and organization of the java metadata parsing implementation - very much appreciated.

PR #137, opened 2019-02-04
Make use of ANCHORE_LOG_LEVEL env variable

jbouse:

This PR includes the use of the ANCHORE_LOG_LEVEL env variable that is already defined in the Dockefile but not used within the /config/config.yaml file.

**Which issue this PR fixes** 

Issue not filed on problem


nurmi:
Good catch @jbouse - thank you for the contribution!

Issue #138, opened 2019-02-04
Feature: Add ability to enable external_tls along with external_port for services


jbouse:
**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): FEATURE REQUEST

**Version of Anchore Engine and Anchore CLI if applicable**:
0.3.2
 
**What happened**:
The /config/config.yaml already supports passing ANCHORE_EXTERNAL_PORT to set the 'external_port' for a given service but if that external port should be TLS enabled it has `external_tls: false` as default. As documented by #49 this should be set to `external_tls: true` in this case. Proposed solution is to add an ANCHORE_EXTERNAL_TLS env variable, defaulted to false, that can be used to set the `external_tls` property in the config.yaml.

**What did you expect to happen**:
To maintain current configuration set `ANCHORE_EXTERNAL_TLS=false` in Dockerfile  and changing the conf/default_config.yaml to use `external_tls: ${ANCHORE_EXTERNAL_TLS}` which will maintain the `external_tls: false` setting but if it needs to be enabled then ANCHORE_EXTERNAL_TLS can be set to true resulting in a parsed configuration of `external_tls: true`

**What docker images are you using**:
anchore/anchore-engine:v0.3.2

PR #139, opened 2019-02-04
Adds ANCHORE_EXTERNAL_TLS env variable

jbouse:
* To compliment ANCHORE_EXTERNAL_PORT which sets the external_port for
services, add ANCHORE_EXTERNAL_TLS to set external_tls as well
* Defaults to 'false' maintaining backwards compatibility with what is
available now

Adds support for ANCHORE_EXTERNAL_TLS to compliment the ANCHORE_EXTERNAL_PORT env variable to enable setting the `external_tls` setting for services.

**Which issue this PR fixes**

fixes: #138


nurmi:
thank you @jbouse !  LGTM, and very much appreciated as this will help as we transition toward using this new 'default config in the container, configure using ENV variables' method for deploying engine moving forward.

PR #156, opened 2019-03-12
Add optional policy parameter for vulnerabilities older than N days

### LINKED ISSUE #149: Feature: Add optional policy parameter for vulnerabilities older than N days
It would be great if I could add policy rules which WARN if vulnerabilities are found of a certain type and BLOCK if they have been discovered for longer than a certain number of days. Ideally for our use case this should be driven off of the actual CVE date or fix date but as long as updates run frequently the created date (which appears to be written to the db already) would work for now.

For example,

Trigger Name | Description | Parameter | Description | Example
-- | -- | -- | -- | --
package | Triggers if a found vulnerability in an image meets the comparison criteria. | days_past_creation | Matches when the vulnerability was reported more than the provided value of days ago. | 30

Then end users could define policies similar to...

```
{      
  "name": "CustomPolicy",
  "version": "1_0",
  "comment": "Policy for date based vulnerabilities",
  "id": "policy1", 
  "rules": [
      {
        "action": "STOP",
        "gate": "vulnerabilities",
        "id": "stophigh",
        "params": [
          { "name": "package_type", "value": "all" },
          { "name": "severity_comparison", "value": ">=" },
          { "name": "severity", "value": "high" },
          { "name": "days_past_creation", "value": 30}
        ],
        "trigger": "package"
      },
      {
        "action": "WARN",
        "gate": "vulnerabilities",
        "id": "warnhigh",
        "params": [
          { "name": "package_type", "value": "all" },
          { "name": "severity_comparison", "value": ">=" },
          { "name": "severity", "value": "high" }
        ],
        "trigger": "package"
      }
  ]
}
```

Is this something that makes sense as a feature for this project?
Is there a better approach to implementing a feature like this that is preferred?

emaildanwilson:
fixes #149
cc @nurmi 

Let me know how this looks, and what other changes would need to be made for it to be included. How are tests handled for policy filters like this? Does the cli need to be updated to include the creation date in vulnerability output?

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
@@ -39,7 +39,7 @@ class VulnerabilityMatchTrigger(BaseTrigger):
     severity = EnumStringParameter(name='severity', example_str='high', description='Severity to compare against.', enum_values=SEVERITY_ORDERING, is_required=True, sort_order=3)
     fix_available = BooleanStringParameter(name='fix_available', example_str='true', description='If present, the fix availability for the vulnerability record must match the value of this parameter.', is_required=False, sort_order=4)
     vendor_only = BooleanStringParameter(name='vendor_only', example_str='true', description='If True, an available fix for this CVE must not be explicitly marked as wont be addressed by the vendor', is_required=False, sort_order=5)
-    max_days_since_sync = IntegerStringParameter(name='max_days_since_sync', example_str='7', description='If provided, this CVE must be older than the days provided to trigger.', is_required=False, sort_order=6)
+    max_days_since_creation = IntegerStringParameter(name='max_days_since_creation', example_str='7', description='If provided, this CVE must be older than the days provided to trigger.', is_required=False, sort_order=6)
 
     def evaluate(self, image_obj, context):
         is_fix_available = self.fix_available.value()
@@ -47,8 +47,8 @@ class VulnerabilityMatchTrigger(BaseTrigger):
         comparison_idx = SEVERITY_ORDERING.index(self.severity.value().lower())
         comparison_fn = self.SEVERITY_COMPARISONS.get(self.severity_comparison.value())
         timeallowed = time.time()
-        if self.max_days_since_sync.value() is not None:
-            timeallowed -= int(int(self.max_days_since_sync.value()) * 86400)
+        if self.max_days_since_creation.value() is not None:
+            timeallowed -= int(int(self.max_days_since_creation.value()) * 86400)
         if not comparison_fn:
             raise KeyError(self.severity_comparison)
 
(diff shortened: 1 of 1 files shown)


zhill:
Thanks @emaildanwilson ! We'll take a look at comment here!

emaildanwilson:
Any chance someone could take a look this week?

nurmi:
hello!

We’re finishing up some feature work, and you’ll see some flurry of work in the next couple of weeks as we gear up for the next oss engine release - your pr is on the list for adressing for that next release - stay tuned and appreciate your patience!

zhill:
I think the primary question is the semantics you're trying to achieve. The created_at timestamp on vulnerability records in anchore is when the db entry was created, not when the vulnerability was created in in the upstream source. So, for example, none of the created_at timestamps will be older than the anchore deployment date itself. Is that what you're expecting?

emaildanwilson:
Yes, given that some of the sources do not contain a discovery date it seems like it's the best we can do right now. Aside from the caveat of the installation being brand new or the local scan database getting deleted/recreated, the created_at date should match the discovery date within the polling period. Using the created_at date for feeds where the discovery date is not available might create even more confusion. I'm open to ideas if there are ways we could make this better.

zhill:
Ack, just wanted to make sure the semantics were understood and acceptable. I tend to agree that there isn't a cleaner way to do it in this case given the diversity of input data.

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -39,12 +39,16 @@ class VulnerabilityMatchTrigger(BaseTrigger):
     severity = EnumStringParameter(name='severity', example_str='high', description='Severity to compare against.', enum_values=SEVERITY_ORDERING, is_required=True, sort_order=3)
     fix_available = BooleanStringParameter(name='fix_available', example_str='true', description='If present, the fix availability for the vulnerability record must match the value of this parameter.', is_required=False, sort_order=4)
     vendor_only = BooleanStringParameter(name='vendor_only', example_str='true', description='If True, an available fix for this CVE must not be explicitly marked as wont be addressed by the vendor', is_required=False, sort_order=5)
+    days_past_creation = IntegerStringParameter(name='days_past_creation', example_str='7', description='If provided, this CVE must be older than the days provided to trigger.', is_required=False, sort_order=6)
```
zhill: For consistency with our other gates and trigger parameter names, can you change the name field (the var name is fine) in this to 'max_days_since_creation'. We use 'max_days_since_sync' in our feeds timeliness check which is similar semantically, so I want to keep the naming similar.

emaildanwilson: Thank you for the review. I updated this to match and went ahead and updated the variable names too so everything matches.

zhill: I think there was some confusion on the name, I wasn't proposing the exact same name, but something like 'max_days_since_creation' for this particular parameter, which follows a similar naming scheme to the other feed sync param, but is clear in this vuln context. Apologies for that lack of clarity. You can just push a new commit and I'll merge-squash it down. That makes it easier to review here as well because github will notice the change and map it to the comment. Overall, though, this is great work and super useful stuff. Thanks!

emaildanwilson: You were clear and I find/replaced the wrong thing. fix coming soon.

zhill: Thanks! looks good.

zhill:
Thanks! Much appreciated, this is a super-useful feature.

PR #171, opened 2019-04-11
Improve error message from parse_dockerimage_string

0xDEC0DE:
- Throw a `ValueError` rather than a bare `Exception` when a bad
  character is found

- Emit the problem-causing characters as part of the error message


nurmi:
thank you for the contribution @0xDEC0DE - nice improvement!

PR #225, opened 2019-07-02
Support passing external_id in the secret_access_key for cross-accoun…

naman:
Support passing external_id in the secret_access_key for cross-account ECR access

The PR allows passing the external_id which allows a cross-account role to be assumed in case the corresponding role's account requires an external_id to be sent as a request parameter (in order to allow the role to be assumed). This is achieved by passing the secret_access_key / registry_pass variable in this format (separated by a semi-colon) "role_arn;external_id".

**Special notes**:

### DIFF
--- a/anchore_engine/auth/aws_ecr.py
+++ b/anchore_engine/auth/aws_ecr.py
@@ -54,7 +54,11 @@ def refresh_ecr_credentials(registry, access_key_id, secret_access_key):
         elif access_key_id == '_iam_role':
             try:
                 sts = boto3.client('sts')
-                session = sts.assume_role(RoleArn=secret_access_key, RoleSessionName=str(int(time.time())))
+                role_arn, external_id = secret_access_key.split(";") # pass = "role_arn;external_id"
+                if external_id:
+                    session = sts.assume_role(RoleArn=role_arn, RoleSessionName=str(int(time.time())), ExternalId=external_id)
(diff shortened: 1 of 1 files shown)


zhill:
@naman thanks for the PR. This is a useful enhancement.

anchore_engine/auth/aws_ecr.py
```diff
@@ -54,7 +54,11 @@ def refresh_ecr_credentials(registry, access_key_id, secret_access_key):
         elif access_key_id == '_iam_role':
             try:
                 sts = boto3.client('sts')
-                session = sts.assume_role(RoleArn=secret_access_key, RoleSessionName=str(int(time.time())))
+                role_arn, external_id = secret_access_key.split(";") # pass = "role_arn;external_id"
```
zhill: This will throw a ValueError if the secret_access_key value does not contain a ';' due to the tuple assignment. If you assign the split output to a single (list) value then test the length of the list for existence of the 2nd element then it should be safer.

e.g. 
```
components = secret_access_key.split(';')
role_arn = components[0]
if len(components) > 1:
  external_id = components[1]
```

naman: Thanks for the input. Will make these changes asap.

zhill:
Thanks @naman !

PR #274, opened 2019-09-19
Improve version comparison for dpkg and rpm

### LINKED ISSUE #265: Presence of epoch in image package and fix versions can cause the vulnerability to not match correctly
Issue was found against an ubuntu 14.04 image with the package `file` version `1:5.14-2ubuntu3`. Expected to find a match to CVE-2014-9620 but anchore engine does not handle version comparison between epoch versions correctly

Engine DB Version: 0.0.11
Engine Code Version: 0.5.1

nightfurys:
For dpkg and rpm
    - compare package version with the epoch included fixed-in version
    - compare epochs only if all versions being compared contain an epoch, otherwise ignore it

Fixes #265

### DIFF
--- a/anchore_engine/util/rpm.py
+++ b/anchore_engine/util/rpm.py
@@ -49,43 +49,72 @@ def split_rpm_filename(rpm_filename):
     return name, version, release, epoch, arch
 
 
-def compare_versions(pkg_a, ver_a, pkg_b, ver_b):
+def split_version(version):
+    """
+    Splits version string into a tuple (epoch, version, release). Returns a null for epoch if the version string does not contain epoch.
+    Use this function for splitting versions already processed by anchore engine. For parsing info from rpm file name, use split_rpm_filename()
+
+    '2.27-34.base.el7' -> (null, '2.27', '34.base.el7')
+    '1:2.27-34.base.el7' -> ('1', '2.27', '34.base.el7')
+
+    :param version: Version string with or without an epoch prefix
+    :return: tuple (epoch, version, release)
+    """
+    ver_comp = version.rsplit('-', 1)
+
+    if len(ver_comp) > 1:
+        release = ver_comp.pop()
+    else:
+        release = None
+
+    epoch_comp = ver_comp[0].split(':', 1) if ver_comp else []
+
+    if len(epoch_comp) == 1:
+        epoch = None
+        version = epoch_comp[0]
+    elif len(epoch_comp) > 1:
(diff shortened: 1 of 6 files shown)


anchore_engine/util/rpm.py
```diff
@@ -49,43 +49,72 @@ def split_rpm_filename(rpm_filename):
     return name, version, release, epoch, arch
 
 
-def compare_versions(pkg_a, ver_a, pkg_b, ver_b):
+def split_version(version):
```
zhill: can we tweak the name to 'split_fullversion'. in these parsing packages 'version' does have a specific meaning and epoch is not a part of the version in that definition. Basically, I'd like to make it very clear where we deviate from standard rpm version behavior for the special cases we see comparing actual real-world data.

test/unit/anchore_engine/util/test_rpm.py
```diff
@@ -0,0 +1,38 @@
...
+from anchore_engine.util.rpm import compare_versions
+
+
+class TestRpmVersionHandling(unittest.TestCase):
+    """
+    Tests for version comparisons of the rpm version parser and compare code.
+
+    """
+
+    def test_version_comparision(self):
```
zhill: typo

test/unit/anchore_engine/util/test_rpm.py
```diff
@@ -0,0 +1,38 @@
...
+        test_epoch = [
+            ('1:0', '0:1', 1),
+            ('1:0', '1', -1),
+            ('1:2', '1', 1),
+            ('2:4.19.01-1.el7_5', '4.19.1-1.el7_5', 0),
```
zhill: Please add comments here indicating that it is expected these are non-rpm-spec comparisons. An external viewer would say this is incorrect behavior, so as a note to ourselves indicate how/why these cases are correct.

test/unit/anchore_engine/util/test_rpm.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class TestRpmVersionHandling(unittest.TestCase):
+    """
+    Tests for version comparisons of the rpm version parser and compare code.
+
+    """
+
+    def test_version_comparision(self):
+        test_epoch = [
```
zhill: I notice we don't have any tilde test cases. Since you're in here, please add a test for that case (e.g. 1.2.3-el7_5~snapshot1 < 1.2.3-3-el7_5). as well as for the dpkg

PR #326, opened 2020-01-20
exception fixes

alfredodeza:
**What this PR does / why we need it**: Removes unused `err` variables from try/except blocks, and removes one instance of a redundant try/except that is re-raising exceptions.

**Special notes**: There are a few of these try/except blocks around, didn't want to go through all of them. I believe the blocks _are_ redundant, but maybe it needs a wider discussion. Happy to clean them all up if this goes through.

### DIFF
--- a/anchore_manager/cli/analyzers.py
+++ b/anchore_manager/cli/analyzers.py
@@ -203,6 +203,7 @@ def exec(docker_archive, anchore_archive, digest, parent_digest, image_id, tag, 
             try:
                 image_content_data[content_type] = anchore_engine.common.helpers.extract_analyzer_content(image_data, content_type, manifest=input_manifest_data)
             except Exception:
+                logger.exception("Unable to determine content_type, will fallback to {}")
                 image_content_data[content_type] = {}
 
         anchore_engine.common.helpers.update_image_record_with_analysis_data(image_record, image_data)
--- a/anchore_manager/cli/analyzers.py
+++ b/anchore_manager/cli/analyzers.py
@@ -134,10 +134,7 @@ def exec(docker_archive, anchore_archive, digest, parent_digest, image_id, tag, 
                 userId = 'admin'
 
             if created_at:
-                try:
-                    if int(created_at) < 0 or int(created_at) > now+1:
-                        raise Exception()
-                except Exception as err:
+                if int(created_at) < 0 or int(created_at) > now+1:
                     raise ValueError("created_at must by a unix timestamp between 0 and now ({})".format(now))
             else:
                 created_at = now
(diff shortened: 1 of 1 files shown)


anchore_manager/cli/analyzers.py
```diff
@@ -137,7 +137,7 @@ def exec(docker_archive, anchore_archive, digest, parent_digest, image_id, tag,
                 try:
                     if int(created_at) < 0 or int(created_at) > now+1:
                         raise Exception()
```
zhill: Shouldn't this be a ValueError directly? I think the outer wrapper can be removed as it just obscures possible other issues.

anchore_manager/cli/analyzers.py
```diff
@@ -193,27 +193,23 @@ def exec(docker_archive, anchore_archive, digest, parent_digest, image_id, tag,
...
-            for content_type in anchore_engine.common.image_content_types + anchore_engine.common.image_metadata_types:
-                try:
-                    image_content_data[content_type] = anchore_engine.common.helpers.extract_analyzer_content(image_data, content_type, manifest=input_manifest_data)
-                except:
-                    image_content_data[content_type] = {}
+        image_content_data = {}
+        for content_type in anchore_engine.common.image_content_types + anchore_engine.common.image_metadata_types:
+            try:
+                image_content_data[content_type] = anchore_engine.common.helpers.extract_analyzer_content(image_data, content_type, manifest=input_manifest_data)
+            except Exception:
```
zhill: I'd like a warning or something indicating this path was taken as it is an exception path that already wasn't handled well

anchore_manager/cli/analyzers.py
alfredodeza: I think it should, but I am being overly cautious here :) The thing that made me pause is that if the outer try/except is removed, it wouldn't catch a `ValueError` if `int(created_at)` fails. 

If you think there might not be a chance for that to happen, I am OK with further cleaning up this

anchore_manager/cli/analyzers.py
alfredodeza: The `logger.exception` will now capture whatever fails with the content analyzer

zhill: lgtm

anchore_manager/cli/analyzers.py
zhill: yes, the exception is the same either way, just a question of clear message to the user. Ideally, it would convert to an int once, handle that error case, then do the range check with a different msg, but as you have it is ok since its only better than what it was. I think we can cleanup the input checks here a lot for testability & maintainability, but that's another topic.

PR #327, opened 2020-01-23
Update README.md

ykocaman:
I added simplier way of downloading docker-compose.yaml file

**What this PR does / why we need it**: making simplier installing steps.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #: No

**Special notes**: None

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,7 @@ To quickly bring up an installation of Anchore Engine on a system with docker (a
 mkdir ~/aevolume
 cd ~/aevolume
 
-docker pull docker.io/anchore/anchore-engine:latest
-docker create --name ae docker.io/anchore/anchore-engine:latest
-docker cp ae:/docker-compose.yaml ~/aevolume/docker-compose.yaml
-docker rm ae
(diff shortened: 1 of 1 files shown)


zhill:
Thanks for the PR, but right now I want to keep the instructions using the container. Specifically, it ensures that the compose you use matches the container it was built for. Master will drift from what was in the previous release so what you pull from GH may not work against the released containers. We may change it to use the maintenance branches (e.g. 0.6) instead of master for pulling from GH, but even those drift over time and we want to ensure that users are getting the compose with the right config for the containers they pull.

I'm open to using the release tagged commits (https://github.com/anchore/anchore-engine/blob/v0.6.0/docker-compose.yaml) but that doesn't provide well for a 'latest' version either. Any thoughts, feedback certainly appreciated on how to make the process simpler than the current method but also reliable. One alternative is to update README with the version for each release using tags...

ykocaman:
Thanks for information. I'm closing PR.

PR #341, opened 2020-02-20
allow multiple arguments in bootstrap log functions

### LINKED ISSUE #339: bootstrap logging functions do not allow multiple args for string formatting
**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): BUG REPORT

**What happened**: The logging functions in `anchore_engine/subsys/logger.py` do not like extra arguments passed in for safe formatting. This breaks the logging:

```python
 logger.warning('message: %s, %s', 'arg 1', 'arg 2')
```

**What did you expect to happen**: Multiple arguments should be allowed for safer string formatting, instead of forcing the formatting beforehand:

```
logger.warning('message: %s, %s' % ('arg 1', 'arg 2'))
```

The fix is to catch multiple args and attempt (but not break) logging:

```diff
diff --git a/anchore_engine/subsys/logger.py b/anchore_engine/subsys/logger.py
index 848a6fc..90186b8 100644
--- a/anchore_engine/subsys/logger.py
+++ b/anchore_engine/subsys/logger.py
@@ -139,22 +139,42 @@ def spew(msg_string):


 @bootstrap_logger_intercept(logging.DEBUG)
-def debug(msg_string):
+def debug(msg_string, *args):
+    if len(str(args)):
+        try:
+            msg_string = msg_string % args
+        except TypeError:
+            exception('unable to produce log record: %s' % msg_string)
     return (_msg(msg_string, msg_log_level='DEBUG'))
```

alfredodeza:
**What this PR does / why we need it**: Fixes issue #339 by creating a helper to safely format log entries. This prevents breakage when formatting does not work and is how the logging module proceeds to prevent issues.

Additionally, it fixes a problem with the swagger API test which was using no longer valid keyword arguments (see https://github.com/zalando/connexion/issues/1161) . This problem would cause the test to fail but it would not get caught by tox/pytest, I think because of the `fork()`.  It fixes that by using the test facilities that Flask offers and is used by the `connexion` team.

There is room for more exhaustive testing, but these changes gets us to the same validation and more clarity when it fails.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #339

**Special notes**:

### DIFF
--- a/test/unit/anchore_engine/services/test_api_specs.py
+++ b/test/unit/anchore_engine/services/test_api_specs.py
@@ -24,72 +24,33 @@ service_swaggers = [
     'anchore_engine/services/kubernetes_webhook/swagger/swagger.yaml'
 ]
 
-def test_api_service():
+@pytest.mark.parametrize('service', service_swaggers)
+def test_api_service(service):
     """
-    Creates a mocked interface for each specified swagger spec and runs a server with a forked process to ensure swagger validates fully.
+    Creates a mocked interface for each specified swagger spec and creates
+    a server to ensure swagger validates fully.
 
-    :return:
+    If invalid specs are detected the spec will raise `InvalidSpecification`.
+
+    Further enhancement of this test is to make actual requests to the Apps
+    generated.
     """
 
     port = 8081
+    name = service.rsplit('/', 3)[2]
+    resolver = MockResolver(mock_all='all')
+    api_extra_args = {'resolver': resolver}
 
-    for swagger in service_swaggers:
-        pid = os.fork()
-        if not pid:
-            name = swagger.rsplit('/', 3)[2]
-            logger.info(('Starting server for: {} at: {}'.format(name, swagger)))
-            resolver = MockResolver(mock_all='all')
-            api_extra_args = {'resolver': resolver}
+    options = {"serve_spec": False, "swagger_ui": False}
+    app = connexion.FlaskApp(name, options=options)
 
-            app = connexion.FlaskApp(name,
-                                     swagger_json=False,
-                                     swagger_ui=False)
+    app.add_api(service,
+                resolver_error=True,
+                validate_responses=True,
+                strict_validation=True,
+                **api_extra_args)
 
-            app.add_api(swagger,
-                        resolver_error=True,
-                        validate_responses=True,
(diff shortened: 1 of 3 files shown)


test/unit/anchore_engine/services/test_api_specs.py
```diff
@@ -24,72 +24,33 @@
...
+    options = {"serve_spec": False, "swagger_ui": False}
+    app = connexion.FlaskApp(name, options=options)
+
+    app.add_api(service,
+                resolver_error=True,
+                validate_responses=True,
+                strict_validation=True,
+                **api_extra_args)
+
+    client = app.app.test_client()
```
zhill: Does this fully test the spec itself? The objective of this test was to identify problems in the swagger spec (beyond just syntax linting) that the startup of the server identifies.

Specifically, things like references not resolving properly, missing required elements for a route spec, etc.

As long as those are covered with this usage, I'm happy with the change since it is certainly simpler.

alfredodeza: I was able to fiddle around with one of the YAMLs and see that an invalid reference was caught:

```text
test/unit/anchore_engine/services/test_api_specs.py::test_api_service[anchore_engine/services/apiext/swagger/swagger.yaml] FAILED
...
            try:
                document = document[part]
            except (TypeError, LookupError):
>               raise RefResolutionError(
                    "Unresolvable JSON pointer: %r" % fragment
                )
E               jsonschema.exceptions.RefResolutionError: Unresolvable JSON pointer: 'Ddefinitions/ApiErrorResponse'

../../.virtualenvs/anchore-engine/lib/python3.8/site-packages/jsonschema-2.6.0-py3.8.egg/jsonschema/validators.py:420: RefResolutionError
```

And I think this would also cover the missing required elements:

```text
E           connexion.exceptions.InvalidSpecification: Path parameter 'subscriptionId' for 'get' operation in '/subscriptions/{subscriptionId}' was not resolved
```

alfredodeza:
@zhill anything else here that you want addressed? Or do you want to hold off on this for now?

alfredodeza:
@zhill ping

PR #345, opened 2020-03-04
Ci use makefile

Btodhunter:
Uses a makefile for doing CI pipeline tasks
Adds e2e testing
Streamlines release process by using git tags for building RC/release images
refactors Dockerfile to better take advantage of build caching

**Special notes**:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,70 +1,95 @@
 # CircleCI 2.1 configuration file
-#
 # Check https://circleci.com/docs/2.0/configuration-reference/ for more details
-#
-# The following environment variables must be set in the circleci project UI
-# - $IMAGE_NAME -> name of image to push to dockerhub
-# - $IMAGE_TAG -> tag on image to push to dockerhub
-# - $DOCKER_USER -> dockerhub user with permissions to repo
-# - $DOCKER_PASS -> password for dockerhub user
 #
 # Note: YAML anchors allow an object to be re-used, reducing duplication.
 # The ampersand declares an alias for an object, then later the `<<: *name`
 # syntax dereferences it.
-# See http://blog.daemonl.com/2016/02/yaml.html
-# To validate changes, use an online parser, eg.
-# http://yaml-online-parser.appspot.com/
 #
 # Define YAML anchors
-var_1: &load_docker_image
+.global_environment_vars: &global_environment_vars
+  PROD_IMAGE_REPO: anchore/anchore-engine-dev
+  RELEASE_BRANCHES: 0.2 0.3 0.4 0.5 0.6
+  VERBOSE: true
+  TERM: xterm
+
+.attach_workspace: &attach_workspace
+  attach_workspace:
+    at: /home/circleci/workspace/
+
+.load_docker_image: &load_docker_image
   run:
     name: Load Docker image layer cache
     command: |
-      docker load -i "/home/circleci/workspace/caches/${CIRCLE_PROJECT_REPONAME}-ci.tar" || true
+      docker load -i "/home/circleci/workspace/caches/${CIRCLE_SHA1}.tar" || true
 
-var_2: &attach_workspace
-  attach_workspace:
-    at: /home/circleci/workspace/
+.ssh_forward_port: &ssh_forward_port
+  run:
+    name: Forward remote-docker ports for testing
+    command: |
+      ssh -MS anchore-api -fN4 -L 8228:localhost:8228 remote-docker
+      ssh -MS k8s-api -fN4 -L 32768:localhost:32768 remote-docker
+
+.filter_rc_tags: &filter_rc_tags
+  branches:
+    ignore: /.*/
+  tags:
+    only: /^v[0-9]+(\.[0-9]+)*-rc[0-9]+$/
+
+.filter_rebuild_tags: &filter_rebuild_tags
+  branches:
+    ignore: /.*/
+  tags:
+    only: /^v[0-9]+(\.[0-9]+)*-r[0-9]+$/
+
+.filter_release_branch: &filter_release_branch
+  branches:
+    only: 
+      - /^[0-9]+(\.[0-9]+)$/
+      - master
+      - /^ci-.*/
+
+.filter_semver_tags: &filter_semver_tags
+  branches:
+    ignore: /.*/
+  tags:
+    only: /^v[0-9]+(\.[0-9]+)*$/
 
 # Start circleci configuration
 version: 2.1
 orbs:
-  anchore: anchore/anchore-engine@1.5.0
+  anchore: anchore/anchore-engine@1.6.5
 commands:
-  run_tox_tests:
-    description: Pass path to tests to be run with tox
+  run_tests:
+    description: |
+      Uses make script to run tests
+      Saves the tox cache & stores log output as artifacts
     parameters:
-      test_path:
+      test_type:
         type: string
     steps:
-    - checkout
-    - restore_cache:
-        keys:
-        - v1-dependencies-{{ .Branch }}-{{ checksum "requirements.txt" }}
-    - run:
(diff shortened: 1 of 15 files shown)


scripts/ci/make/tasks
```diff
@@ -0,0 +1,148 @@
...
+
+######################################################
+###   PROJECT SPECIFIC FUNCTIONS - RUNTIME ORDER   ###
+######################################################
+
+build() {
+    # dont use the docker build cache if building from a tag
+    print_colorized WARN "building image ${TEST_IMAGE_NAME}"
+    if [[ -n "${GIT_TAG}" ]]; then
+        local build_cache="--no-cache"
```
alfredodeza: abstractions like these should have some way to advertise themselves either on terminal output or a log file. When debugging why things aren't working exactly right, or trying to extend/modify what is currently being done, the commands should be logged somewhere. Is it possible that these are being logged and I am just not seeing it?

scripts/ci/make/tasks
```diff
@@ -0,0 +1,148 @@
...
+    elif [[ "${GIT_BRANCH}" =~ "${RELEASE_BRANCHES}" ]]; then
+        print_colorized WARN "tagging & pushing image -- ${branch_image}"
+        docker tag "${dev_image}" "${branch_image}"
+        docker push "${branch_image}"
+    fi
+}
+
+e2e_tests() {
+    # Use the RC image if a semver tag is pushed
+    if [[ "${GIT_TAG}" =~ ^v[0-9]+(\.[0-9]+)*$ ]]; then
```
alfredodeza: this is another instance where logging would be necessary: it relies on a regular expression that will fallback to something else if it doesn't match. Nothing here would advertise this situation. The regex looks fine, but it is obscuring the `else` condition.

scripts/ci/make/util
```diff
@@ -0,0 +1,173 @@
...
+}
+
+clean_python_env() {
+    print_colorized ERROR "deleting virtualenv & build data"
+    sleep 3
+    rm -rf "${VENV}" \
+        *.egg-info \
+        dist \
+        build
+    find . -name '*.pyc' -exec rm -f {} \;
```
alfredodeza: newer Python versions leave a `__pycache__` directory behind as well. One way I do this is with:

```bash
find . \( -type f -name "*.py[co]" -o -type d -name "__pycache__" \) -delete
```

scripts/ci/make/util
```diff
@@ -0,0 +1,173 @@
...
+    docker-compose -f scripts/ci/docker-compose-ci.yaml down
+}
+
+docker_compose_up() {
+    docker-compose -f scripts/ci/docker-compose-ci.yaml up -d
+}
+
+install() {
+    print_colorized WARN "installing ${GIT_REPO} to ${VENV}"
+    pip install .
```
alfredodeza: why is this using `pip install` ? I would prefer to use the more standard way with `python setup.py install`. This function is also saying it is going to install in a `${VENV}` but that is only true if `pip` is coming from the virtualenv, it isn't checking whether that is true or not

scripts/ci/make/util
```diff
@@ -0,0 +1,173 @@
...
+    if [[ ! -x "${VENV}/bin/kubectl" ]]; then
+        print_colorized WARN "installing kubectl"
+        curl -sSLo "${VENV}/bin/kubectl" "https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/${arch}/amd64/kubectl"
+        chmod +x "${VENV}/bin/kubectl"
+    fi
+}
+
+install_testing_deps() {
+    print_colorized WARN "installing dependencies: $*"
+    mkdir -p .tox
```
alfredodeza: why is this directory being created?

scripts/ci/make/util
```diff
@@ -0,0 +1,173 @@
...
+    for p in "$@"; do
+        if [[ ! "$(echo ${VENV}/bin/*)" =~ "$p" ]]; then
```
alfredodeza: is this trying to match the dependency to the script name? If so... there are many situations where this doesn't hold true

scripts/ci/make/util
```diff
@@ -0,0 +1,173 @@
...
+
+setup_venv() {
+    local python_binary_name="${PYTHON##*/}"
+
+    if [[ ${CI} == true ]]; then
+        hash pip || apt install -y python-pip
+        hash virtualenv || pip install virtualenv
+    else
+        hash pip || (print_colorized ERROR "ensure python-pip is installed before attempting to setup virtualenv" && exit 1)
+        hash virtualenv || (print_colorized ERROR "ensure virtualenv is installed before attempting to setup a virtual environment" && exit 1)
```
alfredodeza: for python3 this isn't necessary. Is this a helper for Python 2 ? In Python3 you can do:

```
$ python3 -m venv venv
```

scripts/ci/make/util
Btodhunter: It's trying to find if any of the binaries in your venv match the name of the dependency. I was trying to have it not install the dependency if it already exists in your venv. In particular, the issue was that anchore-cli is installed using the pip package `anchorecli` but the binary is called `anchore-cli` so I couldn't do `hash $p` to see if it was availabe in your PATH. This is one of the few situations where ansible would be really helpful. Or maybe you know another bash trick to make this determination?

scripts/ci/make/util
Btodhunter: that's not needed anymore - forgot to remove it.

scripts/ci/make/util
alfredodeza: If you really wanted to create a faithful delta on what is installed vs. what is being asked to install, `pip freeze` could get the output and then a `grep` to see if it is there. Alternatively, this could just do without the trouble of doing the check and install. `pip` will not re-install if the package is already there.

And yes, a framework like Ansible would prevent this situation with:

```yaml
- pip:
    name: anchorecli
    virtualenv: { VENV }
```

scripts/ci/make/tasks
Btodhunter: added a print statement there

scripts/ci/make/tasks
Btodhunter: logging added

scripts/ci/make/tasks
```diff
@@ -0,0 +1,155 @@
...
+
+    print_colorized WARN "tagging and pushing image -- ${dev_image}"
+    docker tag "${TEST_IMAGE_NAME}" "${dev_image}"
+    docker push "${dev_image}"
+
+    if [[ "${GIT_BRANCH}" == 'master' ]]; then
+        print_colorized WARN "tagging & pushing image -- ${latest_image}"
+        docker tag "${dev_image}" "${latest_image}"
+        docker push "${latest_image}"
+    elif [[ "${GIT_BRANCH}" =~ "${RELEASE_BRANCHES}" ]]; then
```
zhill: From Shellcheck (pycharm highlighted this for me): https://github.com/koalaman/shellcheck/wiki/SC2076

scripts/ci/make/tasks
```diff
@@ -0,0 +1,155 @@
...
+    if [[ "${CI}" == true ]]; then
+        print_colorized WARN "pulling RC image for release -- ${rc_image}"
+        docker pull "${rc_image}"
+
+        print_colorized WARN "tagging and pushing image -- ${prod_image}"
+        sleep 3
+        docker tag "${rc_image}" "${prod_image}"
+        docker push "${prod_image}"
+
+        if [[ "${GIT_BRANCH}" =~ "${LATEST_RELEASE_BRANCH}" ]]; then
```
zhill: From Shellcheck (pycharm highlighted this for me): https://github.com/koalaman/shellcheck/wiki/SC2076

PR #363, opened 2020-03-19
tests: cleanup langpack utils and increase test coverage

alfredodeza:
**What this PR does / why we need it**: Brings up the coverage up to 100% of the langpack utils increasing robustness:

* Removes unneeded usage of `global` declarations
* Replaces use of `return(object)` with `return object`
* Removes unused/dead code
* Removes code paths that are impossible to get to
* Replaces use of bare excepts with granular exceptions 
* Adds a few thousand tests with parametrization via pytest
* Ports the TestCase loop test that uses print statements to parametrize: No need for loops or printing output, treats each variation as an individual test making it easier to debug on failures.
* Brings up the coverage of `langpack.py` to 100%


anchore_engine/util/langpack.py
```diff
@@ -124,137 +120,25 @@ def normalized_version_match(rawsemver, rawpkgver, language='python'):
...
+
                 if not inrange:
                     violation = True
                     break
-            else:
-                raise Exception("unknown range format {}".format(rangecheck))
 
         if not violation:
             inrange = True
-            rangematch = vrange
```
alfredodeza: unused

anchore_engine/util/langpack.py
```diff
@@ -124,137 +120,25 @@ def normalized_version_match(rawsemver, rawpkgver, language='python'):
...
-                op,verraw = (patt.group(1), patt.group(2))
+                op, verraw = (patt.group(1), patt.group(2))
                 inrange = language_compare(rawpkgver, op, verraw, language=language)
-                
+
                 if not inrange:
                     violation = True
                     break
-            else:
-                raise Exception("unknown range format {}".format(rangecheck))
```
alfredodeza: impossible to reach this because both `tokre` and `patt` are preventing it

anchore_engine/util/langpack.py
```diff
@@ -124,137 +120,25 @@ def normalized_version_match(rawsemver, rawpkgver, language='python'):
...
             inrange = True
-            rangematch = vrange
             break
 
     if inrange:
         versionmatch = True
 
     return(versionmatch)
 
-def convert_mrange_to_srange(rawsemver):
```
alfredodeza: none of these are used anywhere

test/unit/anchore_engine/util/test_langpack.py
```diff
@@ -1,168 +1,300 @@
-import subprocess
-import unittest
+import pytest
 from anchore_engine.util.langpack import compare_versions
 
 enable_training = False
 
-class TestSemverVersionHandling(unittest.TestCase):
-    versions = [
```
alfredodeza: all of these, plus some new ones, are ported to fixtures+parametrize for a total of about 4 thousand tests

Issue #366, opened 2020-03-19
Remove/update any API that expects a body on a DELETE request


zhill:
While not forbidden, body on a DELETE is not common practice and thus tooling (e.g. connexion) no longer handles it properly. All Anchore API routes with this behavior must be updated. This is a breaking change that will require a new api version.

zhill:
Closing as resolved by PR #368

PR #368, opened 2020-03-20
use query args instead of JSON body when doing an HTTP DELETE

alfredodeza:
**What this PR does / why we need it**:  While not forbidden, body on a DELETE is not common practice and thus tooling (e.g. connexion) no longer handles it properly. All Anchore API routes with this behavior must be updated. This is a breaking change that will require a new api version.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes issue #366

### DIFF
--- a/anchore_engine/plugins/authorization/client.py
+++ b/anchore_engine/plugins/authorization/client.py
@@ -227,8 +227,7 @@ class AuthzPluginHttpClient(BasicApiClient):
             resp.raise_for_status()
 
     def delete_domain(self, domain: str):
-        req = Domain(domain)
-        resp = self.call_api(requests.delete, 'domains', body=json.dumps(req.to_json()))
+        resp = self.call_api(requests.delete, 'domains', query_params={'domain': domain})
 
         if resp.status_code in [200, 204]:
             return True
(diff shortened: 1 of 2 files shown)


anchore_engine/plugins/authorization/swagger/swagger.yaml
```diff
@@ -63,7 +63,7 @@ paths:
       description: Delete information about a domain, if applicable
       parameters:
         - name: domain
-          in: body
+          in: query
           required: true
           schema:
             $ref: "#/definitions/Domain"
```
zhill: These schema refs need to be changed to string types rather than json objects.

anchore_engine/plugins/authorization/swagger/swagger.yaml
```diff
@@ -93,7 +93,7 @@ paths:
       description: Delete information about a principal, if applicable
       parameters:
         - name: principal
-          in: body
+          in: query
           required: true
           schema:
             $ref: "#/definitions/Principal"
```
zhill: Same here, can be string type instead

anchore_engine/plugins/authorization/client.py
```diff
@@ -228,15 +228,15 @@ def initialize_domain(self, domain: str):
 
     def delete_domain(self, domain: str):
         req = Domain(domain)
```
zhill: Can remove this, as not used.

anchore_engine/plugins/authorization/client.py
```diff
@@ -252,15 +252,15 @@ def initialize_principal(self, principal: str):
 
     def delete_principal(self, principal: str):
         req = Principal(principal)
```
zhill: Same here: Can remove this, as not used.

PR #370, opened 2020-03-20
Convert centos, fedora, and rhel images to use new rhel feed for CVE-based data instead of RHSA data

zhill:


zhill:
Will wait and rebase this after #369 is merged to keep the db upgrade clean

Issue #364, opened 2020-03-19
tests: create a linter, start cleanup process


alfredodeza:
**Is this a request for help?**:

---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):
Start the cleanup process so it gets to a point where it is possible to catch problems with a linter. At the moment, linters are reporting too much and it is hard to detect an actual issue, like for example this one:

./anchore_engine/clients/localanchore_standalone.py:290:57: F821 undefined name 'p'

The `flake8` tool allows to look for specific violations only, so after cleaning up each item, this one needs to get added to tox. For example, if this issue addresses first the removal of bare exceptions, then the `E722` check needs to be explicitly added:

```ini
[flake8]
command = "flake8 --select=E722"
```

This way, once the cleanup is completed, it will prevent PRs from adding bare exceptions again.

This is the complete list of items to fix:

* `E111` 
* `E117`
* `E122`
* `E124`
* `E125`
* `E127`
* `E128`
* `E131`
* `E201`
* `E202`
* `E203`
* `E211`
* `E221`
* `E222`
* `E225`
* `E231`
* `E252`
* `E261`
* `E262`
* `E265`
* `E266`
* `E272`
* `E301`
* `E302`
* `E303`
* `E305`
* `E306`
* `E401`
* `E402`
* `E501`
* `E703`
* `E711`
* `E712`
* `E722`
* `E741`
* `F401`
* `F403`
* `F405`
* `F811`
* `F821`
* `F841`
* `F901`
* `W291`
* `W292`
* `W293`
* `W391`
* `W605`

PR #373, opened 2020-03-23
lint: removes unnecesary use of parens on statements

alfredodeza:
**What this PR does / why we need it**: There is no need for parentheses on statements. For example this:

```python
    return(1)
```

Is the same as:

```python
    return 1
```

Although stylistic, the unnecessary parentheses can cause confusion, making a contributor double-check what is actually occurring (e.g. a tuple or a callable?). 

This PR also adds `make lint` to the CI config so that it prevents newer PRs from getting it in.

**Which issue this PR fixes** Partially fixes issue #https://github.com/anchore/anchore-engine/issues/364

**Special notes**: I *think* I managed to get this right, I double checked, but will need to rely on tests to ensure I didn't remove something that was actually needed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ commands:
             - v1-dependencies-{{ .Branch }}-{{ checksum "requirements.txt" }}
       - run:
           name: run tests
-          command: make test-<< parameters.test_type >>
+          command: make << parameters.test_type >>
       - save_cache:
           paths:
             - ~/project/.tox
@@ -91,7 +91,16 @@ jobs:
     steps:
       - checkout
       - run_tests:
-          test_type: unit
+          test_type: test-unit
+
+  lint:
+    docker:
+      - image: circleci/python:3.6
+    environment: *global_environment_vars
+    steps:
+      - checkout
+      - run_tests:
+          test_type: lint
 
   integration_tests:
     docker:
@@ -105,7 +114,7 @@ jobs:
         environment:
           POSTGRES_PASSWORD: "postgres"
     resource_class: large
-    environment: 
+    environment:
       <<: *global_environment_vars
       ANCHORE_TEST_S3_ACCESS_KEY: "9EB92C7W61YPFQ6QLDOU"
       ANCHORE_TEST_S3_SECRET_KEY: "TuHo2UbBx+amD3YiCeidy+R3q82MPTPiyd+dlW+s"
@@ -124,7 +133,7 @@ jobs:
     steps:
       - checkout
       - run_tests:
-          test_type: integration
+          test_type: test-integration
 
   build:
     docker:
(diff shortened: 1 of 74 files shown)


.circleci/config.yml
```diff
@@ -88,7 +88,15 @@ jobs:
     steps:
       - checkout
       - run_tests:
-          test_type: unit
+          test_type: test-unit
+
+  lint:
```
alfredodeza: @Btodhunter this didn't seem like it got picked up in the current PR. Is this something that is going to be triggered *after* this is merged? I could swear I've seen configs to be executed in the same PR that implements them

Btodhunter: It will run on the same PR that implements it, you just need to add a lint job to the workflows.

zhill:
To be clear: this isn't needed for 0.7.0 release, correct?

alfredodeza:
It isn't needed for the release, no.

alfredodeza:
Now that the release is almost behind us, would love to get this reviewed and merged, because it touches on so many different places that rebasing is a pain :(

.circleci/config.yml
```diff
@@ -198,6 +206,7 @@ workflows:
   default_workflow:
```
Btodhunter: The `lint` job should gate either unit_tests or integration_tests & build to fail the pipeline early when the linter fails. The lint job should also be added to the RC & Rebuild workflows (a rebase might be needed to see where lint fits into the RC workflow)

alfredodeza: I've added linting for the RC, Build, and Rebuild workflows. I don't think we should gate on unit tests or integration tests however. As a developer, I would like to see all unit tests and integration passing even if the linter complains. 

Preventing a build does sound great though, so I've made sure it is required on those workflows.

Btodhunter: lgtm!

.circleci/config.yml
```diff
@@ -231,8 +241,12 @@ workflows:
 
   rc_image_workflow:
     jobs:
+      - lint:
+          filters: *filter_rebuild_tags
```
Btodhunter, line 245: this should be `*filter_rc_tags`

alfredodeza: oof good catch

.circleci/config.yml
```diff
@@ -91,7 +91,15 @@ jobs:
     steps:
       - checkout
       - run_tests:
-          test_type: unit
+          test_type: test-unit
+
+  lint:
+    docker:
+      - image: circleci/python:3.6
```
Btodhunter, line 98: can you add `environment: *global_environment_vars` to this job to be consistent with the rest of the jobs? I don't think this job will actually _need_ any of the current env vars, but the idea of that anchor was to provide a place to set global vars for every job.

alfredodeza:
@Btodhunter from the CircleCi perspective this is good to go right?

.circleci/config.yml
Btodhunter: @alfredodeza did your force push overwrite this fix?

alfredodeza: Not sure... just pushed again to fix this

Issue #395, opened 2020-04-01
upgrade helm chart to anchore-engine 0.70, database is not right version


rodrigc:
**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):

**Version of Anchore Engine and Anchore CLI if applicable**:

anchore-cli, version 0.7.0

```
Name: anchore-engine
Version: 0.7.0
Summary: Anchore Engine
Home-page: http://www.anchore.com
Author: Anchore Inc.
Author-email: dev@anchore.com
License: Apache License 2.0
Location: /usr/local/lib/python3.6/site-packages
Requires: toastedmarshmallow, python-dateutil, SQLAlchemy, Authlib, Werkzeug, yosai, prometheus-flask-exporter, Flask, Click, urllib3, prettytable, six, connexion, passlib, ijson, psycopg2-binary, psutil, zope.interface, python-keystoneclient, semantic-version, zope.component, Twisted, requests, pytz, python-swiftclient, pathlib, swagger-spec-validator, jsonschema, boto3, watchdog, typing, pg8000, prometheus-client, treelib, retrying, PyYAML
```

**What happened**:

I originally had anchore-engine 0.6.1 installed.
I upgraded the helm chart, and did:

```
helm -n anchore upgrade -f values.yaml anchore ./
```

**What did you expect to happen**:

I expected the upgrade to complete successfully.

However, I saw this:

```
kubectl -n anchore logs anchore-anchore-engine-simplequeue-7985c89749-wp4v8

[MainThread] [anchore_manager.cli.service/start()] [WARN] this version of anchore-engine requires the anchore DB version (0.0.13) but we discovered anchore DB version (0.0.12) in the running DB - it is safe to run the upgrade while seeing this message - will retry for 3485 more seconds.
[MainThread] [anchore_manager.cli.service/start()] [WARN] this version of anchore-engine requires the anchore DB version (0.0.13) but we discovered anchore DB version (0.0.12) in the running DB - it is safe to run the upgrade while seeing this message - will retry for 3480 more seconds.
```

Looks like the database did not upgrade for me.
How can I force the database to upgrade?

rodrigc:
I read the source code to anchore-manager, and based on the environment variables that were set,
I tried this command:

```
anchore-manager db --db-connect "postgresql://$ANCHORE_DB_HOST/$ANCHORE_DB_NAME?user=$ANCHORE_DB_USER&password=$ANCHORE_DB_PASSWORD" upgrade
```

[...]

PR #396, opened 2020-04-01
Add pointer to database upgrade docs when database out-of-version errors are logged

rodrigc:
Signed-off-by: Craig Rodrigues <rodrigc@crodrigues.org>

When the user sees an error message about the database being out of date, it is not
obvious what to do to fix the problem.

Fixes: https://github.com/anchore/anchore-engine/issues/395
Addresses: https://github.com/anchore/anchore-engine/issues/395#issuecomment-607225619


zhill:
Thanks @rodrigc. I'll take a look and we'll make sure this condition is both handled better in the chart as well as in the docs and system itself. Great first step that direction!

anchore_manager/cli/service.py
```diff
@@ -340,6 +340,7 @@ def start(services, auto_upgrade, anchore_module, skip_config_validate, skip_db_
                             logger.warn(
                                 "this version of anchore-engine requires the anchore DB version ({}) but we discovered anchore DB version ({}) in the running DB - it is safe to run the upgrade while seeing this message - will retry for {} more seconds.".format(
                                     str(code_versions['db_version']), str(db_versions['db_version']), str(max_timeout - int(time.time() - timer))))
+                            logger.warn("See: https://docs.anchore.com/current/docs/engine/engine_installation/upgrade/#advanced--manual-upgrade-procedure")
```
zhill, line 343: Please remove this message as this is an expected state while the components wait for the system to upgrade itself. The other message you've added handles the case where the retry count completed with no upgrade found and thus the user *may* need to take action.

rodrigc: @zhill OK I removed it

PR #402, opened 2020-04-07
Bump requests and Twisted dependencies

alfredodeza:
**What this PR does / why we need it**: Bumps Twisted that fixes two CVEs (CVE-2020-10108 and CVE-2020-10109) and requests to avoid the `idna<2.9` dependency that collides with Twisted's

It also moves optional dependencies down, under a commented block so that they are easier to identify as a group.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,4 +1,3 @@
 pytest
-requests==2.22.0
 Flask-Testing
 memory_profiler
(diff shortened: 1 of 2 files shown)


zhill:
This looks good but I think needs: https://github.com/anchore/anchore-cli/pull/73 to pass tests to get the idna version aligned with requests version

PR #403, opened 2020-04-07
Initial test coverage for the analyzer

alfredodeza:
**What this PR does / why we need it**: Adds test coverage for the `analyze_image()` function that downloads manifests, images, squashes, runs analyzers, and produces a report. It concentrates in retrieving the actual report (a large Python dictionary mixed with JSON) and uses a Pytest harness to run tests against it.

The initial tests do not cover everything. It covers the one case of going to a registry and retrieving an existing image (CentOS 8).

There is room to add additional variations of other images in other locations (directory and/or with skopeo).

By default it caches the initial+expensive processing so that all tests don't need to re-do the processing. 

**Special notes**: A separate command-line tool and library is added, understanding that there is a similar tool in `anchore-manager analyzer`. The reason for keeping this tool is decoupling from anchore-engine as much as possible and concentrating in the result produced, so that analyzers and other Docker interactions can be tested directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,8 +24,7 @@ RUN set -ex && \
 RUN set -ex && \
     echo "installing anchore" && \
     pip3 wheel --wheel-dir=/build_output/wheels . && \
-    # pip3 wheel --wheel-dir=/build_output/wheels/ git+git://github.com/anchore/anchore-cli.git@$CLI_COMMIT\#egg=anchorecli && \
-    pip3 wheel --wheel-dir=/build_output/wheels/ git+git://github.com/robertp/anchore-cli.git@$CLI_COMMIT\#egg=anchorecli && \
+    pip3 wheel --wheel-dir=/build_output/wheels/ git+git://github.com/anchore/anchore-cli.git@$CLI_COMMIT\#egg=anchorecli && \
     cp ./LICENSE /build_output/ && \
     cp ./conf/default_config.yaml /build_output/configs/default_config.yaml && \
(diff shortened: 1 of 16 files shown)


scripts/ci/make/pipeline_tasks
```diff
@@ -77,7 +77,14 @@ functional_tests() {
     activate_venv "${VENV}"
     hash tox || "${PYTHON}" -m pip install tox
     hash anchore-cli || "${PYTHON}" -m pip install -q anchorecli
-
+    if [[ "$CI" == "true" ]] && ! hash skopeo; then
```
alfredodeza, line 80: I generally don't mind scripts as long as they are kept super short and don't deal with too much logic. These lines below ensure `skopeo` is installed, but there are many problems with this approach and I am concerned it will start becoming a copy/pasta pattern repeated everywhere:

* `sudo` handling looks a bit odd because it is piping via `echo` so it requires `su -c` and then the command as a string. This decreases readability as there is no longer any highlighting.
* The `apt-get update` has to be done regardless if this is the first or tenth package being added. I anticipate this will not be the last dependency that requires installation and others will pile up on the `apt-get update`
* Enlarges the `functional_tests` to a point where it is no longer concise as it has to deal with dependencies.

This is fine for now, but it is important to raise awareness of the potential problems here.

alfredodeza:
@Btodhunter, @zhill , @wagoodman this is good for a re-review if you guys can spare a few. I have more changes I could add but this is good enough to iterate over.

zhill:
@alfredodeza ugh, looks like conflicts again. Can you rebase/resolve conflict and go ahead and merge it.

PR #434, opened 2020-04-23
Compose cleanup

zhill:
Clarify docker-compose in this repo is for development. New updates to the documentation will provide a docker-compose.yaml there for getting started with released versions.

Removes the docker-compose.yaml from being bundled in the image itself, breaking update dependencies on new image builds.

We can add a released-version docker-compose here later if needed, but docs are better spot for updates outside of code development.


zhill:
I will also copy/move the nginx and prometheus configs into the docs as well so there is one consistent location for all the quickstart stuff. Those can remain here but also are not bundled into the image itself.

docker-compose-dev.yaml
```diff
@@ -1,99 +1,99 @@
...
     expose:
     - 8228
     logging:
       driver: "json-file"
       options:
         max-size: 100m
     environment:
-    - ANCHORE_ENDPOINT_HOSTNAME=engine-analyzer
+    - ANCHORE_ENDPOINT_HOSTNAME=analyzer
     - ANCHORE_DB_HOST=anchore-db
```
Btodhunter, line 91: missed a `anchore-db` that needs to change to `db`

Btodhunter, line 91: missed a anchore-db that needs to change to db

zhill: :+1:

devorbitus:
Just an FYI, this change broke this blog post https://anchore.com/blog/docker-image-security-in-5-minutes-or-less/

PR #439, opened 2020-04-27
Add licenses field to image content response

### LINKED ISSUE #313: The license field needs better individual license delimiting
**Is this a request for help?**:
No
---

FEATURE REQUEST

The license field today is a single string field of space-separated licenses. This would be fine except in the case where a license itself has spaces in it.

I would request that the license field be a pure JSON list field instead so it's easier for API clients to parse.

wagoodman:
Adds a new `licenses` array field to the content response and adds a deprecation notice for the `license` field. Additionally moves the existing `make_image_content_response` into a separate `helpers` module to aid in testing.

Note: an additional `anchore-cli` PR will be opened soon to utilize this new field.

(Closes #313 )


.gitignore
```diff
@@ -1,3 +1,5 @@
+*.cache
```
alfredodeza, line 1: Why is this needed?

anchore_engine/services/apiext/api/helpers/image_content_response.py
```diff
@@ -0,0 +1,220 @@
...
+    for package in list(content_data.keys()):
+        el = {}
+        try:
+            el["package"] = content_data[package]["name"]
+            el["type"] = "PYTHON"
+            el["location"] = content_data[package]["location"]
+            el["version"] = content_data[package]["version"]
+            el["origin"] = content_data[package]["origin"] or "Unknown"
+            el["license"] = content_data[package]["license"] or "Unknown"
+            el["licenses"] = content_data[package]["license"].split(" ") or ["Unknown"]
```
alfredodeza, line 103: Why does it need to default to a list with `["Unknown"]` instead of an empty list?

anchore_engine/services/apiext/api/helpers/image_content_response.py
```diff
@@ -0,0 +1,220 @@
...
+    try:
+        for package in list(content_data.keys()):
+            el = {}
+            try:
+                el["package"] = content_data[package]["name"]
+                el["type"] = content_data[package]["type"].upper()
+                el["location"] = content_data[package].get("location", None) or "Unknown"
+                el["version"] = content_data[package].get("version", None) or "Unknown"
+                el["origin"] = content_data[package].get("origin", None) or "Unknown"
+                el["license"] = content_data[package].get("license", None) or "Unknown"
```
alfredodeza, line 196: If the line below splits on whitespace, wouldn't this cause multiple licenses to be pilled up together?

test/unit/anchore_engine/services/apiext/api/helpers/test_images.py
```diff
@@ -0,0 +1,314 @@
...
+            "release": "N/A",
+            "size": "4064000",
+            "sourcepkg": "apt-1.8.2",
+            "type": "dpkg",
+            "version": "1.8.2",
+        }
+
+    @pytest.fixture
+    def muli_license_content_data(self, content_data_entry):
+        response = content_data_entry.copy()
```
alfredodeza, line 29: Does it need to copy here? The default for the fixture will be to create the dictionary for every test, just like `setup()` would. Did you experience a different behavior?

anchore_engine/services/apiext/api/helpers/image_content_response.py
```diff
@@ -0,0 +1,220 @@
...
+        except:
+            continue
+        response.append(el)
+    return response
+
+
+def _build_docker_history_response(content_data):
+    try:
+        return utils.ensure_str(
+            base64.encodebytes(utils.ensure_bytes(json.dumps(content_data)))
```
alfredodeza, line 162: This encoding pattern is repeated a few times, could this be extracted into a helper that accepts the incoming string? In this case `content_data` gets dumped, but the use is the same on the next three

.gitignore
wagoodman: I noticed when running locally that a cache file is created (`twisted/plugins/dropin.cache`), didn't want to accidentally commit it

anchore_engine/services/apiext/api/helpers/image_content_response.py
wagoodman: Good point, I was trying to keep to the same behavior as before, which was not to list an empty string when you could not determine the presence of licenses, but to explicitly list that the license is "Unknown"... I kept suite with the same logic for multiple licenses (that is, there aren't explicitly 0 licenses, instead the licenses could not be positively determined).

I'm not 100% in love with `["Unknown"]` but am open for suggestions.

anchore_engine/services/apiext/api/helpers/image_content_response.py
wagoodman: `el['license']` is intended to have multiple licenses in a single string field (delimited by spaces --presumably). That is the current behavior/expectation for this field today.

`el['licenses']` is intended to be foremost a list type, even if the string parse fails. Assuming a licenses string (e.g. `"MIT GPLv2"`) splitting by space would be sufficient (e.g. `["MIT", "GPLv2"]`). However, if a single license has a space in it, or they are comma delimited, then this default behavior will fail. Note that this is the "default" handler, that is, there is no guide what is "right" here... it's best guess.

test/unit/anchore_engine/services/apiext/api/helpers/test_images.py
wagoodman: This is in case both fixtures are used in the same test, and depending on the load order for the test, one fixture would affect the other. I admit I didn't try that out though, and copied just for safety --I'll give it a shot and report back.

wagoodman: yup, the copy is needed for safety. Though, currently I'm not using both fixtures in a single test, in case that ever ends up being the case, they are safely decoupled with the `copy()` statement:

```python
   
 # note: "muli_license_content_data" is not using copy()

    def test_thing(self, content_data_entry, muli_license_content_data):
        assert content_data_entry['license'] == 'MIT'
        assert muli_license_content_data['license'] == "MIT GPLv2"
...
```

results in:
```
...
    def test_thing(self, content_data_entry, muli_license_content_data):
>       assert content_data_entry['license'] == 'MIT'
E       AssertionError: assert 'MIT GPLv2' == 'MIT'
E         - MIT
E         + MIT GPLv2

test/unit/anchore_engine/services/apiext/api/helpers/test_images.py:133: AssertionError
```

alfredodeza: Ah yes, you can't mix and match a fixture that uses another fixture, and then request that same fixture expecting it to be the same. This is because the `return {...}` is already associated with the fixture name, making it effectively a variable value that has already been computer. 

When that is _not the case_, the fixtures don't need to do a copy at all. Using this class for example:

```python
class TestFixture:

    @pytest.fixture
    def obj(self):
        return {'key': 'value'}

    def test_mangle_obj(self, obj):
         assert  obj['key'] == 'value'
         assert obj.get('other') is None
         obj['key'] = 1
         obj['other'] = 2

    def test_mangle_obj_2(self, obj):
         assert  obj['key'] == 'value'
         assert obj.get('other') is None
         obj['key'] = 1
         obj['other'] = 2

    def test_mangle_obj_3(self, obj):
         assert  obj['key'] == 'value'
         assert obj.get('other') is None
         obj['key'] = 1
         obj['other'] = 2
```

Does not produce any failures:

```text
collecting ... collected 3 items

test_common.py::TestFixture::test_mangle_obj PASSED                                                                                                                                                                                                                      [ 33%]
test_common.py::TestFixture::test_mangle_obj_2 PASSED                                                                                                                                                                                                                    [ 66%]
test_common.py::TestFixture::test_mangle_obj_3 PASSED                                                                                                                                                                                                                    [100%]
```

anchore_engine/services/apiext/api/helpers/image_content_response.py
wagoodman: @alfredodeza see the updated commit; these were slightly different actions (one required `json.dumps`) so a lambda was used to encapsulate providing data within the try-except block.

PR #457, opened 2020-05-22
Dev/nurmi/173

nurmi:


anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -496,13 +496,57 @@ def evaluate(self, image_obj, context):
...
+                        if vid == vulnerability_cpe.vulnerability_id:
+                            found = True
+                            break
+                    if found:
+                        break
+            if found:
+                found_vids.append(vid)
+                
+        if found_vids:
+            self._fire(msg="Blacklisted vulnerabilities detected: {}".format(found_vids))
```
zhill, line 539: Functionally this works, but creates a strange trigger match for multiple matches. I think it may be preferrable to use a single vulnerability ID per rule rather than one rule that matches many/multiple IDs (I'm thinking of temporary whitelisting or other false-positive handling use-cases).

zhill: This can be mitigated by the user using a single cve per rule.

PR #470, opened 2020-06-02
Fix package type filter in vulnerabilities.package trigger

zhill:
Moves filter to be per-vulnerability for the ImagePackageVulnerability matches since both os and non-os packages can use that mechanism now that we have vulnerability sources that match that way (GitHub Advisories).


zhill:
Closing this PR, will use same branch pushed to anchore-engine repo instead of my fork to enable better debug of the CI issue failing the build.

Issue #473, opened 2020-06-02
Upgrade to 0.0.13 db version can fail if old 'nvd' data present in db


zhill:
This affects installations that have data from 0.5.x using the deprecated 'nvd' feed instead of the newer 'nvdv2' feed.

During db upgrade the system updates counts for all feeds it finds in the 'feeds' and 'feed_groups' tables, and since the code for the 'nvd' feed was removed in 0.6.0, a KeyError is thrown and the upgrade fails.

This was introduced in the 0.7.0 upgrade from 0.6.2 where the db schema is updated from 0.0.12 -> 0.0.13

To reproduce the issue:
If you have db of version `0.0.12` and a record with name='nvd' in the `feeds` table, run `anchore-manager db --db-connect <db str> upgrade`.
You should see an error return with `KeyError 'nvd'` and a non-zero exit code indicating the upgrade did not complete.

PR #475, opened 2020-06-03
Adds old nvd feed class back, but with sync disabled to fix upgrade issue

zhill:
Fix for #473

### DIFF
--- a/anchore_engine/services/policy_engine/engine/feeds/sync.py
+++ b/anchore_engine/services/policy_engine/engine/feeds/sync.py
@@ -64,8 +64,11 @@ class DataFeeds(object):
     @staticmethod
     def update_counts():
         for feed in get_all_feeds_detached():
-            f = feed_instance_by_name(feed.name)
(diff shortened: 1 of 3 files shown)


anchore_engine/services/policy_engine/engine/feeds/sync.py
```diff
@@ -64,8 +64,11 @@ def instance(cls):
     @staticmethod
     def update_counts():
         for feed in get_all_feeds_detached():
-            f = feed_instance_by_name(feed.name)
-            f.update_counts()
+            try:
+                f = feed_instance_by_name(feed.name)
```
nightfurys, line 68: Is there where it's failing in the upgrade path? I'm guessing all other changes are necessary to support flush ops on the feed

zhill: Yes, this would have fixed upgrade without the other changes, but leaves the data in limbo without the other changes.

Issue #491, opened 2020-06-17
Change image analysis queue processing behavior to 'fair share' across accounts


nurmi:
In anchore engine, the image analysis tasks are processed from an internal queue with first-come-first-served behavior.  When multiple accounts are active, this can lead to an effect where on account may have a small number of images that must wait in queue for a potentially long time if another account has many images pending analysis that were adding first, in time.  

We're intending to alter the behavior of the analysis queue such that, at any given time, if multiple accounts all have images pending analysis, at least one image from each account is prioritized at the head of the queue, which will result in a 'fair share' of image analysis tasks as opposed to the current 'first-come-first-served' behavior.

Prioritized images themselves will still retain FCFS within the set that gets promoted, and candidates for promotion will also be processed FCFS.

PR #492, opened 2020-06-17
change image analysis queue processing behavior from first come first served to fair share across account

nurmi:
Implements #491


KnoxAnderson:
Have you thought about prioritizing images coming from the inline scanner at the top of the list as well?

Ie images submitted with an analyze in the inline scan get prioritized above a standard image add submission

nurmi:
> Have you thought about prioritizing images coming from the inline scanner at the top of the list as well?
> 
> Ie images submitted with an analyze in the inline scan get prioritized above a standard image add submission

Hi @KnoxAnderson - we're going to be tracking the ability for a user to explicitly 'prioritize' the analysis of an image separately from adjusting the default queue task processing behavior.  This would extend to any image analysis 'add' operation ideally, and may be part of a bigger enhancement related to task queue management - not on the roadmap yet but it is on the list of enhancements for future consideration!

anchore_engine/services/catalog/__init__.py
```diff
@@ -979,22 +985,39 @@ def handle_analyzer_queue(*args, **kwargs):
...
+                                    queue_rebalance[userId]['lowest_queueId'] = q_record.get('queueId')
+                                if q_record.get('queueId') < 0 and q_record.get('queueId') >= highest_neg_queueId:
+                                    highest_neg_queueId = q_record.get('queueId')
+                            except Exception as err:
+                                logger.error("failed to store image current queueID - excpetion: {}".format(err))
+                                
                     except Exception as err:
                         logger.error("failed to check/queue image for analysis - exception: " + str(err))
 
+    # promote queued tasks into the analysis queue such that one image from each account is prioritized, to implement a simple 'fair share' across accounts
```
zhill, line 1010: I'd like to see this split out into its own function with some test coverage of the behavior so we can catch any regressions later.

nurmi: the logic has been adjusted to support testability, and a unit test has been added

PR #507, opened 2020-06-27
Fixes global analysis rule processing gaps

zhill:
Fixes analysis archive processing to include all accounts to ensure that global rules are processed for each account correctly.


zhill:
Closing. Accidentally opened against the wrong branch.

Issue #502, opened 2020-06-25
Add API for bulk delete of images


nightfurys:
Add an API for deleting multiple images in a single round-trip to minimize the latency for large flush/delete operations.

The request should provide a bounded list of digests to remove and the response should include a status for each requested digest indicating deleted or error.

nightfurys:
Reopening the issue to update the backend implementation for single and bulk delete operations

PR #509, opened 2020-06-30
bulk deletion of images

nightfurys:
- synchronous checks to verify each image can be deleted
- response contains status of the delete op and a message when the checks fail and image can't be deleted
- actual deletion of image triggered in an async thread

Implements #502

### DIFF
--- a/anchore_engine/services/apiext/api/controllers/images.py
+++ b/anchore_engine/services/apiext/api/controllers/images.py
@@ -653,6 +653,33 @@ def list_images(history=None, image_to_get=None, fulltag=None, detail=False):
     return return_object, httpcode
 
 
+@authorizer.requires([ActionBoundPermission(domain=RequestingAccountValue())])
+def delete_images_async(imageDigests, force=False):
+    return_object = {}
+    httpcode = 500
+
+    try:
+        logger.debug('Handling DELETE on imageDigests: %s' % imageDigests)
+
+        client = internal_client_for(CatalogClient, ApiRequestContextProxy.namespace())
+
+        rc = client.delete_images_async(imageDigests, force=force)
+
+        if rc:
+            return_object = rc
+            httpcode = 200
+        else:
+            httpcode = 500
+            raise Exception('Operation failed due to an error/connectivity issue with catalog')
+
+    except Exception as err:
+        logger.exception('Error in asynchronous deletion of images')
+        return_object = make_response_error(err, in_httpcode=httpcode)
+        httpcode = return_object['httpcode']
+
+    return return_object, httpcode
+
+
 def validate_pullstring_is_tag(pullstring):
     try:
         parsed = parse_dockerimage_string(pullstring)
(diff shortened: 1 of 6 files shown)


nightfurys:
@zhill, @nurmi, 1st draft of the API and end-to-end implementation for the async deletion of list of images. Still working on the unit tests and haven't included them in this commit. I've refactored some of the existing code to perform image deletion checks in the synchronous path. The async cleanup of image artifacts is implemented simply by invoking the deletion bits in a new thread. Was initially considering updating the status of the image to 'deleting' on the synchronous path and using a periodic handler to perform the deletion. But as I started going down that route, I found image state referred to the analysis status. Updating analysis state to 'deleting' didn't make a lot of sense to me. Let me know if there's a different image state/status that I may have overlooked. There's still the issue of surfacing deletion errors from the async process with either approach and clean up actions on such failures. I'll keep hacking away at it. Wanted to get some eyes on the API first

anchore_engine/services/catalog/swagger/swagger.yaml
```diff
@@ -189,6 +189,36 @@ paths:
...
+        required: false
+      produces:
+      - "application/json"
+      responses:
+        200:
+          description: "success"
+          schema:
+            type: array
+            items:
+              type: object
```
zhill, line 220: Can this have a spec? I think it can be a simple [{"digest": "xxxxx", "status": "deleting", "errors": []},....] kind of structure.

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -429,6 +429,37 @@ paths:
...
+      x-swagger-router-controller: anchore_engine.services.apiext.api.controllers.images
+      operationId: delete_images_async
+      x-anchore-authz-action: deleteImage
+      summary: Bulk mark images for deletion
+      description: Delete analysis for image digests in the list asynchronously
+      parameters:
+      - name: imageDigests
+        in: query
+        required: true
+        x-nullable: false
```
zhill, line 444: is that required if required=true?

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -429,6 +429,37 @@ paths:
...
+        type: array
+        items:
+          type: string
+        collectionFormat: csv
```
zhill, line 448: We have support for array params in the /query/vulnerabilities route as well. It uses the default collection format which is csv. I think this is good, should be explicit. I can update the other value to be explicit as well.

anchore_engine/services/apiext/api/controllers/images.py
```diff
@@ -1207,6 +1219,40 @@ def analyze_image(account, source, force=False, enable_subscriptions=None, annot
         raise err
 
 
+def images_imageDigests(request_inputs, imageDigests):
```
zhill, line 1222: Can we just use a better model here? I'd prefer to not continue the non-verb specific function decomposition in this file, and move to a format that is actually testable in unit/integration tests.

nightfurys: I went between making this way more simpler/cleaner vs fitting into the current model. I even have the changes in a different branch as I didn't want to lose them entirely. Will switch it

PR #512, opened 2020-07-06
Update skopeo version to v0.2.0

### LINKED ISSUE #484: Update to skopeo 1.1.0
Upgrade the version of skopeo to 1.1.0 or at least 0.2.0 if 1.1.0 is too big a change. Will facilitate removing the strange search tag for repo-add operations where there isn't a 'latest' tag in the repo as the new versions include a command `list-tags` that works without a known tag to inspect.

nightfurys:
Fixes #484

**Special notes**:


zhill:
Since then they've updated to 1.0 and 1.1.0. I think we should try 1.1.0 and use that if it works. I'll update #484 as well.

nightfurys:
> Since then they've updated to 1.0 and 1.1.0. I think we should try 1.1.0 and use that if it works. I'll update #484 as well.

Was just trying to see if the build process would go through with this PR. 1.0.0 seems like a big version jump, sure it won't break our usage?

Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG CLI_COMMIT
 ENV LANG=en_US.UTF-8 LC_ALL=C.UTF-8
 
 ENV GOPATH=/go
-ENV SKOPEO_VERSION=v0.1.41
+ENV SKOPEO_VERSION=v0.2.0
```
zhill, line 10:
```suggestion
ENV SKOPEO_VERSION=v1.1.1
```

zhill:
I'm going to close this so we can try 1.1.1 as well. I'll give that a try and open a PR if tests look good.

PR #516, opened 2020-07-07
VulnerabilityMatchTrigger: use solution date when available

luca-sd:
When evaluating a VulnerabilityMatchTrigger, if the solution_date field is available for a specific non-OS entry, use it to compute the fix date.

Signed-off-by: Luca Guerra <luca.guerra@sysdig.com>

Usually the fix date is computed based on when the CPE was processed, but in at least the case of RBS VulnDB there may be a field indicating more accurately when the vendor made the fix available.

**Special notes**:

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
@@ -1,5 +1,6 @@
 import calendar
 import time
+from datetime import datetime
 from collections import OrderedDict
 
 from anchore_engine.services.policy_engine.engine.feeds.db import get_feed_group_detached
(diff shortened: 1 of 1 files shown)


zhill:
Thanks for the PR @luca-sd! We'll check it out.

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -218,10 +219,19 @@ def evaluate(self, image_obj, context):
 
                                     if is_fix_available and fix_timeallowed is not None:
                                         if fix_available_in:
-                                            if calendar.timegm(vulnerability_cpe.created_at.timetuple()) > fix_timeallowed:
+                                            fix_date = vulnerability_cpe.created_at
+
+                                            # VulnDB entries might have a solution_date which is more accurate than the CPE created_at field
+                                            if vulnerability_cpe.parent.vuln_metadata:
+                                                vulndb_solution_date = vulnerability_cpe.parent.vuln_metadata['solution_date']
```
zhill, line 226: I'd like this to be a bit more protected against KeyErrors using:

```suggestion
                                                vulndb_solution_date = vulnerability_cpe.parent.vuln_metadata.get('solution_date')
```

luca-sd: Good catch, thanks!

luca-sd:
Thanks a lot @zhill ! I agree that there may be other places where the fix date is considered and thanks again for thinking about how this would work.

zhill:
Hi @luca-sd I've not forgotten about this. I've swamped but I am looking into the availability of a similar timestamp for the other feed sources for consistency.

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -218,10 +219,18 @@ def evaluate(self, image_obj, context):
 
                                     if is_fix_available and fix_timeallowed is not None:
                                         if fix_available_in:
-                                            if calendar.timegm(vulnerability_cpe.created_at.timetuple()) > fix_timeallowed:
+                                            fix_date = vulnerability_cpe.created_at
+
+                                            # VulnDB entries might have a solution_date which is more accurate than the CPE created_at field
+                                            if vulnerability_cpe.parent.vuln_metadata:
```
nightfurys, line 225: This is accessing `vuln_metadata` attribute under the assumption that `parent` is an instance of `VulnDBMetadata`. It will fail for other types such as nvd cpes with parent is `NvdV2Metadata` that are schlepped through the same gate logic. Checks and additional error handling are necessary here

luca-sd: Missed this comment. Good point as well, so we should test both if `vuln_metadata` exists and if `solution_date` is a field.

zhill:
I'm closing this as stale. Feel free to update and re-open.

Issue #340, opened 2020-02-19
Add busybox vulnerability scanning support for the platform itself


zhill:
Currently, engine does not support matching vulnerabilities against the busybox "package" itself. Detected applications are scanned (npm, python, java, etc), but because there aren't os packages as with Centos, Deb, etc there is no "os package" vuln scanning support.

The solution is to add a virtual package to the anchore analysis for the busybox binary itself, which anchore already detects as the distro and distro version in the image metadata output, and use that for vulnerability matching against NVD data.

PR #535, opened 2020-07-24
add support for a new content type (binary) with analyzers to detect …

### LINKED ISSUE #315: Anchore CVE FPs on centos non-os packages
**Is this a request for help?**:  NA

---

BUG REPORT

anchore-cli, version 0.6.0-dev
Engine DB Version: 0.0.11
Engine Code Version: 0.5.2
 
**What happened**:
Anchore finds CRITICAL CVEs on many non-os packages that don't seem correct.

i expect when a program has been patched so the software flaw no longer exists, anchore should not incorrectly identify as vulnerable software.  Though i believe this may be an issue with the software package version number versus the release number.  I give a concrete example below with python 2.7.

**Any relevant log output from /var/log/anchore**: NA, anchore is not failing, it is completing the scan, the issue seems to be in the matching logic.  just matching the version number alone of a software package is not sufficient.

**What docker images are you using**:
centos:7

1. $ anchore-cli image add centos:7
2. $ anchore-cli image vuln centos:7 all | grep CVE-2016-5636
CVE-2016-5636           Python-2.7.5        Critical        None        CVE-2016-5636           https://nvd.nist.gov/vuln/detail/CVE-2016-5636
3. $ docker exec -it centos:7 /bin/bash
4.  [container]$ yum info python
    Installed Packages
    Name        : python
    Arch        : x86_64
    Version     : 2.7.5
    Release     : 80.el7_6
5. [container]$  rpm -q --changelog python | grep -B 1 CVE-2016-5636
    * Mon Jul 11 2016 Charalampos Stratakis <cstratak@redhat.com> - 2.7.5-47
    - Fix for CVE-2016-5636: possible integer overflow and heap corruption in zipimporter.get_data()

https://git.centos.org/rpms/python/releases .  this link has an example of various release numbers of python version 2.7.5, to show that just version matching against CVE info will not be sufficient in this case.

### LINKED ISSUE #531: Add "binary" image content support for non-os language runtime installation detection
This enhancement will add a new content type (alongside existing [os, files, npm, gem, python, java, nuget, ...] named 'binary' as a way for anchore engine analyzers to discover and store software binaries that do not fall into specific language or os-package categories.  

For example, language runtime environments such as golang, python that are installed directly from a container build, rather than as os-packages (i.e. rpm, dpkg, apkg, etc).

This new category may also encapsulate other stand alone binary installations of packages in the future (nginx, nodejs, etc).

Software that is categorized by anchore analyzers into this group will match against any vulnerability feeds that support CPE matching (NVD), similar to the language package content types.

### LINKED ISSUE #533: Add ability to detect and match vulnerabilities against non-os python runtime installations
Using the 'binary' content type feature (see issue #531 ), this enhancement will add an analyzer that can detect python runtime installations that are not installed using 'os' package (rpm, dpkg, apkg, etc) install methods.  For example, many of the python:* images in dockerhub include several python runtimes, some installed via the base OS package, but the default is installed explicitly outside the underlying OS package manager.

Python runtimes installed in this way will match using anchore's CPE matching, which reference NVD and other CPE based vulnerability feeds.

### LINKED ISSUE #534: Add ability to detect and match vulnerabilities against non-os golang runtime installations
Using the 'binary' content type feature (see issue #531 ), this enhancement will add an analyzer that can detect golang runtime installations that are not installed using 'os' package (rpm, dpkg, apkg, etc) install methods.  For example, the golang:* images in dockerhub include the golang runtime that matches the tag version, installed expicitly in a location not maintained by any underlying OS package manager.

golang runtimes installed in this way will match using anchore's CPE matching, which reference NVD and other CPE based vulnerability feeds.

nurmi:
…python and golang runtimes that are installed outside underlying os package manager. Fixes #531. Fixes #533. Fixes #534. Fixes #315 .

Signed-off-by: Daniel Nurmi <nurmi@anchore.com>


nurmi:
@zhill @nightfurys - latest push has added a busybox package/version detection to the new binary analyzer in addition to the golang and python detection (appropriate spot to address an older ticket #340 )

anchore_engine/analyzers/modules/32_python_packages.py
```diff
@@ -25,6 +25,9 @@
 unpackdir = config['dirs']['unpackdir']
 squashtar = os.path.join(unpackdir, "squashed.tar")
 
+# known modules that generate false positives since they are part of core Python (2) runtime, for which vulnerabilities are handled separately
+omit_pymods = ['Python']
```
zhill, line 29: Can you add a 'fixes...' to the PR merge request for the issues this resolves? I think it will be good to call out in the changelog.

nurmi: done!

PR #545, opened 2020-07-31
Increase test coverage for common registry auth functions

### LINKED ISSUE #544: Increase test coverage for common registry auth functions
Increase unit test coverage for common registry authentication functions.

Test coverage for `anchore_engine/auth/common.py` at **17%** currently.

luhring:
Fixes #544 

Increases unit test coverage to 100% for `anchore_engine/auth/common.py`


tests/unit/anchore_engine/auth/test_common.py
```diff
@@ -0,0 +1,130 @@
...
+        'registry_type': 'awsecr',
+        'registry_meta': 'this-is-not-valid-json!}',
+    }
+    record_not_ecr = {
+        'registry_type': 'other-registry',
+        'registry_user': _test_username,
+        'registry_pass': _test_password,
+    }
+
+    test_cases = [
```
zhill, line 27: pytest has some nice annotations for parameterizing a test, useful in the future but non-blocking here. See as an example: https://github.com/anchore/anchore-engine/blob/master/tests/unit/anchore_engine/services/test_api_specs.py#L26

luhring: Thanks @zhill — I went ahead and reworked this a bit, since it was a good excuse for me to learn more Python idioms. I checked out the example you mentioned, as well as the pytest docs on [parameterization](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) and [expected exceptions](https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions).

zhill: Great, looks good!

PR #549, opened 2020-08-01
Adds malware scanning via ClamAV

### LINKED ISSUE #498: Support malware scanning of image content
As a user of Engine I would like to be able to scan my image content for malware so that I have additional checks for malicious components as well as vulnerable components of an image when I am consuming images that I don't have a known or trusted provenance.

Initial implementation to support ClamAV as the scanner to scan the squashed tar output of the analyzer and reporting simple signature checks.

Will add API route to query results:
`GET /v1/images/<digest>/content/scans` that return a payload roughly like:
```
[
  {
    "scanner": "clamav", 
     "findings": [
       {"path": "/some/file/with/badness", finding": "<some malware signature key>"},
       ...
      ]
  }
]
```

Suggested policy gate and trigger for matches:

Gate = 'external_scans', trigger = 'findings',
Params:
'scanner', value of which scanner name to filter on, default is no filter and trigger on all scanners and all findings.
'filter_path_regex', value is a regex to filter path matches by so trigger fires only findings that fit the regext
'exclude_path_regex', value is a regex filter to exclude matches from
'match_filter', value is the signature name to fire on if found. Default is to fire on all matches unless this filter is applied.

TriggerIDs (used for whitelisting) could be formatted as `<finding-value>+<path>`. This is subject to change, as it needs additional consideration, but provides an example.

zhill:
Adds a new malware analyzer, gate, and content APIs. This scan is disabled by default since it adds considerable (minutes) to the analysis time.

Resolves #498


zhill:
Not quite ready: remaining work are tweaks to run the scan better as non-root user, and config where to consume temp space during the scan so that clamav doesn't fill /tmp inside a container for a large image being analyzed.

Issue #193, opened 2019-05-14
Image query filters


zhill:
Allow search/filter criteria for images in the GET /images route that includes annotations, registry, and repository for better query granularity. Support these in query params with optional list types:

GET /images?registry=docker.io&annotations[key1]=value&annotations[key2]=value2

The exact serialization of annotations parameter is tbd based on swagger support, but the objective is to support querying arbitrary sets of annotations.

zhill:
Current implementation includes image_status and analysis_status parameters. Splitting this into other issues to track the annotations filter: #561

PR #553, opened 2020-08-03
Adds image_status and analysis_status query params for filtering image listings

zhill:
Implements the status-centric part of #193


nightfurys:
lgtm

PR #555, opened 2020-08-05
Update API response for image deletion by image ID

nightfurys:
@zhill, missed this image deletion API before (never used it). Its a minor update to the API, verified the change works

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -483,7 +483,7 @@ paths:
         200:
           description: "successful operation"
           schema:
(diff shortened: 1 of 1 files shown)


anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -601,6 +601,8 @@ paths:
       responses:
         200:
           description: Image deletion success
+          schema:
+            $ref: "#/definitions/DeleteImageResponse"
```
nightfurys, line 605: > This looks like just a naming update.

And defining a response object schema for `DELETE /image/by_id/{imageId}`

PR #564, opened 2020-08-06
Deduplicate multiple copies of parse_dockerimage_string

0xDEC0DE:
Rather than maintaining two slightly-out-of-sync copies of the same function that work slightly differently, just pull in one of the copies into the other module's namespace, and make it able to handle both use cases.

This adds the kwarg `strict` to `parse_dockerimage_string`, which if set to `False`, will NOT throw exceptions on invalid input characters (i.e., so the "shape checking" use case of the policy engine can use it)

Also provide some code cleanups, e.g., using raw-strings for regex, tuples instead of lists, etc.

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/bundles.py
+++ b/anchore_engine/services/policy_engine/engine/policy/bundles.py
@@ -696,7 +696,7 @@ class MappingRule(object):
         if not tag:
             raise ValueError('Tag cannot be empty or null for matching evaluation')
         else:
-            match_target = parse_dockerimage_string(tag)
+            match_target = parse_dockerimage_string(tag, strict=False)
 
         # Must match registry and repo first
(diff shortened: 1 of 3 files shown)


zhill:
Thanks @0xDEC0DE. Looks like the changes have broken some consumers of the parser that use wildcards and are failing your bad-chars check.

You can run tests locally with: `make test-unit` to see the failures. I think the failure is showing a place where that function should probably not be used since the place it's failing is in some wildcard checking code, so perhaps a flag for 'allow-wildcards' or splitting into another function that is used in the other location but doesn't do a strict check is a better solution. The usage currently expects the wildcards to be parsed correctly based on the structure, which is a slightly different check.

0xDEC0DE:
Thanks, I was able to get the tests passing, a new revision has been force-pushed.

PR #567, opened 2020-08-08
specify timestamp override as current time or previous image created_…

### LINKED ISSUE #552: Do not always require a created_timestamp_override field for analysis by digest
If no override is provided, the API should set the current time or use the timestamp from an existing record if the force flag is set and there is an existing record (e.g. in analysis_failed).

dakaneye:
…at value if force is specified

When an image is submitted for analysis by digest & tag, right now it will throw an error if the created_at timestamp is not specified.

Now, if it's missing, we will default it to the current time epoch, and if the force parameter is specified, we will try to retrieve it from the previous image's value. If it still doesn't exist on the previous image, we will set it to the current time epoch

**Which issue this PR fixes** : fixes #552

**Special notes**:
Manual Functional test below to verify it is working. As you can see, no creation_timestamp_override is specified and therefore, one is set to the current time (which was 1:58 pm)
```zsh
samdacanay@samsanchorembp ~ % curl -X POST -H "Authorization: Basic YWRtaW46Zm9vYmFy" -H "Content-Type: application/json" -d '{"source": {"digest": {"pullstring": "swaggerapi/swagger-ui@sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e","tag": "docker.io/swaggerapi/swagger-ui:latest"}}}' "http://localhost:8228/v1/images"
[
  {
    "analysis_status": "not_analyzed",
    "analyzed_at": null,
    "annotations": {},
    "created_at": "2020-08-07T20:58:13Z",
    "imageDigest": "sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e",
    "image_content": {
      "metadata": {
        "arch": null,
        "distro": null,
        "distro_version": null,
        "dockerfile_mode": null,
        "image_size": null,
        "layer_count": null
      }
    },
    "image_detail": [
      {
        "created_at": "2020-08-07T20:58:13Z",
        "digest": "sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e",
        "dockerfile": null,
        "fulldigest": "docker.io/swaggerapi/swagger-ui@sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e",
        "fulltag": "docker.io/swaggerapi/swagger-ui:latest",
        "imageDigest": "sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e",
        "imageId": "af194872176ffeadbd6f0d2918372c0f8e49406c4cb7e64e450957e00ddbbe63",
        "last_updated": "2020-08-07T20:58:14Z",
        "registry": "docker.io",
        "repo": "swaggerapi/swagger-ui",
        "tag": "latest",
        "tag_detected_at": "2020-08-07T20:58:14Z",
        "userId": "admin"
      }
    ],
    "image_status": "active",
    "image_type": "docker",
    "last_updated": "2020-08-07T20:58:14Z",
    "parentDigest": "sha256:f75b398a3a3d0b4ffa03693a001af52367b13ec55b61e7fbebe078cbd69a060e",
    "userId": "admin"
  }
]
```

### DIFF
--- a/anchore_engine/services/apiext/api/controllers/images.py
+++ b/anchore_engine/services/apiext/api/controllers/images.py
@@ -1163,7 +1163,7 @@ def analyze_image(account, source, force=False, enable_subscriptions=None, annot
                     if not ts:
                         # Timestamp required for analysis by digest & tag (if none specified,
                         # default to previous image's timestamp)
-                        ts = image_check['created_at']
+                        ts = image_check.get('created_at', utils.datetime_to_epoch(datetime.datetime.now()))
                 except Exception as err:
                     raise ValueError("image digest must already exist to force re-analyze using tag+digest")
             elif not ts:
--- a/anchore_engine/services/apiext/api/controllers/images.py
+++ b/anchore_engine/services/apiext/api/controllers/images.py
@@ -1,6 +1,7 @@
 import json
 import stat
 import datetime
+import time
 import base64
 import re
 import tarfile
@@ -16,7 +17,8 @@ from anchore_engine.clients.services.policy_engine import PolicyEngineClient
 from anchore_engine.clients.services.catalog import CatalogClient
 from anchore_engine.clients.services import internal_client_for
 import anchore_engine.common
-from anchore_engine.common.helpers import make_response_error, make_anchore_exception, make_eval_record, make_policy_record, make_response_routes
+from anchore_engine.common.helpers import make_response_error, make_anchore_exception
+from anchore_engine.db.entities.common import anchore_now
 import anchore_engine.common.images
 import anchore_engine.configuration.localconfig
 from anchore_engine.subsys import taskstate, logger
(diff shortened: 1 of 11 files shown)


zhill:
Thanks @dakaneye I'll take a look!

anchore_engine/services/apiext/api/controllers/images.py
```diff
@@ -1160,11 +1160,16 @@ def analyze_image(account, source, force=False, enable_subscriptions=None, annot
                     image_check = client.get_image(digest)
                     if not image_check:
                         raise Exception('No image found for digest {}'.format(digest))
+                    if not ts:
+                        # Timestamp required for analysis by digest & tag (if none specified,
+                        # default to previous image's timestamp)
+                        ts = image_check.get('created_at', utils.datetime_to_epoch(datetime.datetime.now()))
```
zhill, line 1166: time.time() does the same thing, obviating the need for the custom function to convert time formats. But also, be sure that it's using UTC time not the local time on the host.

anchore_engine/services/apiext/api/controllers/images.py
```diff
@@ -1160,11 +1160,16 @@ def analyze_image(account, source, force=False, enable_subscriptions=None, annot
...
                 except Exception as err:
                     raise ValueError("image digest must already exist to force re-analyze using tag+digest")
             elif not ts:
-                # If a new analysis of an image by digest + tag, we need a timestamp to insert into the tag history properly
-                raise ValueError("must supply creation_timestamp_override when adding a new image by tag+digest")
+                # If a new analysis of an image by digest + tag, we need a timestamp to insert into the tag history
+                # properly. Therefore, if no timestamp is provided, we use the current time
+                ts = utils.datetime_to_epoch(datetime.datetime.now())
```
zhill, line 1172: Same here, can use time.time()

anchore_engine/services/apiext/api/controllers/images.py
zhill: Also see: https://github.com/anchore/anchore-engine/blob/master/anchore_engine/db/entities/common.py#L78

Could probably hoist that into the utils module but that can be in different commit/PR since that impacts a lot of other code. But the above links shows how we set the epochs in the image timestamps.

dakaneye: Nice, didn't see the other common method on the first pass. Will use the same method (**anchore_now()**) in this change, then we can overhaul it into utils later.

PR #568, opened 2020-08-08
Dont allow admin account disablement fix 522

### LINKED ISSUE #522: Prevent admin account from being disabled by API
The admin account should not be able to be disabled since it cannot be re-enabled at that point.

Add a check in the account state update to prevent the admin account state toggle.

dakaneye:

This PR addresses the issue where it is technically possible to disable the admin account, yet hard to re-enable afterwards

**Which issue this PR fixes**: fixes  #522

**Special notes**:
Verified fix works locally, with following result:
```zsh
samdacanay@samsanchorembp forks % anchore-cli account disable admin
Error: Cannot disable the admin account
HTTP Code: 400
Detail: {u'error_codes': []}
```

### DIFF
--- a/anchore_engine/db/db_accounts.py
+++ b/anchore_engine/db/db_accounts.py
@@ -70,7 +70,7 @@ def update_state(name, new_state, session=None):
     if accnt.state == AccountStates.deleting or (accnt.state == AccountStates.enabled and new_state == AccountStates.deleting):
         raise InvalidStateError(accnt.state, new_state)
 
-    # Both Account Name and Type should be admin, but just to be safe...
+    # Both Account Name and Type should be equal to "admin" for the Admin Account, but just to be safe...
     if (accnt.name == 'admin' or accnt.type == 'admin') and new_state != AccountStates.enabled:
         raise DisableAdminAccountError()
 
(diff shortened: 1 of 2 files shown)


zhill:
Thanks @dakaneye! I'll review early next week.

anchore_engine/db/db_accounts.py
```diff
@@ -65,6 +70,10 @@ def update_state(name, new_state, session=None):
     if accnt.state == AccountStates.deleting or (accnt.state == AccountStates.enabled and new_state == AccountStates.deleting):
         raise InvalidStateError(accnt.state, new_state)
 
+    # Both Account Name and Type should be equal to "admin" for the Admin Account, but just to be safe...
+    if (accnt.name == 'admin' or accnt.type == 'admin') and new_state != AccountStates.enabled:
```
zhill, line 74: There is a constant and an enum (AccountTypes.admin) for the admin account name and the admin account type respectively. Please use those so that any future updates are applied consistently.

See: https://github.com/anchore/anchore-engine/blob/master/anchore_engine/configuration/localconfig.py#L68 for the constant.

dakaneye: Nice! Fixed.

Issue #489, opened 2020-06-17
Add a webhook test capability to verify configuration and test handlers


zhill:
When developing webhook handlers it is very helpful to have a clear way to send a test event for each specific payload type to test the handling logic end-to-end.

Engine should provide a simple API call to send a test webhook of a specific type.

zhill:
Closed by #570

PR #569, opened 2020-08-09
Expose possible webhook schemas, validate them as they are sent, and allow them to be tested

### LINKED ISSUE #490: Provide json schemas for webhook payloads
Each webhook payload should have a jsonschema available in the codebase to provide documentation, validation, and guidance to developers writing consumers for the webhooks.

dakaneye:
…) for a notification from the queues table, and can be accessed via a new endpoint. Also add some method comments

**What this PR does / why we need it**: Proposes a solution for #490 where the **genson** package is used to create a JSON schema from an example notification json (should be the same payload as what is used in actual webhook notifications), and expose that schema via a new system endpoint: /system/webhook_schema

**Which issue this PR fixes**: fixes #490 

**Special notes**:
Endpoint View & Test from Swagger UI (endpoint has since been updated to /webhooks/schema). I decided to structure it this way due to the fact that this change dovetails with #489 and it will make sense to have another endpoint to perform the test
image


dakaneye:
This needs some more work, so we can make sure to have a comprehensive list of the different JSON Schemas that are possible. Actually, it appears there could be a different schema for each [event type](anchore_engine/subsys/events/types.py)

zhill:
Interesting idea @dakaneye , though I'm not sure an API is necessary for this unless it is dynamic based on runtime configuration, which it shouldn't be. I'd actually prefer to make the schemas more rigid internally so that inputs for the notification queue and just document that schema via the swagger.json. There can still be an API involved to test the webhooks. I think there is a lot of value in a `POST /system/webhooks/<type>/test` capability to help test integrations, so that could also have a GET option for the schema as well but not a hard requirement.

dakaneye:
Going to close this for now as I'll be in an intermediate state working things through for a bit.

anchore_engine/service.py
```diff
@@ -546,8 +546,8 @@ def _init_wsgi_app(self, service_name, api_spec_dir=None, api_spec_file=None):
 
             return self._application
         except Exception as err:
-            traceback.print_exc()
-            raise err
+            logger.exception('Error initializing WSGI application')
```
zhill, line 549: Glad to get that statement replaced with a proper log call

anchore_engine/services/apiext/api/controllers/system.py
```diff
@@ -376,3 +378,101 @@ def describe_error_codes():
         httpcode = return_object['httpcode']
 
     return return_object, httpcode
+
+
+@authorizer.requires([])
```
zhill, line 383: I think only the admin should be able to run this command, since it creates output to an external system and it's the admin that configures the webhooks in the config.yaml.

You can enforce that with:
`@authorizer.requires_account(with_names=[localconfig.ADMIN_ACCOUNT_NAME])`

That will make it only authorized for users that are in the 'admin' account.

dakaneye: fixed

PR #570, opened 2020-08-12
Add webhook test capability

### LINKED ISSUE #489: Add a webhook test capability to verify configuration and test handlers

dakaneye:
Users may configure webhooks in anchore_engine so that they may be notified of events within the product, ranging from subscription changes, policy changes, image analysis updates, feed syncing, etc. However, it's useful to be able to test that the webhook is working when it's being setup (since the HTTP request details come from the config / environment vars)
This endpoint allows the user to select a webhook by type and attempt delivery of a test payload. 

**Which issue this PR fixes**: fixes #489

**Special notes**:
Tested manually with a requestbin configured:
image
image


anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -2585,6 +2585,32 @@ paths:
           description: Internal error
           schema:
             $ref: "#/definitions/ApiErrorResponse"
+  /system/webhooks/{webhook_type}/test:
```
nightfurys, line 2588: Hey @dakaneye, good to see so many PRs and fixes! Quick question about this new API, is there a particular reason it's nested under `/system`? I generally think of that path as ops internal to the system such as services, events etc. I think webhook is user facing enough to merit its own endpoint, worth the discussion

dakaneye: Hey @nightfurys ! I originally based the design as you had mentioned: `/webhooks/test/{webhook_type}`, but I updated it based on the comment that @zhill made [here](https://github.com/anchore/anchore-engine/pull/569). I'm happy to make it it's own base path though, depending on the team's preference. I think with the other enhancement issue, to expose a schema, an API may be a good way. But also, as Zach commented, the schema of a webhook shouldn't be changing statically, so it may be best to expose it as a swagger definition... 

Let me know what sounds best!

anchore_engine/service.py
```diff
@@ -546,6 +546,7 @@ def _init_wsgi_app(self, service_name, api_spec_dir=None, api_spec_file=None):
 
             return self._application
         except Exception as err:
+            logger.exception('Error initializing WSGI application: {}'.format(err))
```
alfredodeza, line 549: `logger.exception()` already captures the traceback and error. It is common to add an error description but not use the `err` variable again since it would be repeated a few lines later.
Since all that is going on, the `traceback.print_exec()` can be removed, and `raise err` can be swapped for `raise`

dakaneye: Nice. Great learning moment for me with python logging and error handling.

dakaneye:
Just want to point out before anyone does a deep dive review that there are changes in the schema PR: https://github.com/anchore/anchore-engine/pull/569 which refactor/change a lot of what is going on here...

@zhill if you want, I can just combine the requests to reduce confusion.

dakaneye:
no longer valid

PR #593, opened 2020-08-28
Fix exception response handler and create dedicated exception for image conflict errors

### LINKED ISSUE #592: Analysis Archive Restore Image throws incorrect HTTP Response code

---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): Bug Report

```
samdacanay@SDACANAY ~ % a --version
anchore-cli, version 0.7.2
```

**What happened**: Tried to restore an image from the archive, but the image hadn't yet been deleted from the engine context. This was due to the fact that the image was archived via the CLI as well (which doesn't automatically delete the image from the engine context.

**What did you expect to happen**:  Originally expected that the restore would work (not realizing the image had to be manually deleted outside of the archive flow). Now, just expect a different HTTP response code. Instead of 500, should probably be a 409

**Anchore CLI terminal dump**:
```
samdacanay@SDACANAY ~ % a analysis-archive images add sha256:08ca4f2c45316cb0b79faf8f07625046dbbc16f6f140b0c59da03f1323ada8a4
Image Digest                                                                   Archive Status        Details
sha256:08ca4f2c45316cb0b79faf8f07625046dbbc16f6f140b0c59da03f1323ada8a4        archived              Completed successfully
samdacanay@SDACANAY ~ % a analysis-archive images restore sha256:08ca4f2c45316cb0b79faf8f07625046dbbc16f6f140b0c59da03f1323ada8a4
Error: failed post url=http://localdev-anchore-engine-catalog:8082/v1/images
HTTP Code: 500
Detail: {u'error_codes': []}
samdacanay@SDACANAY ~ % a --debug analysis-archive images restore sha256:08ca4f2c45316cb0b79faf8f07625046dbbc16f6f140b0c59da03f1323ada8a4
DEBUG:anchorecli.clients.apiexternal:As Account = None
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): localhost:8228
DEBUG:urllib3.connectionpool:http://localhost:8228 "GET /v1 HTTP/1.1" 200 5
DEBUG:anchorecli.clients.apiexternal:As Account = None
DEBUG:anchorecli.clients.apiexternal:POST url=http://localhost:8228/v1/images
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): localhost:8228
DEBUG:urllib3.connectionpool:http://localhost:8228 "POST /v1/images HTTP/1.1" 500 149
DEBUG:anchorecli.cli.utils:fetched httpcode from response: 500
Error: failed post url=http://localdev-anchore-engine-catalog:8082/v1/images
HTTP Code: 500
Detail: {u'error_codes': []}
```

```
[service:enterprise-apiext] 2020-08-27 21:21:45+0000 [_GenericHTTPChannelProtocol,25485,127.0.0.1] [PoolThread-twisted.internet.reactor-0] [anchore_engine.clients.services.internal/dispatch()] [ERROR] Failed client call to service catalog for url: http://localdev-anchore-engine-catalog:8082/v1/images. Response: {'httpcode': 500, 'anchore_error_raw': 'b\'"Conflict: Image already exists in system. No restore possible"\\n\'', 'anchore_error_json': 'Conflict: Image already exists in system. No restore possible'}
```

**What docker images are you using**: Enterprise, 2.4.0-rc1

**How to reproduce the issue**: Archive an image, try to restore it before deleting it from engine

**Anything else we need to know**: N/A

dakaneye:

**What this PR does / why we need it**: Improves Engine's ability to parse errors into readable responses, and allows HTTP codes from catalog request to pass through to the API response should they be formatted correctly. Use a dedicated Exception for the case where there is an Image Conflict when a restore-from-archive operation happens

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #592 

**Special notes**:

### DIFF
--- a/anchore_engine/common/helpers.py
+++ b/anchore_engine/common/helpers.py
@@ -29,19 +29,24 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
 
     if isinstance(errmsg, Exception):
         if 'anchore_error_json' in errmsg.__dict__:
-            if {'message', 'httpcode', 'detail'}.issubset(set(errmsg.__dict__['anchore_error_json'])):
-                ret.update(errmsg.__dict__['anchore_error_json'])
+            # Try to load it as json
+            try:
+                err_json = json.loads(errmsg.__dict__.get('anchore_error_json', None))
+            except ValueError:
+                # Then it may just be a string, we cannot do anything with it
+                logger.debug('Failed to parse anchore_error_json as json')
+                return ret
+
+            if {'message', 'httpcode', 'detail'}.issubset(set(err_json)):
+                ret.update(err_json)
 
             try:
-                if {'error_code'}.issubset(set(errmsg.__dict__['anchore_error_json'])) and errmsg.__dict__['anchore_error_json'].get('error_code', None):
+                if {'error_code'}.issubset(set(err_json)) and err_json.get('error_code', None):
                     if 'error_codes' not in ret['detail']:
                         ret['detail']['error_codes'] = []
-                    ret['detail']['error_codes'].append(errmsg.__dict__['anchore_error_json'].get('error_code'))
-            except Exception as err:
-                try:
-                    logger.warn("unable to marshal error details: source error {}".format(errmsg.__dict__))
-                except:
-                    pass
+                    ret['detail']['error_codes'].append(err_json.get('error_code'))
+            except KeyError:
+                logger.warn("unable to marshal error details: source error {}".format(errmsg.__dict__))
     return ret
 
 
--- a/anchore_engine/services/catalog/api/controllers/default_controller.py
+++ b/anchore_engine/services/catalog/api/controllers/default_controller.py
@@ -4,6 +4,7 @@ import anchore_engine.apis
 from anchore_engine import db
 import anchore_engine.services.catalog.catalog_impl
 import anchore_engine.common
+from anchore_engine.services.catalog.archiver import ImageConflict
 from anchore_engine.subsys import logger
 import anchore_engine.configuration.localconfig
 import anchore_engine.subsys.servicestatus
@@ -95,7 +96,9 @@ def add_image(image_metadata=None, tag=None, digest=None, created_at=None, from_
 
             with db.session_scope() as session:
                 return_object, httpcode = anchore_engine.services.catalog.catalog_impl.image(session, request_inputs, bodycontent=image_metadata)
-
+    except ImageConflict as img_err:
+        httpcode = 409
+        return_object = str(img_err)
     except Exception as err:
         logger.exception('Error processing image add')
         httpcode = 500
--- a/anchore_engine/services/catalog/archiver.py
+++ b/anchore_engine/services/catalog/archiver.py
@@ -79,6 +79,20 @@ class MetadataNotFound(Exception):
     pass
 
 
+class ImageConflict(Exception):
+
+    def __init__(self, msg, detail=None):
+        self.msg = msg
+        self.detail = detail
+
+    def __str__(self):
+        return str(json.dumps({
+            'httpcode': 409,
+            'message': self.msg,
+            'detail': self.detail
+        }))
+
+
 class ObjectStoreLocation(JsonMappedMixin):
     class ObjectStoreLocationV1Schema(JitSchema):
         bucket = fields.Str()
@@ -716,7 +730,7 @@ class RestoreArchivedImageTask(object):
         with session_scope() as session:
             if db_catalog_image.get(self.image_digest, self.account, session):
                 logger.info('Image archive restore found existing image records already. Aborting restore.')
-                raise Exception('Conflict: Image already exists in system. No restore possible')
+                raise ImageConflict('Conflict: Image already exists in system. No restore possible')
 
             rec = db_archived_images.get(session, self.account, self.image_digest)
             if not rec:
@@ -844,7 +858,7 @@ class RestoreArchivedImageTaskFromArchiveTarfile(RestoreArchivedImageTask):
         with session_scope() as session:
             if db_catalog_image.get(self.image_digest, self.account, session):
                 logger.info('Image archive restore found existing image records already. Aborting restore.')
-                raise Exception('Conflict: Image already exists in system. No restore possible')
+                raise ImageConflict('Conflict: Image already exists in system. No restore possible')
 
         dest_obj_mgr = object_store.get_manager()
 


anchore_engine/common/helpers.py
```diff
@@ -29,15 +29,23 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
 
     if isinstance(errmsg, Exception):
         if 'anchore_error_json' in errmsg.__dict__:
-            if {'message', 'httpcode', 'detail'}.issubset(set(errmsg.__dict__['anchore_error_json'])):
-                ret.update(errmsg.__dict__['anchore_error_json'])
+            # Try to load it as json
+            try:
+                err_json = json.loads(errmsg.__dict__['anchore_error_json'])
+            except:
```
alfredodeza, line 35: I know that bare `try/except` blocks are pervasive around the Anchore Engine codebase, but we should be very careful on where/why we are doing these. At the very least we should log the exception if we truly don't have any idea what exception is going to be raised.

In this particular case a `json.loads()` is being called along with `errmsg`. For the JSON part, the exception to catch would be `ValueError`, the `dict` could either have a `.get()` or catch `KeyError` as well

anchore_engine/common/helpers.py
```diff
@@ -29,15 +29,23 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
...
 
             try:
-                if {'error_code'}.issubset(set(errmsg.__dict__['anchore_error_json'])) and errmsg.__dict__['anchore_error_json'].get('error_code', None):
+                if {'error_code'}.issubset(set(err_json)) and err_json.get('error_code', None):
                     if 'error_codes' not in ret['detail']:
                         ret['detail']['error_codes'] = []
-                    ret['detail']['error_codes'].append(errmsg.__dict__['anchore_error_json'].get('error_code'))
-            except Exception as err:
+                    ret['detail']['error_codes'].append(err_json.get('error_code'))
+            except:
```
alfredodeza, line 48: no bare excepts here as well...

anchore_engine/services/catalog/archiver.py
```diff
@@ -79,6 +79,24 @@ class MetadataNotFound(Exception):
...
+    def __str__(self):
+        return str(self.to_json())
+
+    def to_json(self):
+        e_dict = dict()
+        e_dict['httpcode'] = 409
+        e_dict['message'] = self.msg
+        e_dict['detail'] = self.detail
+
+        return json.dumps(e_dict)
```
alfredodeza, line 97:
```suggestion
        return json.dumps({
            'httpcode' : 409
            'message' : self.msg
            'detail' : self.detail
        })
```

anchore_engine/common/helpers.py
dakaneye: Addressed!

anchore_engine/common/helpers.py
```diff
@@ -29,19 +29,24 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
...
                     if 'error_codes' not in ret['detail']:
                         ret['detail']['error_codes'] = []
-                    ret['detail']['error_codes'].append(errmsg.__dict__['anchore_error_json'].get('error_code'))
-            except Exception as err:
-                try:
-                    logger.warn("unable to marshal error details: source error {}".format(errmsg.__dict__))
-                except:
-                    pass
+                    ret['detail']['error_codes'].append(err_json.get('error_code'))
+            except KeyError:
```
alfredodeza, line 48: yessssssss

anchore_engine/common/helpers.py
```diff
@@ -29,19 +29,24 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
 
     if isinstance(errmsg, Exception):
         if 'anchore_error_json' in errmsg.__dict__:
-            if {'message', 'httpcode', 'detail'}.issubset(set(errmsg.__dict__['anchore_error_json'])):
-                ret.update(errmsg.__dict__['anchore_error_json'])
+            # Try to load it as json
+            try:
+                err_json = json.loads(errmsg.__dict__.get('anchore_error_json', None))
```
alfredodeza, line 34: I usually prefer not to define `None` in a `.get()` because that is the default, so it reads redundant. I know other folks prefer it because they believe this is more explicit. Just want to emphasize that it isn't necessary to define `None` as the fallback.

anchore_engine/services/catalog/api/controllers/default_controller.py
```diff
@@ -98,7 +99,10 @@ def add_image(image_metadata=None, tag=None, digest=None, created_at=None, from_
 
     except Exception as err:
         logger.exception('Error processing image add')
-        httpcode = 500
+        if isinstance(err, ImageConflict):
```
alfredodeza, line 102: this could add another `try` block catching `ImageConflict` instead of inspecting the class hierarchy:
 ```suggestion
    except ImageConflict:
```

Although... because the previous `except Exception` comes first... then this would not be possible. This is yet another issue why overusing `except Exception` is not a good idea :(

File: anchore_engine/services/catalog/api/controllers/default_controller.py [hunk 7]
```diff
@@ -95,7 +96,9 @@ def add_image(image_metadata=None, tag=None, digest=None, created_at=None, from_
 
             with db.session_scope() as session:
                 return_object, httpcode = anchore_engine.services.catalog.catalog_impl.image(session, request_inputs, bodycontent=image_metadata)
-
+    except ImageConflict as img_err:
```
alfredodeza, line 99: this is a good way to deal with exceptions... first the specific ones, then the generic ones. Good changes!

alfredodeza (approved):

anchore_engine/common/helpers.py
```diff
@@ -29,19 +29,24 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
...
+            except ValueError:
```
nightfurys, line 35: @dakaneye, I'm seeing this fail in a lot of because anchore_error_json is a dict. `json.loads()` can only be used for loading a str, bytes or bytearray types. Is there a case where anchore_error_json is anything but a dict? 

Another issue here is the limited exception handling. In the case where `json.loads()` fails because the input is not one of the above types, the library throws a TypeError. That's not caught here. The end result to the user is an error handling an internal exception, not the actual error itself. This is one of the reasons why I have used broader catch clauses

zhill: This should be in its own ticket now that the change is merged.

dakaneye: Will open and fix.

PR #598, opened 2020-08-31
Add input parameters to Oauth swagger spec

### LINKED ISSUE #597: Oauth Swagger Endpoint documentation can be improved

---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): Neither? Enhancement request

```
Engine DB Version: 0.0.13
Engine Code Version: 2.4.0
```

**What happened**: Oauth Swagger endpoint doesn't provide enough information to call the API, needed to reference [documentation](https://docs.anchore.com/current/docs/engine/usage/api_usage/authentication/#bearer-tokensoauth) in order to form a request:
image

**What did you expect to happen**: It would be nice to be able to just use the swagger spec in order to know how to call the API

**Any relevant log output from /var/log/anchore**: N/A

**What docker images are you using**: anchore/enterprise-dev:latest

**How to reproduce the issue**: Open swagger UI, navigate to Oauth section

dakaneye:

**What this PR does / why we need it**: Adds the required parameters to the OAuth Token endpoint swagger spec

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #597 

**Special notes**:

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -2576,6 +2576,25 @@ paths:
       description: Request a jwt token for subsequent operations, this request is authenticated with normal HTTP auth
       consumes:
         - application/x-www-form-urlencoded
+      parameters:
+        - name: grant_type
+          in: formData
+          description: 'OAuth Grant type for token'
+          type: string
+          default: password
+        - name: username
+          in: formData
+          type: string
+          description: 'User to assign OAuth token to'
+        - name: password
+          in: formData
+          type: string
+          description: 'Password for corresponding user'
+        - name: client_id
+          in: formData
+          description: 'The type of client used for the OAuth token'
+          type: string
+          default: anonymous
       responses:
         200:
           description: Resulting JWT token


anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -2576,6 +2576,31 @@ paths:
...
+          required: true
+          type: string
+          description: 'Password for corresponding user'
+        - name: client_id
+          in: formData
+          required: true
+          description: 'The type of client used for the OAuth token'
+          type: string
+          enum:
+            - anonymous
```
nightfurys, line 2603: @dakaneye, does this have to be an enum? The service should just default to `anonymous` if the value is not supplied. I don't think API should fail validation

nightfurys (changes requested):
@dakaneye, good catch!

anchore_engine/services/apiext/swagger/swagger.yaml
dakaneye: I wasn't sure if that was required for now, the [docs](https://docs.anchore.com/current/docs/engine/usage/api_usage/authentication/#bearer-tokensoauth) seem to suggest it. I will just add it as the default value though 👍

nightfurys:
@dakaneye Another thing I missed the first time around is none of those params are strictly required. The controller pulls it off the (flask) request if they are not explicitly supplied. This is usually how the rest of the system processes auth context. So marking any of the params required breaks existing api (when used without any form params)
https://github.com/anchore/anchore-engine/blob/af7067be6e09955d05dbc79e93c26693abaa3282/anchore_engine/services/apiext/api/controllers/oauth.py#L37-L41

The swagger spec doesn't allow combining the conditionality on params such as all or none. I guess my main question is - what form params are actually required and optional. Is there really a need for username and password to be provided in the form data? As in, is there a use case for requesting a token for a different account?

anchore_engine/services/apiext/swagger/swagger.yaml
nightfurys: Defining and default value for a parameter that's required doesn't do anything. If the client does not supply the param, api validation kicks in and fails the request. Defaults are assumed when a param is optional and the value is not supplied by the client

nightfurys:
@dakaneye I know you opened the PR based off of the docs, but we can always fix the issue here and update the docs if necessary. Not trying to make this difficult, just want to be careful about breaking API changes and backwards compatibility. We can do that but just need to clearly document it and test it for regressions

dakaneye:
As discussed, we will make the parameters optional, and fix the username/password issue later on (i.e. if they're not specified)

nightfurys (approved):

PR #602, opened 2020-09-01
JSON logging support

### LINKED ISSUE #298: Feature Request: Add a structured JSON based formatter for logging
### Feature Request
**Add a structured JSON based formatter for logging**
Add structured JSON as alternative log format, which can be enabled via config/env. 
This cloud make the processing of logs (especially filtering) much easier. By mapping JSON fields, it allows a better integration in ELK or similar.

dakaneye:

**What this PR does / why we need it**: Adds JSON logging support. 

<Draft>: I've got the actual log file in JSON working flawlessly, but the stdout (captured by docker logs) is still inconsistent...:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #298 

**Special notes**:
* `spew` and `debug_exception` are not available in the standard logging library so have been moved to `debug` and `exception`, respectively
* `warn` is no longer supported, so has been moved to `warning`
* The `/var/log/anchore/anchore-<service_name>.log` file is no longer supported, stdout only
* The logic for using the log file as a watchdog has been removed
* I added logic for moving access logging to it's own file stream so that when json logging is enabled it doesn't mess up the formatting


dakaneye:
[anchore-api.log](https://github.com/anchore/anchore-engine/files/5157995/anchore-api.log)
[docker-logs-output.txt](https://github.com/anchore/anchore-engine/files/5157996/docker-logs-output.txt)

dakaneye:
[anchore-engine-api.log](https://github.com/anchore/anchore-engine/files/5159228/anchore-engine-api.log)

anchore_manager/util/logging.py
```diff
@@ -68,7 +73,9 @@ def log_config(config: dict):
         if config['debug']:
             log_level = 'DEBUG'
 
-        logger.configure_logging(log_level, log_to_stdout=True)
+        enable_json_logging = os.getenv('ANCHORE_JSON_LOGGING_ENABLED') == 'true'
```
alfredodeza, line 76: I understand that the defaults will set this to `"false"`, but it would be safer to fallback to an empty string
```suggestion
        enable_json_logging = os.getenv('ANCHORE_JSON_LOGGING_ENABLED', "") == 'true'
```

anchore_engine/twisted.py
```diff
@@ -269,7 +270,7 @@ def _build_api_service(self):
         root = rewrite.RewriterResource(self.root_resource, self._default_version_rewrite)
 
         # Build the main site server
-        site = server.Site(root)
+        site = server.Site(root, logPath=b'/var/log/anchore/access.log')
```
alfredodeza, line 273: woah, this needs to be bytes?

anchore_manager/cli/service.py
```diff
@@ -415,27 +354,16 @@ def start(services, no_auto_upgrade, anchore_module, skip_config_validate, skip_
...
+                while True:
+                    time.sleep(1)
+                    if localconfig.get('auto_restart_services', False):
+                        for service_thread in keepalive_threads:
+                            if not service_thread.thread.is_alive():
+                                logger.info("restarting service: {}".format(service_thread.thread.name))
+                                service_thread.start()
             except Exception as err:
-                logger.error("failed to startup log watchers - exception: {}".format(str(err)))
+                logger.error("Exception received during handling of auto restarting services: {}".format(str(err)))
```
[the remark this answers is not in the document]
dakaneye: It does, good catch. I was just rewriting the message but didn't realize! will fix.

anchore_engine/twisted.py
zhill: What's the purpose of this entry? Does this now split the output for access and info/debug logs?

dakaneye: @zhill yes, it separates those logs out so that it doesn't mix into the normal output of the logs. Twisted wasn't capturing these in it's normal Log Observers so this is the only way to not screw up the JSON logging piece. I could change this logic to only split it out if JSON logging is enabled, if you prefer?

zhill: Ack. I'd like a configuration option for this to allow the user to specify if they want a unified log stream or a split out model.

dakaneye: @zhill I think my assumption above was incorrect. Access logging doesn't seem to mess up the JSON logs (May have been a relic of some other local issues I had during development). I noticed that enabling/disabling access logging wasn't even working so I removed that for now.

anchore_manager/util/logging.py
```diff
@@ -68,7 +73,9 @@ def log_config(config: dict):
         if config['debug']:
             log_level = 'DEBUG'
 
-        logger.set_log_level(log_level, log_to_stdout=True)
+        enable_json_logging = os.getenv('ANCHORE_JSON_LOGGING_ENABLED', '') == 'true'
```
zhill, line 76: Is this for handling bootstrap prior to log setup? Generally, the use of env vars is for the substitution in the config file not directly from the environment in case it conflicts with the config file.

dakaneye: Yes it was for the bootstrap prior!

anchore_engine/twisted.py
```diff
@@ -346,29 +343,15 @@ def _build_api_service(self):
         )
 
         # Build the main site server
-        site = server.Site(root)
+        twistd_logfile = self.service_config.get(
+            "twistd_logfile", self.global_configuration.get("twistd_logfile", None)
+        )
+        if twistd_logfile:
+            site = server.Site(root, logPath=bytes(twistd_logfile, "utf-8"))
```
dakaneye, line 350: might be able to unify streams using a log formatter, should look into https://twistedmatrix.com/documents/15.3.0/web/howto/web-in-60/access-logging.html

dakaneye:
Might need to look into maintaining backwards compatibility w/ having a log file (in addition to stdout) and have a config that allows it to be stdout-only
* support might have debugging tools around checking the current log file
* can affect anchore deployment in ECS

Getting rid of spew and debug_exception may warrant some further investigation that is decoupled from this change. (filtering debug level by component, etc)

Vijay-P:
@dakaneye is this PR a draft? Can you mark it as such? Thanks!

frittentheke:
@dakaneye  did you actually want to close this?

PR #607, opened 2020-09-03
update dockerfile to use python38 from python36. Fixes #605.

### LINKED ISSUE #605: Update Dockerfile to move to python 3.8
In current (0.8.X) series, engine is using python36 environment from UBI8.  Noting here that the python36 series is set to EOL toward the end of 2021

https://www.python.org/dev/peps/pep-0494/#lifespan

Updates to the Dockerfile are minimal (python, pip, wheel) should be updated to move to python38

nurmi:
Signed-off-by: Daniel Nurmi <nurmi@anchore.com>


zhill (approved):
lgtm

PR #618, opened 2020-09-14
Additional archival rules

dakaneye:
**What this PR does / why we need it**: Allows archive rules to have an exclude block with its own expiration logic

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #: #699 #700 

**Special notes**:

### DIFF
--- a/anchore_engine/db/db_catalog_image_docker.py
+++ b/anchore_engine/db/db_catalog_image_docker.py
@@ -227,17 +227,6 @@ def get_tag_histories(session, userId, registries=None, repositories=None, tags=
     :param tags:
     :return: constructed query to execute/iterate over that returns tuples of (CatalogImageDocker, CatalogImage) that match userId/account and digest
     """
-
-    # select_fields = [
-    #     CatalogImageDocker.userId,
-    #     CatalogImageDocker.imageDigest,
-    #     CatalogImageDocker.registry,
-    #     CatalogImageDocker.repo,
-    #     CatalogImageDocker.tag,
-    #     CatalogImageDocker.tag_detected_at,
-    #     CatalogImage.analyzed_at
-    #     ]
-
     select_fields = [CatalogImageDocker, CatalogImage]
 
     order_by_fields = [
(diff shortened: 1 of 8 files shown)


anchore_engine/db/db_catalog_image_docker.py
```diff
@@ -158,29 +147,34 @@ def get_tag_histories(session, userId, registries=None, repositories=None, tags=
...
-            if exact:
-                conditions.append(field.in_(exact))
-
-            if conditions:
-                qry = qry.filter(or_(*conditions))
+    query_tuples = [(CatalogImageDocker.registry, registries),
+                    (CatalogImageDocker.repo, repositories),
+                    (CatalogImageDocker.tag, tags)]
+    for field, filters, exclude in query_tuples:
+        if not filters:
```
zhill, line 154: I'm not quite sure this code is doing the right thing. The iterator in the for loop should iterate over each of the tuples, but it looks like its expecting 3 values instead of 2. Is the 'exclude' necessary here? It doesn't look like it's used at all in the loop.

zhill (changes requested):
A couple of questions but overall looks good.

anchore_engine/db/db_catalog_image_docker.py
dakaneye: You were right, it was left over when I was mulling over my strategy for applying the exclusion. I've removed it now.

dakaneye:
Forgot to add the DB upgrade part, so just pushed that change:
```
[MainThread] [anchore_manager.cli.service/start()] [INFO] Performing upgrade.
'creating new table columns'
'creating new column (exclude_selector_registry) in table (catalog_archive_transition_rules)'
'creating new column (exclude_selector_repository) in table (catalog_archive_transition_rules)'
'creating new column (exclude_selector_tag) in table (catalog_archive_transition_rules)'
'creating new column (exclude_expiration_days) in table (catalog_archive_transition_rules)'
Detected anchore-engine version 0.0.14, running DB version 0.0.13.
Performing upgrade.
```
It worked!

zhill (approved):
lgtm

PR #623, opened 2020-09-16
Global max images

dakaneye:
POC for how we can approach the solution asynchronously.

### DIFF
--- a/anchore_engine/db/db_catalog_image.py
+++ b/anchore_engine/db/db_catalog_image.py
@@ -1,6 +1,4 @@
-from sqlalchemy import desc, and_, or_
-from sqlalchemy.orm import load_only, Load
-import time
+from sqlalchemy import desc, and_, or_, func
 
 from anchore_engine import db
 import anchore_engine.db.db_catalog_image_docker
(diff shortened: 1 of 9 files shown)


anchore_engine/services/catalog/archiver.py
```diff
@@ -1168,20 +1168,30 @@ def archive_vuln_history(self, img_archive: ImageArchive) -> list:
...
-    #
-    # def flush_policy_engine(self, image_id):
-    #     try:
-    #         pe_client = internal_client_for(PolicyEngineClient, userId=self.account)
-    #         pe_client.delete_image(self.account, image_id)
-    #     except Exception as ex:
-    #         logger.exception("Error flushing policy engine state for image")
-    #         raise ex
+
+class MaxImageAnalysesArchiver(object):
```
dakaneye, line 1172: This could just run in a monitor thread in the Catalog service or something if we decide to go async. 

Otherwise, if synchronous we can just fire this off when an image is added

anchore_engine/db/db_catalog_image.py
```diff
@@ -353,26 +351,31 @@ def delete(imageDigest, userId, session=None):
...
+    if max_images < 0:
+        return
+
+    image_count = session.query(func.count()).order_by(None).scalar()
+    limit = image_count - max_images
+
+    if limit <= 0:
+        return
+
+    qry = session.query([CatalogImage.imageDigest])\
```
zhill, line 377: Need to filter this by image status as well so you don't get analysis failed or analyzing images (depends on the input params that the user provides in the rule, but I think this should just be for 'analyzed' images.

conf/default_config.yaml
```diff
@@ -10,7 +10,6 @@
 service_dir: ${ANCHORE_SERVICE_DIR}
 tmp_dir: /analysis_scratch
 log_level: ${ANCHORE_LOG_LEVEL}
-cleanup_images: true
```
zhill, line 13: Why is this removed?

dakaneye: I couldn't find any usages in the application, unless I'm missing something.

conf/default_config.yaml
```diff
@@ -42,6 +41,15 @@ user_authentication:
   # WARNING: you should not change this after a system has been initialized as it may cause a mismatch in existing passwords
   hashed_passwords: ${ANCHORE_AUTH_ENABLE_HASHED_PASSWORDS}
 
+# This block is for defining archival rules around the maximum number of image analyses in the Anchore Deployment
```
zhill, line 44: These rules are all configured via API calls, not configuration unless we extend the configuration to support any kind of rule. I don't want to split them up.

dakaneye: Moved into rule schema

conf/default_config.yaml
dakaneye: lmk if youw ant me to remove this change

zhill: This is good. Can leave it.

zhill:
This looks good to merge into the other branch so we can consider it in-toto.

PR #626, opened 2020-09-19
Draft: PoC multi-arch support

nightfurys:

**Special notes**:

### DIFF
--- a/anchore_engine/clients/skopeo_wrapper.py
+++ b/anchore_engine/clients/skopeo_wrapper.py
@@ -4,7 +4,7 @@ import re
 import tempfile
 
 import anchore_engine.configuration.localconfig
-from anchore_engine.utils import run_command, run_command_list, manifest_to_digest, AnchoreException
+from anchore_engine.utils import run_command, run_command_list, manifest_to_digest, AnchoreException, ManifestDigestArch, ImageInfo
 from anchore_engine.subsys import logger
 from anchore_engine.common.errors import AnchoreError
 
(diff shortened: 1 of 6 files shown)


anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -331,6 +331,135 @@ def get_image_manifest_skopeo(url, registry, repo, intag=None, indigest=None, to
...
+        raw_manifest = get_image_manifest_skopeo_raw(pull_string, user=user, pw=pw, verify=verify)
+        s_manifest = json.loads(raw_manifest)
+        s_digest = manifest_to_digest(raw_manifest)
+        return s_manifest, s_digest
+    except Exception as err:
+        logger.warn("CMD failed - exception: " + str(err))
+        raise err
+
+
+def get_image_manifest_v2(url, registry, repo, intag=None, indigest=None, user=None, pw=None, verify=True, default=None):
```
Toure, line 422: default should be set to False as it is a bool.

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -331,6 +331,135 @@ def get_image_manifest_skopeo(url, registry, repo, intag=None, indigest=None, to
...
+                    sout = str(sout, 'utf-8') if sout else None
+                    ret = sout
+                    break
+
+            if not success:
+                logger.error("could not retrieve manifest")
+                raise Exception("could not retrieve manifest")
+
+        except Exception as err:
+            raise err
```
alfredodeza, line 398: This doesn't have any effect and it is equal to not having it at all

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -331,6 +331,135 @@ def get_image_manifest_skopeo(url, registry, repo, intag=None, indigest=None, to
...
+    else:
+        raise Exception("invalid input - must supply either an intag or indigest")
+
+    try:
+        raw_manifest = get_image_manifest_skopeo_raw(pull_string, user=user, pw=pw, verify=verify)
+        s_manifest = json.loads(raw_manifest)
+        s_digest = manifest_to_digest(raw_manifest)
+        return s_manifest, s_digest
+    except Exception as err:
+        logger.warn("CMD failed - exception: " + str(err))
```
alfredodeza, line 418: this doesn't add anything that the Exception wouldn't already show. Does this imply there is something beyond `str(err)` that is useful to log? It seems this could do well without a `try/except` block

PR #637, opened 2020-10-03
Update repo watcher asynchronous process. DB lookup subscription stat…

### LINKED ISSUE #635: Deactivating repository subscription does not halt scanning tags
Watching a repository is handled by an asynchronous catalog process periodically performing inventory of repository and queuing up images for analysis. This process does an atomic check for the subscription state before evaluating each tag manifest in a serial order. This is not very fast process which makes it seem like the tags are trickling into the system. Deactivating or even deleting the watch/subscription record does not halt this stream as the subscription state check is performed once at the top of the process

nightfurys:
…e before queuing image for analysis for every image

Fixes #635

**Special notes**:

### DIFF
--- a/anchore_engine/db/db_subscriptions.py
+++ b/anchore_engine/db/db_subscriptions.py
@@ -101,6 +101,19 @@ def get(userId, subscription_id, session=None):
     return ret
 
 
+def is_active(account, subscription_id, session=None):
+    """
+    Returns the subscription id of the record if one exists for the account and subscription id
+    """
+
+    if not session:
+        session = db.Session
+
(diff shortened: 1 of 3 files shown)


anchore_engine/db/db_subscriptions.py
```diff
@@ -133,10 +146,24 @@ def get_bysubscription_key(userId, subscription_key, session=None):
     return ret
 
 
-def update(userId, subscription_key, subscription_type, inobj, session=None):
+def upsert(userId, subscription_key, subscription_type, inobj, session=None):
```
zhill, line 149: Yes, this is more clear on the actual behavior of this function.

zhill (approved):
lgtm

PR #640, opened 2020-10-05
Fix for GHSA matching and multiple fix versions

### LINKED ISSUE #633: npm packages are not matching from GHSA sources
**Version of Anchore Engine and Anchore CLI if applicable**: current tip of master, and I believe this has been a problem for the GHSA source since it was introduced

**What happened**:

```python
            elif self.pkg_type in ['npm', 'js']:
                likematch = '%js%'
                do_langscan = True
```

Will not match against the namespace in GHSA, which is `"github:npm"`. Other languages might be affected as well, so all need to be double-checked (e.g. Ruby)

Tangentially, the `Fixed In` reporting is not entirely accurate as many affected versions can have many "fixed in" values:

```
Vulnerability ID           Package         Severity        Fix          CVE Refs             Vulnerability URL                                        Type        Feed Group        Package Path                         
GHSA-pp7h-53gx-mx7r        bl-2.2.0        High            1.2.3        CVE-2020-8244        https://github.com/advisories/GHSA-pp7h-53gx-mx7r        npm         github:npm        /node_modules/bl/package.json   
```

The above does not represent what is displayed here https://github.com/advisories/GHSA-pp7h-53gx-mx7r

### LINKED ISSUE #639: Vulnerability fix version isn't correct for vulnerabilities that have multiple fixes available
Some records, like GitHub Security Advisories will include multiple fix versions for different ranges of a library. The system currently (as of v0.8.1) picks the first match on the name, rather than the with the range that matches the actual found artifact in the image.

zhill:
Adds fixes for npm and gem matching with GHSA records and gets correct fix version when multiple fixes available.

Fixes #633 
Fixes #639

### DIFF
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -1883,30 +1883,51 @@ class ImagePackageVulnerability(Base):
         {}
     )
 
+    def fix_candidates(self) -> list:
+        """
+        Return the list of FixedArtifact record given a package has been matched to the vulnerability by package/src-package
+        name. Does not perform a version check. Will return empty list if no matches
+
+        :return: the name-matched FixedArtifact record list
+
+        """
+
+        if self.vulnerability.fixed_in:
+            name_matches = [self.pkg_name, self.package.normalized_src_pkg]
+            return [x for x in self.vulnerability.fixed_in if x.name in name_matches]
+        return []
+
     def fixed_artifact(self):
         """
         Return the FixedArtifact record given a package has been matched to the vulnerability
+
         :return: the matched FixedArtifact record or None if not found
         """
-        if self.vulnerability.fixed_in:
-            name_matches = [self.pkg_name, self.package.normalized_src_pkg]
-            fixed_artifacts = [x for x in self.vulnerability.fixed_in if x.name in name_matches]
-            if fixed_artifacts and len(fixed_artifacts) == 1:
-                fixed_in = fixed_artifacts[0]
-            elif fixed_artifacts and len(fixed_artifacts) > 1:
-                # TODO version comparision logic to select the correct fixed in record.
-                # Temporary workaround: return the first match after sorting by name
-                fixed_in = sorted(fixed_artifacts, key=lambda fixed_artifact: fixed_artifact.name)[0]
-            else:
-                fixed_in = None
-        else:
-            fixed_in = None
+
+        candidates = self.fix_candidates()
+        candidate_count = len(candidates) if candidates else 0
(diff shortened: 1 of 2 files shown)


dakaneye:
LGTM!

tests/unit/anchore_engine/services/policy_engine/engine/test_vulnerability_matches.py
```diff
@@ -63,6 +78,61 @@ def vulnerability_with_fix(empty_vulnerability):
...
+    f.parent = fixed_vuln
+    f.include_later_versions = False
+    f.epochless_version = f.version
+    f.fix_metadata = {'first_patched_version': '1.1.2'}
+    f.created_at = datetime.datetime.now()
+    f.updated_at = datetime.datetime.now()
+    f.fix_observed_at = f.updated_at
+    fixed_vuln.fixed_in = [f]
+
+    f = FixedArtifact()
```
alfredodeza, line 119: this `f` instance doesn't look like it is used in `fixed_vuln`, that is probably intentional, but I am not following how that is working to get two fixed artifacts if the `empty_semver_vulnerability` is only used with `fixed_vuln`

zhill: The parent relationship connects it up in the fixed_in collection (`f.parent = fixed_vuln`) due to how SQLAlchemy classes work with the relationship definition. Line 117 is actually redundant and can be removed.

anchore_engine/db/entities/policy_engine.py
```diff
@@ -1883,30 +1883,51 @@ class ImagePackageVulnerability(Base):
...
+        candidates = self.fix_candidates()
+        candidate_count = len(candidates) if candidates else 0
+        if candidate_count == 0:
+            return None
+
+        fixed_in = None
+
+        # candidate_count >= 1
+        # Do version checks. This package must be affected by the range but not fixed.
+        matched = [x for x in candidates if x.match_but_not_fixed(self.package)]
```
nightfurys, line 1916: Yep, prior this change `match_but_not_fixed` is used for finding matches but was not used for looking up the associated fixed in version (or no fix). This is exactly the change I was thinking of as well. lgtm

tests/unit/anchore_engine/services/policy_engine/engine/test_vulnerability_matches.py
alfredodeza: Ahhh that is what I was hoping for, but wasn't sure that we were dealing with a model here.

PR #642, opened 2020-10-06
Adds db_engine_args handling in config.yaml for more db connection configuration

### LINKED ISSUE #641: Ability to set pool_recycle args for db connection handling
**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): Feature Request

If this is a FEATURE REQUEST, please:
Add ability to specify more db connection pool settings, including at least the ability to set the `pool_recycle` timeout to ensure stale connections are recycled before being used and causing an error if the connection was closed by the server or some intermediate network gear (e.g. a load balancer).

This is useful for cases where load balancers have short idle connection timeouts like 5-10 minutes.

Should add a generic `db_engine_args` section and handlers for config.yaml to allow per-deployment customization.
e.g.
```
credentials:
  database:
    db_connect: ...
    db_connect_args:
      timeout: 120
      ssl: false
    db_pool_size: 30
    db_pool_max_overflow: 100
    db_engine_args:
      <key-values to pass>
      pool_recycle: 600
```

zhill:
Fixes #641

### DIFF
--- a/anchore_engine/db/entities/common.py
+++ b/anchore_engine/db/entities/common.py
@@ -126,7 +126,9 @@ def do_connect(db_params):
 
     db_connect = db_params.get('db_connect', None)
     db_connect_args = db_params.get('db_connect_args', None)
-    db_engine_args = db_params.get('db_engine_args', {})
+    db_engine_args = db_params.get('db_engine_args')
+    if db_engine_args is None:
+        db_engine_args = {}
 
     # for bkwds compat
     if db_params.get('db_pool_size', None):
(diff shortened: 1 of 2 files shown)


anchore_engine/db/entities/common.py
```diff
@@ -126,7 +126,9 @@ def do_connect(db_params):
 
     db_connect = db_params.get('db_connect', None)
     db_connect_args = db_params.get('db_connect_args', None)
-    db_engine_args = db_params.get('db_engine_args', {})
+    db_engine_args = db_params.get('db_engine_args')
```
dakaneye, line 129: I think this change doesn't introduce any different functionality?

anchore_engine/db/entities/common.py
```diff
@@ -155,7 +155,6 @@ def do_connect(db_params):
 
     # set up the global session
     try:
-        #                SerializableSession = sessionmaker(bind=engine.execution_options(isolation_level='SERIALIZABLE'))
```
dakaneye, line 158: +1 for removing dead code

dakaneye (approved):
Good cleanup! less risk of KeyErrors and makes the composition fo the db params more clear.

anchore_engine/db/entities/common.py
zhill: This was to handle the case where there is a 'null' value in the yaml config directly. Ensures that we get an empty dict. Lines 130 & 131 unify the cases. May be able to be handled more elegantly.

PR #644, opened 2020-10-06
Type checks and broaden exception handling around json loading in err…

### LINKED ISSUE #643: Error response builder broken
Changes introduced by https://github.com/anchore/anchore-engine/pull/593  json loads an internal data structure without type checking. When the structure is a dictionary, json load fails as it only expects input of type str, bytes and bytearray. The end result to the user request no longer captures the actual error transmitted by one of the internal services

nightfurys:
…or response builder

Fixes #643

**Special notes**:

### DIFF
--- a/anchore_engine/common/helpers.py
+++ b/anchore_engine/common/helpers.py
@@ -28,23 +28,25 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
         ret['detail']['error_codes'] = []
 
     if isinstance(errmsg, Exception):
-        if 'anchore_error_json' in errmsg.__dict__:
-            # Try to load it as json
-            try:
-                err_json = json.loads(errmsg.__dict__.get('anchore_error_json', None))
-            except ValueError:
-                # Then it may just be a string, we cannot do anything with it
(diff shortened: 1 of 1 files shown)


anchore_engine/common/helpers.py
```diff
@@ -28,23 +28,25 @@ def make_response_error(errmsg, in_httpcode=None, details=None):
...
-            try:
-                err_json = json.loads(errmsg.__dict__.get('anchore_error_json', None))
-            except ValueError:
-                # Then it may just be a string, we cannot do anything with it
-                logger.debug('Failed to parse anchore_error_json as json')
-                return ret
-
-            if {'message', 'httpcode', 'detail'}.issubset(set(err_json)):
-                ret.update(err_json)
+        anchore_error = errmsg.__dict__.get('anchore_error_json', None)
```
zhill, line 31: Non-blocking: I think this should be `getattr(errmsg, 'anchore_error_json', None)` to be more clean than accessing the protected member.

zhill (approved):
Small suggestion, but non-blocking. I want to revisit this function completely, but not in scope for this fix.

PR #646, opened 2020-10-06
Allow for analyzer output to optionally be json encoded

wagoodman:
Today the analyzers have several stages of encoding and decoding --as part of the 3.0 work we will be removing some of these stages. This specific change allows for the json decoding step to be optional, enabling use of the existing analyzers mixed with the new analyzers (from syft). This will help in testing the swap out of the analyzers and remove extra encoding/decoding steps from the analysis path.

### DIFF
--- a/anchore_engine/common/helpers.py
+++ b/anchore_engine/common/helpers.py
@@ -105,7 +105,7 @@ def make_anchore_exception(err, input_message=None, input_httpcode=None, input_d
 
         if error_codes:
             ret.anchore_error_json['detail']['error_codes'].extend(error_codes)
-                                   
+
     return ret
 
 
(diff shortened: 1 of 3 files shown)


anchore_engine/common/helpers.py
```diff
@@ -366,3 +360,12 @@ def make_eval_record(userId, evalId, policyId, imageDigest, tag, final_action, e
...
     return payload
+
+def safe_extract_json_value(value):
+    # support the legacy serialized json string
+    if isinstance(value, str):
+        try:
+            value = json.loads(value)
+        except json.decoder.JSONDecodeError:
+            pass
+    return value
```
luhring, line 371: nit: I don't know if there's an established standard with regard to ending files with newlines, but noticed this was a change from the previous state

luhring (approved):

anchore_engine/common/helpers.py
```diff
@@ -366,3 +360,12 @@ def make_eval_record(userId, evalId, policyId, imageDigest, tag, final_action, e
     payload["last_updated"] = payload['created_at']
 
     return payload
+
+def safe_extract_json_value(value):
+    # support the legacy serialized json string
+    if isinstance(value, str):
```
alfredodeza, line 366: I don't think that is necessary to check for an instance of `str`. The `try/except` block can already handle this

alfredodeza (approved):
This is good to go but needs a rebase. The build failures where my fault, the fix was merged here https://github.com/anchore/anchore-engine/pull/650

Toure (approved):
Looks good, once the type check is removed.

PR #650, opened 2020-10-06
bump urllib3 to prevent reported vulnerability with current version

alfredodeza:
Relevant changelog section in urllib3: https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1259-2020-04-16

Same fix in https://github.com/anchore/anchore-cli/pull/130


dakaneye (approved):
Looks good to me!

PR #659, opened 2020-10-08
Functional Test (conftest) improvements

### LINKED ISSUE #658: Cannot run functional tests locally
PR https://github.com/anchore/anchore-engine/pull/653  is causing all tests to fail locally with:

```
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=8228): Max retries exceeded with url: /v1/accounts/functional_test (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff1d01facd0>: Failed to establish a new connection: [Errno 61] Connection refused'))
```

I believe this is caused by the automatic fixture `create_functional_test_account_with_teardown` which is not needed for every functional test.  By declaring the `autouse=True` in a top-level `conftest.py` it means that every test underneath it in any sub-directory will have to have that running. 

The functional tests as proposed in this PR rely on a running instance of Anchore Engine which I don't have, preventing me from running tests that do not need Anchore Engine running but are functional and belong to that path.

I see a couple of ways to improve this:

1. Move the newly added `conftest.py` into a separate directory for API tests that rely on this automatic functionality
1. Remove all these tests into a separate repository/project

For the time being, I think it is easier to achieve the first option, while trying to get these tests out of the Anchore Engine repo and into its respective project.

dakaneye:

**What this PR does / why we need it**: Makes it so that functional tests that don't require a running anchore don't need it to execute (non-api tests)

**Which issue this PR fixes** fixes #658 

**Special notes**:


tests/functional/services/api/conftest.py
```diff
@@ -31,7 +186,7 @@ def remove_image_by_id():
                 # Therefore, I've observed this finalizer running after the account is deleted. It's not the end of
                 # the world, shouldn't be a failed test. If I make this fixture autouse=True, it has been generating an
                 # extra matrix of tests which is worse than just letting the finalizer skip
-                _logger.info("{} account does not exist, ignoring for teardown".format(FT_ACCOUNT))
+                get_logger(__name__).info("{} account does not exist, ignoring for teardown".format(FT_ACCOUNT))
```
alfredodeza, line 189: not sure I follow this change... I believe the previous behavior would be the same.

dakaneye: Had to move the get_logger method out of the class, and this was the only method that needed it so just made a minor adjustment

alfredodeza: It looks like this file had a top-level `_logger` variable that this change removes:

```python
_logger = get_logger(__name__)
```

But you are saying that this was in a class? It is usually common practice to have that `_logger` (or most of the time `logger`) defined at the top so that it can be used anywhere. This line in particular doesn't look like it is using a custom name (e.g. `get_logger('remove_image_by_id')`), so it doesn't look like there is any benefit to  it - the behavior is the same it seems?

dakaneye: added it back. it was just throwing errors at the time where I combined the upper level conftest with this one.

alfredodeza (approved):

PR #666, opened 2020-10-09
Skip version checks if fix candidate list contains only one element

### LINKED ISSUE #665: Fixed in version broken for some distros
Computation of fixed-in version is broken for distros where vulnerabilities are found using an exact match i.e. vulnerable-in records instead of version comparison with a fixed-in record. Caused by changes introduced for handling multiple fixed-in records by https://github.com/anchore/anchore-engine/pull/640

nightfurys:
Fixes #665

**Special notes**:


zhill (approved):
lgtm

Issue #654, opened 2020-10-07
Package list API should use syft packages (python packages)


wagoodman:
As an engine user I expect the product to leverage `syft` capabilities to produce the listing for **Python packages**. 

**AC**
In order to support the existing UI client implement a translation layer on `syft` for the following:

1. Match the existing packages API call syntax. 

__Steps to Test__
1. Use the Anchore CLI for known image and compare results to a known image matches.
1. Look at the `anchore_engine/analyzers/modules` dir within the anchore container image and ensure the Python analyzer (`32_python_packages.py`) is not present.
1. `pytest -v standalone/test_package_list` runs in the `anchore-engine/tests/functional/clients` directory reporting Python-related tests that are all passing.

__Developer Notes__
New tests and fixtures should be added to ensure that before/after changes do not alter the expectations of the system.

wagoodman:
The syft cataloger 2.0 work is required to continue further. The existing cataloger uses python-specific utils, where as the golang equivalent will need to piece together the information from multiple file sources --requiring the cataloger 2.0 interface to do so.

PR #669, opened 2020-10-13
analyzers: test python packages

alfredodeza:
**What this PR does / why we need it**: Adds a verification that Python packages are correctly cataloged/analyzed currently so that when `syft` integration is done we can ensure the same behavior is preserved.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: _partially_ addresses https://github.com/anchore/anchore-engine/issues/654

**Special notes**:


tests/functional/clients/standalone/test_package_list.py
```diff
@@ -1,5 +1,5 @@
 # from ['image']['imagedata']['analysis_report']['package_list']
-from .fixtures import gems
+from .fixtures import gems, pypkgs
```
Toure, line 2: Should we start to create separate test_package modules per language as this file is going to become very large and if there are additional fields to test it will just add to the size?

tests/functional/clients/standalone/fixtures/gems.py
```diff
@@ -2,6 +2,14 @@
 # Generated from result['image']['imagedata']['analysis_report']['package_list']['pkgs.gems']
 
 pkgs = [
+     ('/usr/lib/ruby/gems/2.7.0/specifications/bundler-2.1.4.gemspec',
```
Toure, line 5: Looks good.

wagoodman (commented):
We haven't gotten a chance to open up a PR for the syft-engine work yet, however, there are some related changes that could be done here as well that improve readability:
- https://github.com/anchore/anchore-engine/compare/syft-gems#diff-20b591ddd4a5f90efeb24bf5050a331791d22c71f0e706b9f4ad794de018b4a9
- https://github.com/anchore/anchore-engine/compare/syft-gems#diff-3bdf355e34c794a312527c1fb407e9c38db8ad22053d4952c2d3adc43691a98a
(consider this a nit)

tests/functional/clients/standalone/test_package_list.py
alfredodeza: @Toure these are already separate files:

```
$ tree standalone/fixtures/
standalone/fixtures/
├── __init__.py
├── gems.py
├── npms.py
└── pypkgs.py

0 directories, 4 files
```

So the sizes are already small. Do you mean I need to do something else with these?

wagoodman (approved):

tests/functional/clients/standalone/test_package_list.py
Toure: sorry, the comment should have been against the test_package_list.py file itself to be broken out into language-specific modules.

alfredodeza:
> We haven't gotten a chance to open up a PR for the syft-engine work yet, however, there are some related changes that could be done here as well that improve readability:
> 
>     * https://github.com/anchore/anchore-engine/compare/syft-gems#diff-20b591ddd4a5f90efeb24bf5050a331791d22c71f0e706b9f4ad794de018b4a9
> 
>     * https://github.com/anchore/anchore-engine/compare/syft-gems#diff-3bdf355e34c794a312527c1fb407e9c38db8ad22053d4952c2d3adc43691a98a
>       (consider this a nit)

@wagoodman I agree on the formatting. Since we are dealing with the unserialized JSON, the formatters break (black can't handle it). I can revise once the syft integration makes it in

tests/functional/clients/standalone/test_package_list.py
alfredodeza: Oh ok, I see what you mean. Would it be OK to get those non-functional changes (along with @wagoodman 's suggestions) in a separate PR?

PR #676, opened 2020-10-14
Added successes and errors Prometheus metrics

juanjolmm:
Signed-off-by: Juanjo <juanjo.lmm@gmail.com>

What this PR does / why we need it:
At my current job, we are using Anchore and we send the Prometheus metrics to Grafana.
We are missing the metrics for successes and errors. We want to know how many images have been successfully analyzed and how many errors there have been too.


zhill (approved):
Makes sense and looks good. Thanks!

PR #679, opened 2020-10-15
Syft gems

### LINKED ISSUE #629: Package list API should use syft packages (for Ruby gems)
As an engine user I expect the product to leverage the anchore OSS toolbox so I understand how extend the software for **Ruby gems**. 

__AC__
In order to support the existing UI client implement a translation layer on syft for the following.  

1. Match the existing packages API call syntax. 

__Steps to Test__
1. Use the Anchore CLI for known image and compare results to a known image matches.
1. Look at the `anchore_engine/analyzers/modules` dir within the anchore container image and ensure the gem analyzer is not present.

__Developer Notes__
Should include at least functional level testing to ensure the replaced analyzer output has not changed.

luhring:

- introduces wrapper for [syft](https://github.com/anchore/syft)
- replaces gem analyzer module with syft functionality
- makes minor improvements to the flow of analyzer data
- modifies the approach to functional tests

Fixes #629 

**Special notes**:


Dockerfile
```diff
@@ -177,7 +178,7 @@ RUN set -ex && \
 HEALTHCHECK --start-period=20s \
     CMD curl -f http://localhost:8228/health || exit 1
 
-USER 1000
-
+USER anchore:anchore
```
zhill, line 181: This should not be changed from the value in master, the value is required for dynamic UID mapping in OpenShift

luhring: @zhill Good catch, I've reverted this change. I'm assuming this relates to this functionality -> https://www.openshift.com/blog/a-guide-to-openshift-and-uids . For future reference, where can I find out more about how Anchore interacts with OCP's UID management?

anchore_engine/clients/syft_wrapper.py
```diff
@@ -0,0 +1,37 @@
+import os
+import json
+import shlex
+
+from anchore_engine.utils import run_command, run_command_list, manifest_to_digest, AnchoreException
+from anchore_engine.subsys import logger
+
+def run_syft(image):
+    proc_env = os.environ.copy()
+    cmd = "syft -vv -o json {image}".format(image=image)
```
dakaneye, line 10: It's possible that some of the command line options for syft may change / adjust in the future, would it be worth making them configurable?

scripts/ci/prep-local-docker-registry-credentials
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
 set -euvo pipefail
+
+# Creates Basic Authentication Context
+pushd tests/functional/artifacts/registry/
```
dakaneye, line 6: if we're not gonna `popd` why not just use `cd`?

scripts/ci/prep-local-docker-registry-credentials
```diff
@@ -1,11 +1,12 @@
...
+
 echo "Generating Local Docker Registry Artifacts at $(pwd)"
 mkdir -p auth certs
 
-# Creates Basic Authentication Context
 docker run --rm --entrypoint htpasswd registry:2.7.0 -Bbn "$ANCHORE_TEST_DOCKER_REGISTRY_USER" "$ANCHORE_TEST_DOCKER_REGISTRY_PASS" > ./auth/htpasswd
-openssl req -newkey rsa:4096 -nodes -sha256 -keyout ./certs/domain.key -x509 -days 365 -out ./certs/domain.crt -subj "/C=US/ST=California/L=Santa Barbara/O=Anchore/CN=functional_test"
-docker build -f tests/functional/artifacts/registry/Dockerfile -t anchore-local-docker-registry:latest .
-rm -rf auth certs
```
dakaneye, line 11: don't want to clean up the directories anymore?

tests/functional/test_systemstatus.py
```diff
@@ -8,9 +8,9 @@
 import os
 
 # Functional tests for system status as a basic connectivity/anchore-engine up test
-anchore_user = os.environ.get('ANCHORE_CLI_USER', 'admin')
```
dakaneye, line 11: why, remove the defaults?

tests/functional/services/api/registries/__init__.py
```diff
@@ -3,8 +3,8 @@
 
 def get_registry_info():
     return {
-        'user': os.environ.get('ANCHORE_TEST_DOCKER_REGISTRY_USER', 'anchore'),
```
dakaneye, line 6: why remove the defaults?

anchore_engine/analyzers/syft.py
```diff
@@ -0,0 +1,50 @@
...
+import sys
+import json
+import collections
+
+import anchore_engine.analyzers.utils
+from anchore_engine.clients.syft_wrapper import run_syft
+
+
+def handle_gem(findings, artifact):
+    pkg_key = artifact['locations'][0]['path']
```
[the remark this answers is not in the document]
wagoodman: Agreed that getting with a default is a good practice, however, in this case there is no "safe" default, and we want the act of accessing to fail with exception when the value is missing.

anchore_engine/clients/syft_wrapper.py
wagoodman: I think adding more features as we go (throughout the 3.0 development process) makes sense, not necessarily on this PR quite yet.

scripts/ci/prep-local-docker-registry-credentials
wagoodman: purely habit :) does no harm though

dakaneye: sure, i guess it's a nit picky comment haha

scripts/ci/prep-local-docker-registry-credentials
wagoodman: correct --this is one of the cleanups. Since we've moved to a VM machine type in CircleCI we can leverage (and as of this PR are leveraging) docker volumes instead of building an image with ephemeral secrets.

These directories are now persisted at `tests/functional/artifacts/registry/` and are mounted into the registry container as a volume, so they are never cleaned up or removed (which we could optionally do after `compose-down`).

tests/functional/test_systemstatus.py
wagoodman: When moving the test harness to a containerized setup (the `job-runner` container in the compose stack), default values were hiding in a few different places, and were non-obvious to find and override. This was an effort to move all defaults to be derived from a single source of truth, failing fast if the value is missing.

For example, the `os.environ.get('ANCHORE_CLI_URL', 'http://localhost:8228/v1')` line works fine when running on localhost or if the container is forwarding ports to the host, but from container-to-container the value needs to change. There isn't a "hint" to a user of the functional test of what values need to be set from a compose file or otherwise --removing the default makes this obvious by forcing a single source of truth, the environment variable.

This may make local dev harder, in which case, we should think about a good compromise --maybe put back some defaults, but place them in a single spot in the functional codebase as a chokepoint.

tests/functional/services/api/registries/__init__.py
wagoodman: [ditto](https://github.com/anchore/anchore-engine/pull/679#discussion_r506653510)

Dockerfile
```diff
@@ -167,6 +167,7 @@ RUN set -ex && \
 RUN set -ex && \
     pip3 install --no-index --find-links=./ /build_output/wheels/*.whl && \
     cp /build_output/deps/skopeo /usr/bin/skopeo && \
+    curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /anchore_engine/bin v0.3.0 && \
```
Btodhunter, line 170: This needs to happen in the first stage. The binary should be installed to the `/build_output` directory, then copied to a directory in the PATH in the second stage. We use the tarball created from the /build_output directory to build our DCAR(DOD) images.

.circleci/config.yml
```diff
@@ -250,7 +240,7 @@ workflows:
             - push_dev_image
           image_name: anchore/anchore-engine-dev:${CIRCLE_SHA1}
           private_registry: true
-          timeout: '2000'
+          timeout: "2000"
```
alfredodeza, line 243: is this a linter thing? I was unaware of these rules

anchore_engine/analyzers/syft.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # craft the artifact document
+    pkg_value = {
+            'name': artifact['name'],
+            'versions': [artifact['version']],
+            'latest': artifact['version'],
+            'sourcepkg': artifact['metadata'].get('homepage', ''),
+            'files': artifact['metadata'].get('files', []) or [],
```
alfredodeza, line 21: At first sight it looks redundant to say `or []` because the `.get()` is falling back to `[]` when `'files'` is not present. But, is it possible this is accounting for a situation where `'files'` is a `None` or some other value and this wants to enforce an empty list if so?

anchore_engine/analyzers/syft.py
```diff
@@ -0,0 +1,50 @@
...
+            'origins': artifact['metadata'].get('authors', []) or [],
+            'lics': artifact['metadata'].get('licenses', []) or [],
+        }
+
+    # inject the artifact document into the "raw" analyzer document
+    findings['package_list']['pkgs.gems']['base'][pkg_key] = pkg_value
+
+def filter_artifacts(artifact):
+    # only allow artifacts which have handlers implemented, ignore the rest
+    return artifact['type'] in artifact_handler_dispatch.keys()
```
alfredodeza, line 31: there is no need to call `.keys()`, it defaults to looking into keys

anchore_engine/analyzers/utils.py
```diff
@@ -2047,3 +2048,22 @@ def write_kvfile_fromdict(file, indict):
...
+def merge_nested_dict(a, b, path=None):
+    if path is None: path = []
+    for key in b:
+        if key in a:
+            if isinstance(a[key], dict) and isinstance(b[key], dict):
+                merge_nested_dict(a[key], b[key], path + [str(key)])
+            elif a[key] == b[key]:
+                pass # same leaf value
+            else:
+                raise Exception('Conflict at %s' % '.'.join(path + [str(key)]))
```
alfredodeza, line 2066: nit: Doesn't look like the `else` here is needed if the loop can `continue`

```suggestion
                continue # same leaf value

            raise Exception('Conflict at %s' % '.'.join(path + [str(key)]))
```

anchore_engine/analyzers/utils.py
```diff
@@ -2047,3 +2048,22 @@ def write_kvfile_fromdict(file, indict):
...
             #thestr = thestr.encode('utf8')
             OFH.write(thestr)
+
+def defaultdict_to_dict(d):
+    if isinstance(d, collections.defaultdict):
+        d = {k: defaultdict_to_dict(v) for k, v in d.items()}
+    return d
+
+def merge_nested_dict(a, b, path=None):
+    if path is None: path = []
```
alfredodeza, line 2058: this doesn't read right because it is inlining code with the `:`. Surprised the linter didn't complain about this! 
```suggestion
    if path is None: 
        path = []
```

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -143,16 +143,29 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
...
             if success:
+                if use_cache_dir and dest_type == 'oci':
+                    # syft expects blobs to be nested inside of the oci image directory. If the --dest-shared-blob-dir skopeo option is used we need to
+                    # provide access to the blobs via a symlink, as if the blobs were stored within the oci image directory
+                    blobs_dir = os.path.join(copydir, "blobs")
+                    if os.path.exists(blobs_dir) and os.path.isdir(blobs_dir):
+                        # if this directory is not empty, there is an issue and we should expect an exception
+                        os.rmdir(blobs_dir)
+
+                    cmd = "ln -s {} ./blobs".format(use_cache_dir)
```
alfredodeza, line 164: why is this using a system command instead of linking with Python?

tests/functional/test_systemstatus.py
alfredodeza: After reading this comment thread and discussing it a couple of days ago I am going to push back on this change. This line is not _changing values_, it is assigning them if they are not set. 

I agree that it is non-obvious that there are tests like this one relying on environment variables, which aren't advertised  clearly. This could be changed to be put into a central location where environment variables are set. In my case, I tend to expect them in a `conftest.py`.

I also agree that there is no "hint" of what values are being used. This is another thing that can be improved with Pytest itself, where the report header can be enhanced to add the environment variables that are crucial for the test.

The single source of truth for the values is an excellent approach, and once again I agree that it is worth pursuing. However, a single source of truth should _not force_ a user to have to source a file before running tests. The addition of a file and the requirement of sourcing it is a step back for encouraging local development as it creates a hassle.

I really like the idea of the compromise suggested by @wagoodman here. I can follow up with a PR that addresses these concerns:

1. Put the defaults back as a fallback only
1. All environment variables should go into a central location - `tests/functcional/conftest.py` is a good spot for these
1. Add a Pytest hook to enhance the report header to include the variables (and their source)

This comment should not prevent this PR from moving forward

anchore_engine/analyzers/syft.py
wagoodman: That's right, with syft v0.2.0 there were fields that were explicitly `null`, so the `get()` would positively get `None` value back. Since this was fixed in v0.3.0 at the end of last week I can remove this logic.

anchore_engine/clients/skopeo_wrapper.py
wagoodman: I'll switch over to `os.symlink` 👍

anchore_engine/analyzers/utils.py
wagoodman: not quite, the first if case doesn't return so would need a `continue` or leave the `else` statement as is (leaning towards the latter, so I'll leave this almost as is)

alfredodeza (approved):

hev:
After some initial problems this looked good to me. I did the following. 

1. Checked out the `syft-integration` branch
1. Added image `ruby:aline`
1. Confirmed that `/usr/local/lib/python3.6/site-packages/anchore_engine/analyzers/modules/12_gem_package_list.py` was no longer present by running `find / | grep analyzers/modules`

For troubleshooting, it would be nice if exceptions were handled in am image fail state but that's a separate concern.

hev:
This is a nit, but appending my previous comment to add some minor feedback. It would be nice if there was a positive test that syft was being used as an analyzer. That is rather than grep to confirm a particular analyzer file didn't appear, one that could prove that the analyzer was in fact using syft.

Issue #680, opened 2020-10-19
Investigate using OCI format exclusively during analysis


luhring:
Currently, there are ways for the analysis flow to result in skopeo to save an image using the `dir:` format instead of the `oci:` format. `dir:` is non-standard and only intended for debugging; moreover, the Toolbox tools support the OCI format but do not support skopeo's proprietary `dir:` format.

The `dir:` format is used if either of these are true:

- The image uses foreign diffs: https://github.com/anchore/anchore-engine/blob/master/anchore_engine/clients/skopeo_wrapper.py#L102
- The image is not a Linux image: https://github.com/anchore/anchore-engine/blob/master/anchore_engine/clients/skopeo_wrapper.py#L113

**Objective:** Determine if it's possible to always use the OCI format for saving images, thus resolving the issue of downstream tools (like syft and grype) being unable to inspect the given image.

### Suggested task breakdown

1. Look into https://github.com/containers/skopeo/issues/545 and see if it's feasible to update skopeo itself to save Windows images using the OCI format and avoid using foreign diffs.
1. Look at what would need to change in the anchore-engine code if the `dir:` format were no longer present anywhere.
1. Determine stereoscope's behavior with Windows images.

luhring:
Regarding task 3 —

Stereoscope appears to be capable of handling Windows OCI images, as long as all blobs are present within the OCI directory structure, and no unsupported media types are present (e.g. `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`).

I tested with the image `mcr.microsoft.com/windows/nanoserver:1903`. The OCI directory's `index.json` listed its manifest at `"sha256:2a8ca7db7332bc6c3b825a4540c33b47ad588947aa130f07b7493a863daa3ba3"`.

The manifest looked like this:

```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:b4382afec288f18a7255c91b9c5dfe37d66dc1aca7be8cefdbbd5b14ae794d3f",
    "size": 354
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.nondistributable.v1.tar+gzip",
      "digest": "sha256:0e31ee8f400716f940be9db76d6eedb1a903860385b4ca8f70513012d53fe1f1",
      "size": 103066803,
      "urls": [
[...]

PR #684, opened 2020-10-22
Always copy images to OCI format and fetch any foreign diffs

luhring:
This PR does the following:

- removes the option to specify non-OCI destination formats for the `skopeo copy` operation as images are ingested
- in order to accommodate images with "foreign diffs" (a.k.a. "[non-distributable layers](https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers)" in OCI nomenclature), after the skopeo copy, we fetch any such layers using the URLs provided, and we update the OCI manifest blob content to remove "`nondistributable.`" from the layer media type

The objectives are:

1. To allow for Syft (and any other Toolbox tools) to count on an OCI-format image being available as input
1. To simplify the flow of execution for pulling an image within the application

This implementation aims to satisfy task 2 from #680.

🔥 Please be blunt with me regarding this Python code! 😄

### DIFF
--- a/anchore_engine/clients/skopeo_wrapper.py
+++ b/anchore_engine/clients/skopeo_wrapper.py
@@ -7,6 +7,7 @@ import anchore_engine.configuration.localconfig
 from anchore_engine.utils import run_command, run_command_list, run_command, manifest_to_digest, AnchoreException
 from anchore_engine.subsys import logger
 from anchore_engine.common.errors import AnchoreError
+from urllib.request import urlretrieve
 
 def manifest_to_digest_shellout(rawmanifest):
     ret = None
@@ -59,7 +60,7 @@ def copy_image_from_docker_archive(source_archive, dest_dir):
         logger.error("command failed with exception - " + str(err))
         raise err
 
-def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=None, parent_manifest=None, use_cache_dir=None, dest_type='oci'):
+def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=None, parent_manifest=None, use_cache_dir=None):
     try:
         proc_env = os.environ.copy()
         if user and pw:
@@ -94,13 +95,18 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
             global_timeout_str = ""
 
         os_overrides = [""]
+        blobs_to_fetch = []
+
         if manifest:
             manifest_data = json.loads(manifest)
 
-            # skopeo doesn't support references in manifests for copy/download operations, with oci dest type - if found, override with dir dest_type
             for l in manifest_data.get('layers', []):
                 if 'foreign.diff' in l.get('mediaType', ""):
-                    dest_type = 'dir'
+                    layer_digest_raw = l.get('digest', "")
+                    layer_digest = get_digest_value(layer_digest_raw)
+                    layer_urls = l.get('urls', [])
+
+                    blobs_to_fetch.append({'digest': layer_digest, 'urls': layer_urls})
 
             if parent_manifest:
                 parent_manifest_data = json.loads(parent_manifest)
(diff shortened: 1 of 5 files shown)


anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -171,6 +194,47 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
 
     return True
 
+def get_digest_value(digest_with_algorithm_prefix: str):
+    digest_raw = digest_with_algorithm_prefix
+
+    if ":" in digest_raw:
```
alfredodeza, line 200: there is no need for the `if`, this can be abbreviated with:

```python
digest_raw.split(':')[-1]
```

Breaking it down, it means to split at `:` and then grab the last element of the list. It would work for situations where there is a `:` and when it is missing.

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -171,6 +194,47 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
...
+        _, digest_value = digest_raw.split(":", 1)
+        return digest_value
+
+    return digest_raw
+
+def ensure_no_nondistributable_media_types(oci_index_file_path: str):
+    manifest_file_path = get_manifest_path_from_index(oci_index_file_path)
+
+    with open(manifest_file_path, 'r') as manifest_file:
+        data = manifest_file.read()
```
alfredodeza, line 210: Since `data` is later loaded into JSON I believe you want to load it here with `json.load()` https://docs.python.org/3/library/json.html#json.load

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -171,6 +194,47 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
...
+        manifest_file.write(json.dumps(manifest))
+
+def remove_nondistributable(layer: dict):
+    updated_media_type = layer.get('mediaType', '').replace('nondistributable.', '')
+    layer['mediaType'] = updated_media_type
+    return layer
+
+def get_manifest_path_from_index(oci_index_file_path: str):
+    with open(oci_index_file_path, 'r') as index_file:
+        data = index_file.read()
```
alfredodeza, line 227: same here with `json.load()` https://docs.python.org/3/library/json.html#json.load

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -153,15 +154,37 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
...
                     # syft expects blobs to be nested inside of the oci image directory. If the --dest-shared-blob-dir skopeo option is used we need to
                     # provide access to the blobs via a symlink, as if the blobs were stored within the oci image directory
-                    blobs_dir = os.path.join(copydir, "blobs")
                     if os.path.exists(blobs_dir) and os.path.isdir(blobs_dir):
                         # if this directory is not empty, there is an issue and we should expect an exception
                         os.rmdir(blobs_dir)
 
                     os.symlink(use_cache_dir, blobs_dir)
+
+                for blob in blobs_to_fetch:
```
alfredodeza, line 168: I would extract this into a function so that you can add unit tests with ease

alfredodeza (changes requested):

anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -153,15 +154,37 @@ def download_image(fulltag, copydir, user=None, pw=None, verify=True, manifest=N
...
+                        blob_destination_path = os.path.join(blobs_dir, "sha256", blob['digest'])
+
+                        try:
+                            urlretrieve(url, blob_destination_path)
+                            blob_was_fetched = True
+                            break
+                        except:
+                            continue
+
+                    if not blob_was_fetched:
```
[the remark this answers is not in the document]
luhring: Good suggestion. I'm going to confirm with Zach later today that there's no action to be taken in this case (I had left this TODO as a reminder for that conversation) — and I'll remove if that assumption holds true.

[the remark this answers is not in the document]
luhring: (confirmed! 👍 )

tests/unit/anchore_engine/clients/test_skopeo_wrapper.py
```diff
@@ -0,0 +1,170 @@
...
+                "digest": digest,
+                "urls": ["https://example.fooo.bar"]
+            }
+
+        return list(map(digest_to_blob, digests))
+
+    return return_value
+
+
+@pytest.fixture
```
alfredodeza, line 45: very nice

alfredodeza (approved):

PR #693, opened 2020-10-30
Replace Java analyzer module with Syft call

### LINKED ISSUE #681: Package list API should use syft (java)
As an engine user I expect the product to leverage `syft` capabilities to produce the listing for **Java packages**. 

**AC**
In order to support the existing UI client implement a translation layer on `syft` for the following:

1. Match the existing packages API call syntax. 

__Steps to Test__
1. Use the Anchore CLI for known image and compare results to a known image matches.
1. Look at the `anchore_engine/analyzers/modules` dir within the anchore container image and ensure the Java analyzer (`32_java_packages.py`) is not present.
1. `pytest -v standalone/test_package_list` runs in the `anchore-engine/tests/functional/clients` directory reporting Java-related tests that are all passing.

__Developer Tasks__
- [x] Add functional-level tests and fixtures (merge this into the `syft-integration` branch first before altering the java analyzer module). (PR https://github.com/anchore/anchore-engine/pull/690)
- [x] Add java use case to the syft inline-compare testing (in the syft repo). (PR https://github.com/anchore/anchore-engine/pull/693)
- [x] Remove the existing java analyzer module and add a new java package handler for syft under `anchore_engine/analyzers/syft.py`. (PR https://github.com/anchore/anchore-engine/pull/693)
- [x] Update syft java cataloger with additional metadata as needed. (PR https://github.com/anchore/syft/pull/247)

wagoodman:
- Adds Java handler to translate syft results for engine
- Removes the original Java analyzer module
- Updates the Java test fixtures to include extra syft results and remove the now unused metadata field
- Removes copies of the MANIFEST.MF and pom.properties file contents from the analysis results (the metadata field)
- Also removed commented out references to metadata code in the loaders to make me feel better

**Note: until https://github.com/anchore/syft/pull/247 is merged, syft is released with the update, and the pinned syft version updated, tests will fail.**

Closes #681


luhring (approved):
Looks good to me. Might want to have more seasoned Python eyes on this too if available 🐍

anchore_engine/analyzers/syft.py
```diff
@@ -48,6 +119,17 @@ def handle_python(findings, artifact):
 
 
 def handle_gem(findings, artifact):
+    """Gem results handler for syft output.
+
+    Args:
```
alfredodeza, line 124: It is great to have more thorough docstrings, but there is no convention in Anchore Engine for them and this is adding another variation of the existing ones. 

Since there is no convention yet, can this PR avoid making it specific?

anchore_engine/analyzers/utils.py
```diff
@@ -2069,3 +2069,26 @@ def merge_nested_dict(a, b, path=None):
         else:
             a[key] = b[key]
     return a
+
+def dig(target, *keys, **kwargs):
```
alfredodeza, line 2073: this needs unit tests

anchore_engine/analyzers/syft.py
```diff
@@ -3,11 +3,82 @@
...
+    
+    pkg_key = dig(artifact, 'metadata', 'virtualPath', default="N/A")
+
+    virtualElements = pkg_key.split(":")
+    if "." in virtualElements[-1]:
+        # there may be an extension in the virtual path, use it
+        java_ext = virtualElements[-1].split(".")[-1]
+    else:
+        # the last field is probably a package name, use the second to last virtual path element and extract the extension
+        java_ext = virtualElements[-2].split(".")[-1]
```
alfredodeza, line 30: most/all of this block would benefit from getting extracted into a separate helper. By having it here in the `handle_java` function, it makes it hard to unit test (there aren't any that I could see for this) and makes the function too large, reducing the readability

alfredodeza (changes requested):

anchore_engine/analyzers/utils.py
wagoodman: I'll add some unit tests for this

anchore_engine/analyzers/syft.py
wagoodman: I'll remove them for now.

anchore_engine/analyzers/syft.py
wagoodman: we haven't been writing unit tests for the analyzer module work, but instead are using the functional test fixtures to cover the necessary cases (consistent with the outlined strategy for the analyzer module work overall and consistent with the other analyzer module PRs). 

Without comments this function is ~30 lines of code, which I feel is not too large.

alfredodeza: That sounds reasonable to me. I was interested in the small if/else blocks, but I'm fine if you feel extracting them would be overkill.

luhring (approved):
🚀

Issue #682, opened 2020-10-21
package list API should use syft packages rpm, dpkg, apk, busybox binaries.


luhring:
Create new handlers within the syft anchore-engine module to replace the existing 10_package_list module.

__Test needed__

1. [x] add functional analyzer tests for debian packages (PR https://github.com/anchore/anchore-engine/pull/694 )
1. [x] add functional analyzer tests for rpm packages 
1. [x] add functional analyzer tests for apline packages (PR https://github.com/anchore/anchore-engine/pull/695)
1. [x] add functional analyzer tests for busybox binaries

**Developer Notes**

- [x] Replace debian analyzer module (PR https://github.com/anchore/anchore-engine/pull/710)
- [x] Replace centos analyzer module (PR https://github.com/anchore/anchore-engine/issues/701)
- [x] Replace alpine analyzer module (https://github.com/anchore/anchore-engine/pull/697)
- [x] Replace busybox analyzer module

PR #695, opened 2020-11-02
Add functional tests for alpine

wagoodman:
Adds functional tests for Alpine packages, setting up for future work in #682 .


tests/functional/clients/standalone/package_list/fixtures/alpine.py
```diff
@@ -0,0 +1,609 @@
+pkgfiles_all = {
+    "/bin": "APKFILE",
```
alfredodeza, line 2: we usually have a subset of packages, not all of them. Is there a particular reason this fixture has more than 10?

wagoodman: I can reduce this down to a subset 👍

tests/functional/clients/standalone/package_list/test_alpine.py
```diff
@@ -0,0 +1,46 @@
...
+        pkgs = result['image']['imagedata']['analysis_report']['package_list']['pkgs.all']['base']
+        assert pkgs.get(pkg) == version
+
+    @pytest.mark.parametrize('pkg,metadata', alpine.pkgs_allinfo.items())
+    def test_pkgs_allinfo(self, analyzed_data, pkg, metadata):
+        result = analyzed_data("py38")
+        pkgs = result['image']['imagedata']['analysis_report']['package_list']['pkgs.allinfo']['base']
+        loaded = dict(pkgs.get(pkg, {}))
+        
+        # a separate test exists just for the files attribute, leave this out of the assertion
```
wagoodman, line 28: note: I separated this (the files compare) out as it becomes a problem when syft gets integrated later

tests/functional/clients/standalone/package_list/fixtures/alpine.py
```diff
@@ -0,0 +1,379 @@
...
+            "/usr/share/apk/keys/s390x",
+            "/usr/share/apk/keys/x86/alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub",
+            "/usr/share/apk/keys/x86/alpine-devel@lists.alpinelinux.org-5243ef4b.rsa.pub",
+            "/usr/share/apk/keys/x86",
+            "/usr/share/apk/keys/x86_64/alpine-devel@lists.alpinelinux.org-4a6a0840.rsa.pub",
+            "/usr/share/apk/keys/x86_64/alpine-devel@lists.alpinelinux.org-5261cecb.rsa.pub",
+            "/usr/share/apk/keys/x86_64",
+        ],
+    },
+    "busybox": {
```
alfredodeza, line 203: busybox? That is surprising... I guess this is an actual package via `apkg`?

tests/functional/clients/standalone/package_list/test_alpine.py
```diff
@@ -0,0 +1,46 @@
...
+        metadata.pop('files')
+
+        assert loaded == metadata
+
+    @pytest.mark.parametrize('pkg,metadata', alpine.pkgs_allinfo.items())
+    def test_pkgs_allinfo_files(self, analyzed_data, pkg, metadata):
+        result = analyzed_data("py38")
+        pkgs = result['image']['imagedata']['analysis_report']['package_list']['pkgs.allinfo']['base']
+        loaded = pkgs.get(pkg, {})
+        assert set(loaded['files']) == set(metadata['files'])
```
alfredodeza, line 40: this implies that `loaded['files']` has duplicates that are different from `metadata['files']`, is my assumption correct? Or what is the benefit of using `set()` here?

alfredodeza (approved):
No blocking comments, this is fine to go once tests are passing

tests/functional/clients/standalone/package_list/fixtures/alpine.py
wagoodman: indeed it is

tests/functional/clients/standalone/package_list/test_alpine.py
wagoodman: this is setup for when running the tests against syft later --syft has a different ordering of the files list than what engine outputs. By testing files explicitly with set instead of a nested comparison (which will check the order of items) we can assert the data is correct without going to too much effort.

I can change this to a sorting of both lists instead of set to ensure the test fails if there are duplicates.

PR #697, opened 2020-11-02
Replace Alpine analyzer module with Syft call

wagoodman:
- Refactors the existing syft handlers into the `syft.handlers` python module
- Adds the Alpine handler that utilizes syft
- Removes the existing Alpine-specific code from the existing `package_list` analyzer module

Partially addresses #682


anchore_engine/analyzers/syft/__init__.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
alfredodeza, line 1: is this a script?

tests/functional/clients/standalone/package_list/test_alpine.py
```diff
@@ -23,7 +22,7 @@ def test_pkgs_all(self, analyzed_data, pkg, version):
     def test_pkgs_allinfo(self, analyzed_data, pkg, metadata):
         result = analyzed_data("py38")
         pkgs = result['image']['imagedata']['analysis_report']['package_list']['pkgs.allinfo']['base']
-        loaded = dict(json.loads(pkgs.get(pkg, '{}')))
+        loaded = dict(pkgs.get(pkg, '{}'))
```
alfredodeza, line 25: I *think* this now needs to fallback to an empty dict, not a JSON object? 

```suggestion
        loaded = dict(pkgs.get(pkg, {}))
```

wagoodman: yup yup, good catch

anchore_engine/analyzers/syft/__init__.py
wagoodman: snuck in there from the initial gems PR work (before it was an imported python module and instead being shelled out to). I'll take it out

alfredodeza (approved):

PR #709, opened 2020-11-10
Replaces older swagger-codegen models

zhill:
Uses new models built with marshmallow. These are easier to maintain and test.


dakaneye (approved):
Looks pretty straightforward to me

PR #714, opened 2020-11-11
analyzers: use syft to provide BusyBox package information

alfredodeza:
**What this PR does / why we need it**: Further integrates `syft` into Anchore Engine by replacing the BusyBox information in `pkgs.all`

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: Related to https://github.com/anchore/anchore-engine/issues/682

**Special notes**: This PR includes tests that were not verifying the output before. Noteworthy is that Anchore Engine parses BusyBox versions with a `v`, for example `v1.32.0`, and `syft` gives you a sanitized version of `1.32.0` which I believe is more correct.


wagoodman (approved):
🚀

alfredodeza:
Didn't realize this but this requires a syft release

anchore_engine/analyzers/syft/__init__.py
```diff
@@ -19,6 +19,16 @@ def catalog_image(image):
     nested_dict = lambda: collections.defaultdict(nested_dict)
     findings = nested_dict()
 
+    # This is the only use case for consuming the top-level results from syft,
```
Toure, line 22: Should this be placed in the handlers package just continuity?

alfredodeza: That is a great question. If I did that, I would still need to import it and call it here because it is needed just once. Is that what you would prefer? There is no other handling at all for busybox

Toure: Nope, that works for me.

Toure (approved):

anchore_engine/analyzers/syft/__init__.py
alfredodeza: Your point is still valuable though, and I think we should create a `busybox` module if this is no longer a "one of".

PR #716, opened 2020-11-13
Added unit tests for dockerfile gate

zburstein:

Adds dockerfile gate unit tests to expand testing coverage


tests/unit/anchore_engine/services/policy_engine/policy/gates/test_dockerfile.py
```diff
@@ -0,0 +1,297 @@
...
+USER admin
+COPY . /app
+RUN make /app
+CMD python /app/app.py
+EXPOSE 3000
+"""
+
+
+@pytest.fixture()
+def image():
```
dakaneye, line 19: global: naming. I've seen methods named like this before for builders and such, but they're more for adding attributes or getting attributes (i.e. auto value in Java). It'd be a bit more clear to subsequent readers if this was `get_image` or `construct_image` or `create_image`

zburstein: Thanks! This got me thinking and I was actually able to refactor down to a single test method

dakaneye (approved):
LGTM, good job!

tests/unit/anchore_engine/services/policy_engine/policy/gates/test_dockerfile.py
```diff
@@ -0,0 +1,278 @@
...
+test_contexts = [
+    *no_dockerfile_trigger_tests,
+    *effective_user_trigger_tests,
+    *exposed_port_trigger_tests,
+    *instruction_check_trigger_tests,
+    *actual_dockerfile_only_tests,
+]
+
+
+@pytest.mark.parametrize("test_context", test_contexts)
```
zhill, line 261: This is just a nit, so not blocking, but while I agree this is an efficient way to do the test, generally I'd prefer separate test functions for each trigger so that the test output is a bit easier to understand and it keeps the matrix for each test function a bit more tractable as the number of parameters and input types grows. Something to keep in mind for the future.

Overall though this looks good! Nice work!

zhill (approved):
lgtm

tests/unit/anchore_engine/services/policy_engine/policy/gates/test_dockerfile.py
zburstein: @zhill I already used this model for the other PR so I'll go ahead and change it for the test I am currently working on. Would the solution have been somewhat like my first commit where the test functions are more or less duplicated with the only distinction being the different trigger creators? Or would it be ok to create a function that handles the assertions that gets called from each test

PR #723, opened 2020-11-18
Format code with https://github.com/psf/black

robertp:


dakaneye (approved):
Full disclosure, I did not look at all 339 files. Also I assume the workflow for making git blame continue to work was tested, and that these changes are ONLY black formatting. 

Given that the above is true, this has my full support

robertp:
> Full disclosure, I did not look at all 339 files. Also I assume the workflow for making git blame continue to work was tested, and that these changes are ONLY black formatting.
> 
> Given that the above is true, this has my full support

@dakaneye Hey Sam, those assumptions are true - the git blame config is simple enough to set up, I can work with you any time to get it going.

robertp:
closing this branch/PR, doing another run at this today

PR #715, opened 2020-11-13
Restore content hints functionality for removed analyzer modules

Toure:
Signed-off-by: Toure Dunnon <toure.dunnon@anchore.com>

**What this PR does / why we need it**: The "content hints" functionality should behave as expected for all content types, which have been replaced by Syft.

**Which issue this PR fixes** fixes #703

PR #710, opened 2020-11-10
Replace dpkg analyzer module with syft call

wagoodman:
Replaces the dpkg analyzer module with a syft call.

Notable differences in behavior:
- The file list does not include directories

Note: tests will fail until https://github.com/anchore/syft/pull/257 is merged and a new syft release is cut and pinned here.

Partially addresses #682

PR #701, opened 2020-11-05
Rpm handler syft package.

Toure:
**What this PR does / why we need it**:
Created a rpm handler for anchore-engine to interface with syft, this will replace a portion of the current 10_packages_list module with a call into syft.   https://github.com/anchore/anchore-engine/issues/682

PR #694, opened 2020-10-30
tests: add functional analyzer tests for debian packages

alfredodeza:
**What this PR does / why we need it**: Adds functional tests for Debian packages. Part of the OS coverage in analyzers

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: Related to the work needed to complete https://github.com/anchore/anchore-engine/issues/682

**Special notes**: This skips OSX testing because there is a requirement of `dpkg` to be present in the system. This `skip` should get removed once `syft` gets in place.

PR #738, opened 2020-11-24
Replace select package-based analyzer modules with Syft

wagoodman:
**Overall changes:** 
- this replaces select analyzer modules with Syft invocations (DPKG, RPM, BusyBox, Alpine, Java, Python, NPM, Gem)
- upgrades the function test harness to run within a docker compose stack in CircleCI on a machine executor.
- adds several functional tests around the existing analyzer module document format
- the analyzer module input image type is always an OCI dir type (previously was multiple types)

**Incorporates the following PRs:**

DPKG support:
- https://github.com/anchore/anchore-engine/pull/710
- https://github.com/anchore/anchore-engine/pull/694

Alpine support:
- https://github.com/anchore/anchore-engine/pull/697
- https://github.com/anchore/anchore-engine/pull/695

Java support:
- https://github.com/anchore/anchore-engine/pull/693
- https://github.com/anchore/anchore-engine/pull/690

Python support:
- https://github.com/anchore/anchore-engine/pull/685
- https://github.com/anchore/anchore-engine/pull/669

Busybox support:
- https://github.com/anchore/anchore-engine/pull/714

RPM support:
- https://github.com/anchore/anchore-engine/pull/696
- https://github.com/anchore/anchore-engine/pull/701

NPM support:
- https://github.com/anchore/anchore-engine/pull/689
- https://github.com/anchore/anchore-engine/pull/675

Gem support:
- https://github.com/anchore/anchore-engine/pull/651
- https://github.com/anchore/anchore-engine/pull/679

General improvements:
- https://github.com/anchore/anchore-engine/pull/646
- https://github.com/anchore/anchore-engine/pull/715
- https://github.com/anchore/anchore-engine/pull/684


pytest.ini
```diff
@@ -5,3 +5,5 @@ log_cli_level=WARN
 log_format=%(asctime)s %(levelname)s %(message)s
 log_date_format=%Y-%m-%d %H:%M:%S
 junit_family=xunit1
+# ease debugging and fast-fail in CI
```
alfredodeza, line 8: I'm going to be -1 on this, because although it is useful when debugging, one is not always debugging in CI. I've been in situations before when I do need to debug in CI, so I ad-hoc add this to debug, and later I revert the commit.

scripts/ci/Dockerfile.functional
```diff
@@ -1,12 +1,9 @@
...
+ARG TEST_IMAGE_NAME
+FROM ${TEST_IMAGE_NAME}
 
 USER root
-RUN set -ex && \
-    echo "installing OS dependencies" && \
-    yum update -y && \
-    yum install -y gcc make git python3-wheel python36-devel
+
+RUN yum install -y make gcc python38 python38-devel python36 python36-devel
```
alfredodeza, line 6: Not sure why this is no longer installing `git` and why it removed the title and `-ex`

tests/functional/local.env
```diff
@@ -9,4 +9,5 @@ export ANCHORE_TEST_DOCKER_REGISTRY_HOST=localhost:5000
 export ANCHORE_BASE_URL=http://localhost:8228/v1
 export ANCHORE_CLI_URL=http://localhost:8228/v1
 export ANCHORE_CLI_USER=admin
-export ANCHORE_CLI_PASS=foobar
\ No newline at end of file
+export ANCHORE_CLI_PASS=foobar
+export TOX_ENV=py36
```
alfredodeza, line 13: this abstraction is making it obscure how/where/what version of Python we are testing. The levels of indirection look like:

1. Makefile
2. scripts/ci/test-functional
3. tox
4. local.env

Someone running functional tests now needs to be aware of a spot that sets these, which is very surprising to anyone not used to running these tests with make

alfredodeza (changes requested):

scripts/ci/Dockerfile.functional
wagoodman: I'll put it back (was it needed to begin with? it doesn't seem like it was)

tests/functional/local.env
wagoodman: It would be nice to get some time to simplify the entire setup, but, this PR is not introducing any of these abstractions (introduced https://github.com/anchore/anchore-engine/pull/679 ). 

Background: The reason for the change is due to the fact that the functional tests are not run within the build docker image, so dependencies were being independently pinned and installed in different places (skopeo, anchore-cli, etc), leading to a fundamental issue when introducing syft. The location where the tests were running needed to be based on the same production artifact being produced (not running on a CI host).

As an added nicety for folks that wanted to run functional tests locally the `local.env` was added --not used in CI. In CI the job parameters control which python version is used (introduced https://github.com/anchore/anchore-engine/pull/698 , and was a conflict in the rebase). 

There was team consensus on this approach and it's use is documented within `DEVELOPING.md` (which I've updated here as well).

scripts/ci/Dockerfile.functional
alfredodeza: Well I don't know either, so I was hoping you would enlighten the removal :)

tests/functional/local.env
alfredodeza: > It would be nice to get some time to simplify the entire setup, but, this PR is not introducing any of these abstractions (introduced #679 ).

I understand these got introduced previously because I raised my concerns in that PR as well. What my suggestion is here is to _avoid adding more_ to this file in favor of defaults. In this particular case, the default are already one abstraction level up, in the script that gets called by `make`.

> 
> Background: The reason for the change is due to the fact that the functional tests are not run within the build docker image, so dependencies were being independently pinned and installed in different places (skopeo, anchore-cli, etc), leading to a fundamental issue when introducing syft. The location where the tests were running needed to be based on the same production artifact being produced (not running on a CI host).

I agree that having dependencies installed in a normalized and repeatable way within the container that is getting tested is the way to go.

> 
> As an added nicety for folks that wanted to run functional tests locally the `local.env` was added --not used in CI. In CI the job parameters control which python version is used (introduced #698 , and was a conflict in the rebase).
> 

Right, this looks like a developer-only change, which is not related to getting tests passing on CI. As a tox user, I do know how to run tests and set the particular Python version I need. Having to source a file that sets that for me is surprising to say the least - even if documented.

alfredodeza (approved):
my comments should not prevent this from moving forward

tests/functional/local.env
wagoodman: I hear you regarding sane defaults as a nicety --right now this branch took out multiple layers of defaults that were hiding and caused hours of troubleshooting, if we put them back in my only request is to try to put the defaults in a single spot (wherever that should be).

Please feel free to change this in a way that makes more sense! I'm all for it.

PR #740, opened 2020-11-25
logging support for syft calls

### LINKED ISSUE #722: Hook up syft output to engine logger within analyzer/worker.
The way `syft` is currently consumed within Anchore Engine eats up all logging and relies on exit status codes to determine success or failure. In addition of exit status codes, it is useful to capture output from `syft` and inject that output into Anchore Engine.

In the event of non-zero exit status, it is specially useful since errors can be reported and captured.

alfredodeza:
**What this PR does / why we need it**: Adds logging support for `syft` calls via `subprocess.Popen`

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: closes #722 

**Special notes**:

* Creates a new `run_check()` helper to handle the logging. There are other utilities (like `skopeo_wrapper`) that can now use this, but this PR didn't update them to keep the down the size/scope.

* Adds a new exception to capture errors from failed `Popen()` calls to replace the other exceptions that had lots of information that was not used at all

* Removes the try/except blocks with exceptions that were granular but are rewritten at the end of `localanchore_standalone`

* Adds tests to the new utility to ensure that the logging calls are done automatically: always logs stdout at DEBUG levels, and adds stderr only when there are non-zero exit status at ERROR level

Note: DNM for now, since the target branch is about to be merged into master

### DIFF
--- a/anchore_engine/clients/syft_wrapper.py
+++ b/anchore_engine/clients/syft_wrapper.py
@@ -2,13 +2,7 @@ import os
 import json
 import shlex
 
-from anchore_engine.utils import (
-    run_command,
-    run_command_list,
-    manifest_to_digest,
-    AnchoreException,
-)
-from anchore_engine.subsys import logger
+from anchore_engine.utils import run_check
 
 
 def run_syft(image):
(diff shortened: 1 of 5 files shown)


Toure (approved):

anchore_engine/utils.py
```diff
@@ -235,6 +235,42 @@ def run_command_list(
     return pipes.returncode, stdout_result, stderr_result
 
 
+def run_check(cmd, **kwargs):
```
wagoodman, line 238: this isn't capturing syft stderr, which is all of the logger (debug/status) output.

anchore_engine/clients/syft_wrapper.py
```diff
@@ -23,45 +17,6 @@ def run_syft(image):
...
 
-    logger.debug("running syft: cmd={}".format(repr(cmd)))
-    rc, stdout, stderr = run_command_list(shlex.split(cmd), env=proc_env)
-    logger.debug(
-        "results: rc={} stdout={} stderr={}".format(rc, repr(stdout), repr(stderr))
-    )
-
-    if rc != 0:
-        raise SyftError(cmd=cmd, rc=rc, out=stdout, err=stderr)
+    stdout, _ = run_check(shlex.split(cmd), env=proc_env)
```
wagoodman, line 20: related to https://github.com/anchore/anchore-engine/pull/740/files#r532760445 , if not logged there, this would be the next best spot to log syft output, but there currently is none.

anchore_engine/utils.py
alfredodeza: This function _is_ capturing `stderr`. Could you point to where you would expect to capture `stderr` that this function is missing?

This function logs the `stdout` always to DEBUG level, and `stderr` to ERROR level when there is a non-zero exit status. For production usage, I don't expect anyone to configure logging at DEBUG where (as you point out) it could potentially be in the MB range for some images.

wagoodman: what I meant to point out is that the syft logging output (to stderr) is not being logged with the engine logger in all cases --here only when the return code != 0 (not what was intended with the issue).

But on second read, this is probably alright behavior --you will get all of the information you would want out of the logs for when something is going wrong.

(also, ack on the MB length and your debug comment... just as long as we know what will happen going in 👍 )

wagoodman (approved):

anchore_engine/utils.py
alfredodeza: this has been modified to:

* always log both stderr and stdout at DEBUG levels
* conditionally log stderr at ERROR level if the level is higher than DEBUG

Note that Engine's logging handling is backwards, so when I say _"higher than DEBUG"_ it is translated in code as _"lower than DEBUG"_

wagoodman: how fun! thanks for the helpful comment

wagoodman (approved):

PR #741, opened 2020-11-25
Update policy engine feeds marshalling logic to prevent double marshalling

zhill:
Update to fix failures in response marshalling in the policy engine due to 0.9.0 dev changes. This never affected a released version.


nightfurys (approved):
lgtm, PR good for merge after resolving the merge conflict

PR #742, opened 2020-11-25
One more pass on code formatting with https://github.com/psf/black

robertp:


robertp:
I forgot to do this as Anchore Coder... redoing.

PR #744, opened 2020-11-26
intial implementation of moving image content retrieval to catalog

dakaneye:

**What this PR does / why we need it**: This moves the image content retrieval logic to the catalog where it makes more sense to be in (because that's the service responsible for storage)

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #: N/A

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/api/controllers/default_controller.py
+++ b/anchore_engine/services/catalog/api/controllers/default_controller.py
@@ -1,19 +1,18 @@
 import connexion
 
 import anchore_engine.apis
-from anchore_engine import db
-import anchore_engine.services.catalog.catalog_impl
 import anchore_engine.common
+import anchore_engine.configuration.localconfig
+import anchore_engine.services.catalog.catalog_impl
+import anchore_engine.subsys.servicestatus
+from anchore_engine import db
+from anchore_engine.apis.authorization import get_authorizer, INTERNAL_SERVICE_ALLOWED
+from anchore_engine.apis.context import ApiRequestContextProxy
+from anchore_engine.common import helpers
+from anchore_engine.common.helpers import make_response_error
+from anchore_engine.services.catalog import archiver, CatalogService
 from anchore_engine.services.catalog.archiver import ImageConflict
 from anchore_engine.subsys import logger
-import anchore_engine.configuration.localconfig
-import anchore_engine.subsys.servicestatus
-from anchore_engine.clients.services import internal_client_for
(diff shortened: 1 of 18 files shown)


anchore_engine/common/helpers.py
```diff
@@ -1,12 +1,246 @@
...
+import json
+import stat
+
+import anchore_engine.configuration.localconfig
+from anchore_engine import utils
 from anchore_engine.subsys import logger
+from anchore_engine.common import os_package_types
+
+
+def make_image_content_response(content_type, content_data):
```
dakaneye, line 17: All the methods in this file are copied over from the apiext helpers, to be more explicit about them being used in potentially multiple services

anchore_engine/services/catalog/api/controllers/default_controller.py
```diff
@@ -196,6 +195,27 @@ def get_image(imageDigest):
...
 
 
+@flask_metrics.do_not_track()
+@authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
+def get_image_content(image_digest, content_type):
+    return_object = {}
+    httpcode = 500
+
+    try:
+        image_content_data = CatalogService.get_image_content(
```
dakaneye, line 205: @zhill , is there a way to lookup the service dynamically rather than using a class method here? I'd like to make this compatible with any extensions of the Catalog Service

anchore_engine/common/helpers.py
dakaneye: They could use a refactor as well but didn't want to do too much in this PR

Toure (approved):

anchore_engine/services/catalog/image_content/get_image_content.py
```diff
@@ -0,0 +1,152 @@
...
+
+import anchore_engine
+from anchore_engine import utils, db
+from anchore_engine.apis.exceptions import ResourceNotFound, BadRequest
+from anchore_engine.common.helpers import make_anchore_exception
+from anchore_engine.db import db_catalog_image
+from anchore_engine.subsys import taskstate, logger
+
+
+class ImageContentGetter:
```
zhill, line 14: Good approach for abstraction here. Enables future use of caches, etc easily (e.g. CachingImageContentGetter...)

zhill (approved):
looks good

Issue #229, opened 2019-07-10
Support path based exclusions for policy evaluations


KnoxAnderson:
Feature Request:

There are sometimes many CVE's that are associated with a particular JAR ie. (a JAR of a 3rd party monitoring agent) instead of whitelisting all of them (and actually whitelisting issues) can we exclude scanning of certain paths.

emaildanwilson:
This seems like a reasonable request. We're hitting this issue with the NewRelic agent jar file which contains additional jar files for each version of various software packages it instruments. these jar files happen to be named in a way which they match the CVE records for the actual software packages which triggers false positives. It would be nice if we could wildcard filter on this monitoring jar instead of unnecessarily whitelisting the other software packages. The jar is downloadable [here](https://download.newrelic.com/newrelic/java-agent/newrelic-agent/current/newrelic-java.zip) if anyone wants to try this for themself.

```Vulnerability ID        Package                        Severity        Fix         Vulnerability URL                                      
CVE-2012-6612           solr-4.0.0                     High            None        https://nvd.nist.gov/vuln/detail/CVE-2012-6612         
CVE-2014-3490           resteasy-3.0                   High            None        https://nvd.nist.gov/vuln/detail/CVE-2014-3490         
CVE-2015-0225           cassandra-2.1.2                High            None        https://nvd.nist.gov/vuln/detail/CVE-2015-0225         
CVE-2016-3092           tomcat-8.5.0                   High            None        https://nvd.nist.gov/vuln/detail/CVE-2016-3092         
CVE-2016-3092           tomcat-8.5.2                   High            None        https://nvd.nist.gov/vuln/detail/CVE-2016-3092         
...
```

emaildanwilson:
It appears that for at least the JAR file use case a filter could be accomplished by allowing for an override of this [regex](https://github.com/anchore/anchore-engine/blob/6ce0caf269b70eb65566914ecb0c7eeac0a11b29/anchore_engine/analyzers/modules/32_java_packages.py#L16).

For example, it could be overridden to `"(.*)\.([jwe]ar|[jh]pi)(?<!newrelic\.jar)$"` as a way to exclude `newrelic.jar`.

A more global approach would be better if I can figure out where it would go.

@zhill since you helped out with my prior PR maybe you can also help here? What do you think about adding a global filter like is being proposed here? Who should be involved in the discussion?

zhill:
[...]

PR #751, opened 2020-12-01
feat: ability to exclude packages by their path

ssthom:
Signed-off-by: Shawn Thompson <ssthomps@us.ibm.com>

Adds the ability for users to exclude packages in the policy bundle based on the package path

```
{
  "name": "package_path_exclude",
  "value": ".*newrelic.*"
}
```

**Which issue this PR fixes** 

fixes #: https://github.com/anchore/anchore-engine/issues/229

**Special notes**:


zhill:
Hi @ssthom thanks for this! We'll give it a look. The approach makes sense.

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -282,6 +296,11 @@ def evaluate(self, image_obj, context):
                                 parameter_data["pkg_class"] = "non-os"
                                 parameter_data["pkg_type"] = image_cpe.pkg_type
 
+                                if package_path_re is not None:
+                                    match_found = package_path_re.match(image_cpe.pkg_path)
+                                    if match_found is not None:
+                                        continue
```
dakaneye, line 302: I would consider adding at least a debug log here. When a package gets excluded by a path it might be helpful for troubleshooting later on.

dakaneye (changes requested):
@ssthom Great job submitting this change! Looks like you did things in a consistent manner across the board, and the logic seems sound.

My debug log comment is optional, however I'd like to request the addition of a unit test for the package exclusion as part of this change. Thanks!

ssthom:
@dakaneye Added debug statement, but I'm new to the repo could you point me to an existing test to use as a basis or in the right direction. I didn't see any vulnerability gate tests in https://github.com/anchore/anchore-engine/tree/master/tests/unit/anchore_engine/services/policy_engine/policy/gates

dakaneye:
> @dakaneye Added debug statement, but I'm new to the repo could you point me to an existing test to use as a basis or in the right direction. I didn't see any vulnerability gate tests in https://github.com/anchore/anchore-engine/tree/master/tests/unit/anchore_engine/services/policy_engine/policy/gates

@ssthom , I saw some [here](https://github.com/anchore/anchore-engine/blob/master/tests/integration/services/policy_engine/engine/policy/gates/test_vulnerabilities.py), sorry looks like they're actually integration tests, my bad

zhill:
Ack, we'll review tests as we add unit tests for this as part of unit test coverage for the gate overall.

PR #755, opened 2020-12-02
Fix for Issue 712 - Remove DB column lengths for policy engine tables

### LINKED ISSUE #712: Remove varchar column length constraints in policy engine tables
The definitions are a hold-over from a long-ago project using MySQL that required those column definitions with length constraints. PostgreSQL has no such requirements and the limits can cause insertion errors when image content violates the lengths.

Remove all length constraints and include upgrade code to remove the sizes during the db upgrade process.

Causes #649 and other failures.

zhill:
Removes column length constraints in policy engine tables.

Fixes #712


anchore_engine/db/entities/upgrade.py
```diff
@@ -1183,6 +1183,245 @@ def db_upgrade_012_013():
...
+        # FixedArtifact
+        ('feed_data_vulnerabilities_fixed_artifacts', 'vulnerability_id'),
+        ('feed_data_vulnerabilities_fixed_artifacts', 'namespace_name'),
+        ('feed_data_vulnerabilities_fixed_artifacts', 'name'),
+        ('feed_data_vulnerabilities_fixed_artifacts', 'version'),
+        ('feed_data_vulnerabilities_fixed_artifacts', 'version_format'),
+        ('feed_data_vulnerabilities_fixed_artifacts', 'epochless_version'),
+
+        # CpeVulnerability
+        # ('feed_data_cpe_vulnerabilities', 'feed_name'),
```
dakaneye, line 1239: should this be commented out?

anchore_engine/db/entities/upgrade.py
```diff
@@ -1183,6 +1183,245 @@ def db_upgrade_012_013():
...
+    for table_name, column_name in to_update:
+        log.err('Updating table {} column {} to varchar'.format(table_name, column_name))
+        try:
+            rc = engine.execute('ALTER TABLE {} ALTER COLUMN "{}" type varchar'.format(table_name, column_name))
+        except Exception as ex:
+            log.err('Failed updating {} column {} type to varchar. Err: {}'.format(table_name, column_name, str(ex)))
+            raise ex
+
+
+def db_upgrade_013_014():
```
dakaneye, line 1422: This will conflict w/ the global max images / additional archival rule PR

dakaneye (changes requested):

anchore_engine/db/entities/upgrade.py
zhill: There was an issue in the postgres casting conversion on this table. This is a deprecated type (superceded by the nvdv2 tables) so I'm not worried about it not getting the update as this is also data that we control from the feed service rather than image content.

anchore_engine/db/entities/upgrade.py
zhill: Yes, I've resolved it in the rebase operation to move your code to a new function name and call that function from this function.

PR #761, opened 2020-12-03
Issue 720 - Support loading more than one policy bundle on account creation

### LINKED ISSUE #720: Support loading more than one policy bundle on account creation
As an engine admin I would like to be able to have new accounts in Engine initialized with more than one policy bundle so I can provide users with additional options without out-of-band distribution.

MVP is to provide loading from an on-disk location that I can mount externally or update using a new container image with the defaults updated.

To provide guidance on which bundle should be the default active bundle, the filename of the bundles will be used to indicate the default to be active. This file must be named 'default_active_bundle.json'.

dspalmer99:
Adds the ability to add multiple policy bundles by default. On startup, rather than looking for a single policy bundle file in the included config files, the system will now load all policy bundles in the configured bundle dir (which defaults to `conf/bundles/`). The filename of the default active bundle remains unchanged, but it is located in `conf/bundles/` now instead of `conf/`.

**Which issue this PR fixes**
fixes #720


anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2517,4 +2518,33 @@ def add_event_json(event_json, dbsession, quiet=True):
...
+def load_policy_bundles(config, process_bundle, process_exception):
+    """
+    A convenience function to avoid code duplication between accounts.py and catalog/__init.py. This
+    function iterates through the one to many policy bundle filepaths in the config, opens them, and
+    converts them to json. Since the (currently two) calling methods do slightly different things with
+    those bundles (and handle exceptions in slightly ways) this function requires two callbacks it can
+    call to do that processing.
+
+    :param config The config:
+    :param process_bundle A callback with logic for what to do with each bundle:
```
dspalmer99, line 2530: FWIW I'm generally avoid callbacks because I think it can make the code harder to read (although these are pretty lightweight). If in the future we can further refactor `accounts.py` and move more of that logic into the `catalog` we could get rid of these.

anchore_engine/services/apiext/api/controllers/accounts.py
```diff
@@ -23,6 +23,7 @@
 from anchore_engine.db.db_account_users import UserAlreadyExistsError, UserNotFoundError
 from anchore_engine.utils import datetime_to_rfc3339
 from anchore_engine.common.helpers import make_response_error
+from anchore_engine.services.catalog.catalog_impl import load_policy_bundles
```
zhill, line 26: Can we move this code into an "anchore_engine.subsys" location or part of "localconfig"? I'd like to avoid one service importing code from another service.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2541,4 +2542,33 @@ def add_event_json(event_json, dbsession, quiet=True):
             raise
 
 
+def load_policy_bundles(config, process_bundle, process_exception):
```
zhill, line 2545: Per comment above, let's move this to a general location. With the callback it seems able to be moved to a general spot.

zhill (changes requested):
A few changes in code structure, but functionality looks good. Will want to be sure the commit message for the merge highlights this changes the load path for including external bundles if anyone has a volume mount etc to provide these outside the container.

dspalmer99:
Thanks @zhill , I've made the requested changes and retested locally. Looks like I have a linting issue now but I will fix that next.

dspalmer99:
@zhill I've made those changes, manually tested locally, and the CI is passing now. If you're good with this now I can merge it, with a comment about the changed bundle path as you asked for above.

zhill (approved):
looks good. i'll merge it!

PR #763, opened 2020-12-03
[DO NOT MERGE] Imports API swagger updates

wagoodman:
The target base branch will be changed to the imports API branch in the future. This draft PR is to show the suggested changes so far (and when we change the target base branch the difference between the original document provided by @zhill  will be more apparent in the code diff).

Summary of changes:
- Ensures that tags in the swagger document are singular. The OpenAPI code generation tool assumes that the tags will be used to hint which service object an API function should be provided under. If multiple tags are provided these functions are duplicated, which is incorrect behavior. The source of truth for an API should be with a static swagger document instead of forcing consumers to modify it for use. (shout out if there is a good technical reason for multiple tags if I missed it, I could only see that it would change the way it's presented on a swagger page and really nothing else)
- Adds `SyftLocation` and `SyftPackage` definitions
- Specifies for import endpoints that take plain text or raw json bytes that the body content is required. This allows for the generated code to accept a body variable.

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -495,7 +495,7 @@ paths:
       x-swagger-router-controller: anchore_engine.services.apiext.api.controllers.images
       operationId: import_image_archive
       x-anchore-authz-action: importImage
-      summary: Import an anchore image tar.gz archive file.
+      summary: Import an anchore image tar.gz archive file. This is a deprecated API replaced by the "/imports/images" route
       consumes:
         - multipart/form-data
       parameters:
@@ -610,7 +610,6 @@ paths:
   /images/{imageDigest}/check:
     get:
       tags:
-      - Images
       - Policy Evaluation
       x-swagger-router-controller: anchore_engine.services.apiext.api.controllers.images
       operationId: get_image_policy_check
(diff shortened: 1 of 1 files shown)


anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -495,7 +495,7 @@ paths:
       x-swagger-router-controller: anchore_engine.services.apiext.api.controllers.images
       operationId: import_image_archive
       x-anchore-authz-action: importImage
-      summary: Import an anchore image tar.gz archive file.
+      summary: Import an anchore image tar.gz archive file. This is a deprecated API replaced by the "/imports/images" route
```
zhill, line 498: Yes, this is planned. I haven't updated it yet in my branch.

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -2911,6 +3163,11 @@ definitions:
         type: string
         description: The image digest identify the analysis. Archived analyses are based on digest, tag records are restored as analysis is restored.
         pattern: "^sha256:[a-fA-F0-9]{64}$"
+  LocalAnalysisSource:
```
zhill, line 3166: What is the purpose of this? I don't see it referenced anywhere

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -2970,9 +3227,6 @@ definitions:
         type: string
       trigger_id:
         type: string
-      expires_on:
```
zhill, line 2973: Yeah, don't want to remove this. It's part of another feature...just highlighting this so we for sure do not merge :)

wagoodman:
this branch has been superseded by https://github.com/anchore/anchore-engine/tree/new_image_import

PR #764, opened 2020-12-03
Global max images

### LINKED ISSUE #700: Add archive rule for total image count
Add ability to set per-account or system global archive rule to archive all but the N newest images in the system.

This facilitates simpler control of the total db size usage deployment size without deleting images but rather archiving them when they fall outside boundary for the N-newest.

dakaneye:
**What this PR does / why we need it**: Fixes #700 

**Special notes**:

### DIFF
--- a/anchore_engine/db/db_catalog_image.py
+++ b/anchore_engine/db/db_catalog_image.py
@@ -1,9 +1,7 @@
-from sqlalchemy import desc, and_, or_
-from sqlalchemy.orm import load_only, Load
-import time
+from sqlalchemy import desc, and_, or_, func
 
-from anchore_engine import db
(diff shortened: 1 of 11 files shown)


zhill:
Will do a closer review, but looks like you're missing the upgrade code for the additional db column added.

dakaneye:
> Will do a closer review, but looks like you're missing the upgrade code for the additional db column added.

@zhill You're right, added now

dakaneye:
This needs another change to expose the new fields to the API, and found some problems with the actual rule processing, gonna push another commit

dakaneye:
Merging with verbal permission from @zhill

PR #765, opened 2020-12-03
Fix files gate and add unit tests

### LINKED ISSUE #756: ContentMatchTrigger in file gate incorrectly compares encoded values with decoded ones
**Explanation of Issue***
The ContentMatchTrigger in file gate will fail to properly fire due to logical errors with how the user supplied `regex_name` param is compared to the values in the artifact_analysis record. See [file](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/engine/policy/gates/files.py). The `regex_name` string is encoded and stored in a var called `match_decoded`. The values from the artifact_analysis are decoded and then there are multiple conditions that end up comparing the encoded values with the decoded ones that will never result in the trigger to properly fire even when the values both decoded are identical.

1.  create the following analyzer config block:
```
content_search:
  match_params:
    - MAXFILESIZE=10000
    - STOREONMATCH=n
  regexp_match:
    - "AWS_ACCESS_KEY=(?i).*aws_access_key_id( *=+ *).*(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9]).*"
    - "AWS_SECRET_KEY=(?i).*aws_secret_access_key( *=+ *).*(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{40}(?![A-Za-z0-9/+=]).*"
    - "PRIV_KEY=(?i)-+BEGIN(.*)PRIVATE KEY-+"
    - "DOCKER_AUTH=(?i).*\"auth\": *\".+\""
    - "API_KEY=(?i).*api(-|_)key( *=+ *).*(?<![A-Z0-9])[A-Z0-9]{20,60}(?![A-Z0-9]).*"
```

2. In the policy bundle create a file gate with ContentMatchTrigger and set the `regex_name` to be `AWS_ACCESS_KEY`. 

3. Run an image analysis on `anchore/test_images:secret_match_fake_v1` and evaluate it with policy bundle

Because the image content_search analyzer will match the `AWS_ACCESS_KEY`, this should cause the trigger to fire, but it passes

zburstein:
Fixes #756 by modeling after the secrets gate and adds unit tests for the entire files gate

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gates/files.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gates/files.py
@@ -27,18 +27,17 @@ class ContentMatchTrigger(BaseTrigger):
     )
 
     def evaluate(self, image_obj, context):
-        match_filter = self.regex_name.value()
+        match_decoded = self.regex_name.value()
 
-        if match_filter:
-            match_decoded = ensure_str(base64.b64encode(ensure_bytes(match_filter)))
-        else:
-            return
+        if match_decoded:
+            match_encoded = ensure_str(base64.b64encode(ensure_bytes(match_decoded)))
 
         for thefile, regexps in list(context.data.get("content_regexp", {}).items()):
             thefile = ensure_str(thefile)
             if not regexps:
                 continue
             for regexp in regexps.keys():
+                found = False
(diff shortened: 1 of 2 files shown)


zhill (approved):
lgtm. good catch

Issue #725, opened 2020-11-19
Potential issues with NPM and Gem gate triggers returning when feed data present


zburstein:
Identified a potential bug while writing a unit test for GemCheckGate and NpmCheckGate that looks like it might cause issues with the ability for triggers to fire when there is feed data available. I have not yet tested this with a running instance yet but there appears to be an issue in the logic

When both gates are prepared they assign two data points to the context. One corresponds to the gems/packages in the image (context.data[GEM_LIST_KEY] and context.data[NPM_LISTING_KEY] respectively). Next the corresponding feed data from the database for the gems/packages found in the image are retrieved and stored in context.data[GEM_MATCH_KEY] and context.data[NPM_MATCH_KEY] respectively. Almost every trigger for these gates initiates the evaluation by assigning the above mentioned context data points to `img_data` and `feed_data` respectively. The issue occurs with the following condition (see [link](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/engine/policy/gates/gems.py#L28))

```
if feed_gems or not img_gems:
    return
```

Logically, I believe this means that whenever feed data is available, the trigger returns before it actually does anything to process the trigger and compare the img and feed data. In most cases the code following this condition directly relies on the presence of the feed_data as well. I believe that the solution should be 

```
if not feed_gems or not img_gems:
    return
```

PR #767, opened 2020-12-04
Fix npms gems gate and add tests

zburstein:
Fixes the bug identified by #725 in both the gems and npms gates and adds unit tests for both to test the fix and expand unit test coverage


anchore_engine/services/policy_engine/engine/policy/gates/gems.py
```diff
@@ -44,7 +44,7 @@ def evaluate(self, image_obj, context):
                 if v and v != feed_names.get(gem):
                     self._fire(
                         msg="Package ({}) version ({}) installed but is not the latest version ({})".format(
-                            gem, v, feed_names[gem]["latest"]
+                            gem, v, feed_names[gem]
```
zhill, line 47: good fix here as well

zhill (approved):
lgtm

PR #769, opened 2020-12-07
Add place where cpes can get updated if necessary

dakaneye:


dakaneye:
[Do Not Merge] Need to update submodule to correct branch

anchore_engine/services/policy_engine/__init__.py
```diff
@@ -413,6 +413,6 @@ class PolicyEngineService(ApiService):
 
     __lifecycle_handlers__ = {LifeCycleStages.pre_register: [(process_preflight, None)]}
 
-    # def _register_instance_handlers(self):
-    #    super()._register_instance_handlers()
-    #    self.register_handler(LifeCycleStages.pre_register, process_preflight, None)
+    def get_cpe_vulnerabilities(self, image, nvd_cls, cpe_cls):
```
zhill, line 416: This works, but I'd prefer a way to use the service definition to set a function reference elsewhere in the config to avoid requiring that the injection point have access to the service object itself to do this lookup.

For example, deeper in the policy engine code we may not want a back-ref to the ApiRequestContext to get the service object, but a "CpeVulnerabilityProvider" singleton is cleanly decouples them.

zhill (approved):
Small suggestion, but not blocking.

PR #772, opened 2020-12-08
Include fix version in check output message if a fix is available reg…

nightfurys:
…ardless of the gate parameter checks

**Special notes**:

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
@@ -282,6 +282,13 @@ class VulnerabilityMatchTrigger(BaseTrigger):
                                 parameter_data["pkg_class"] = "non-os"
                                 parameter_data["pkg_type"] = image_cpe.pkg_type
 
+                                # setting fixed_version here regardless of gate parameter,
+                                fix_available_in = vulnerability_cpe.get_fixed_in()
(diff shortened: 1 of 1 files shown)


dspalmer99 (approved):
LGTM.

For my own information though, what is the rationale behind this change? Is it to ensure the field always exists, even if it is empty in the case where no fix is available?

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -382,17 +389,19 @@ def evaluate(self, image_obj, context):
...
                                     # explicit fix state check matches fix availability
                                     if is_fix_available == (
                                         fix_available_in is not None
                                         and len(fix_available_in) > 0
                                     ):
-                                        if is_fix_available:
-                                            parameter_data["fixed_version"] = ", ".join(
-                                                fix_available_in
-                                            )
+                                        # TODO this instantiates fix version later used for the output message only if the approrpiate gate parameter checks pass. check with Zach why that's necessary. why can't the fix version be made available regardless of the gate params
```
zhill, line 399: IIRC this was added to ensure parameters in the rule are clear when viewing the result. However, this current impl is not good and we'll overhaul the approach when we update the policy evaluation API response. It should be safe to include this regardless.

zhill (approved):
lgtm

PR #773, opened 2020-12-08
Have skopeo remove signatures on signed images when copying them

### LINKED ISSUE #711: Image analysis for signed images fails because skopeo fails to copy content if the image is signed
The analyzer cannot pull image content for signed images because it uses skopeo and skopeo will refuse to copy a signed image to a destination transport that does not support copying signatures, and the "oci" transport destination does not support signatures.

We can mitigate this by adding the `--remove-signatures` option to the skopeo command invocation to ensure the copy will not fail due to the presence of signatures.

dspalmer99:
Have skopeo remove signatures on signed images when copying them. This is needed because the destination currently does not support signed images, so skopeo will currently refuse to copy them.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #711

**Special notes**:

### DIFF
--- a/anchore_engine/clients/skopeo_wrapper.py
+++ b/anchore_engine/clients/skopeo_wrapper.py
@@ -63,9 +63,14 @@ def manifest_to_digest_shellout(rawmanifest):
     return ret
 
 
-def copy_image_from_docker_archive(source_archive, dest_dir):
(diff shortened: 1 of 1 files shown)


zhill (approved):
lgtm

PR #779, opened 2020-12-12
New APIs for uploading analysis data from syft into engine with a multi-phase API

### LINKED ISSUE #783: Implement new import API to support uploads from syft
Add new API to support performing local analysis of images with [Syft](http://github.com/anchore/Syft) and uploading that to the engine to import as an image.

Initial implementation will only support packages and metadata analysis, with deeper analysis data available as becomes supported by Syft itself.

The new API is json and text based and the data is processed by the analyzer worker so that imports are just like analysis jobs that get scheduled and can be throttled by the analyzer count (to limit impact on policy engine etc).

zhill:
Adds a new /imports/images API routes that use an operation id (e.g. transaction id) to associate specific content uploads and then actuate the image being imported using a new option in POST /images. Fixes #783 

Also brings in other refactors and consolidations to help debugging, logging, and clarity on the data flow internally.


anchore_engine/db/entities/common.py
```diff
@@ -51,12 +51,12 @@ class should override this function with an impl that serializes those types pro
...
         """
+
+        # TODO: should normalize DateTime's to use RFC3339 in the future to make the timezone explicit
         return dict(
             (
                 key,
-                value
-                if type(value) != datetime.datetime
-                else datetime_to_rfc3339(value),
+                value if type(value) != datetime.datetime else value.isoformat(),
```
nightfurys, line 59: @zhill, timestamp values in the api response will not contain UTC timezone - 'Z' after this change. Is that the intended purpose of this change?

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+            return None
+
+        return self.image_config.get("architecture")
+
+    def _layer_ids(self):
+        # Filter out all empty layers
+        return [
+            x.get("digest")
+            for x in self.raw.get("layers")
+            if not x.get("empty_layer", False)
```
alfredodeza, line 334: Doesn't look like this needs to fallback to a `False` if it is empty does it? it would still evaluate the same?

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+        ]
+
+    def _history(self):
+        # add support for cases where image metadata does not contain a history element at all
+        rawhistory = copy.copy(self.image_config.get("history"))
+        rawlayers = copy.copy(
+            list(self.raw.get("layers", []))
```
alfredodeza, line 341: why is it cohercing to a list here? is `layers` something that is not a list?

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+        )  # This is carried over from old impl
+
+        if rawhistory is None:
+            # Construct the right number of empty history elements for each layer
+            rawhistory = [{} for layer in rawlayers]
```
alfredodeza, line 346:
```suggestion
            rawhistory = [{}] * len(rawlayers)
```

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+
+        done = False
```
alfredodeza, line 348: why is `done` needed?

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+
+    def _infer_dockerfile(self):
+        dockerfile_contents = "FROM scratch\n"
+        for hel in self._history():
+            patt = re.match(r"^/bin/sh -c #\(nop\) +(.*)", hel["CreatedBy"])
+            if patt:
+                cmd = patt.group(1)
+            elif hel["CreatedBy"]:
+                cmd = "RUN " + hel["CreatedBy"]
+            else:
```
alfredodeza, line 383: same suggestion as before, I think this can continue and then use `+=` on the assignment

alfredodeza (changes requested):

anchore_engine/db/entities/common.py
zhill: This is the original form in master. The change you're seeing is an intermediate commit. Changing it to RFC3339 broke some other cases because it parses with a populated timezone, so things like comparison with datetime.datetime.utcnow() will raise exceptions. I wanted to make the change to standardize on RFC3339 but reverted it due to blast radius of that comparison issue.

anchore_engine/services/analyzer/utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+def get_tempdir(config=None):
+    """
+    Return the configured temp dir from the provided config or loaded from global if no config provided
+
+    """
+    c = get_config() if not config else config
+    try:
+        return c["tmp_dir"]
+    except Exception as err:
```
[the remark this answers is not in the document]
zhill: Agree, this was basically moved code from elsewhere so I was just ensuring the same exception semantics. Most of this file is code that I moved out of the analyzer service __init__.py to be accessible in other places without an import loop. I didn't review/update those functions individually, but agree there is plenty to improve in here.

anchore_engine/services/analyzer/utils.py
```diff
@@ -0,0 +1,126 @@
...
+    Update the analysis and/or image status for the record, one of new_analysis_status and new_image_status must be non-None
+
+    :param client:
+    :param image_digest:
+    :param image_record:
+    :param new_analysis_status: str new analysis_status value, options
+    :param new_image_status: str new image status value, optional
+    :return:
+    """
+    assert image_digest is not None
```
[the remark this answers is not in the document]
zhill: ah, good catch. may be left-over debugging. ValueError is more appropriate here.

anchore_engine/services/catalog/api/controllers/default_controller.py
```diff
@@ -152,15 +167,54 @@ def add_image(
...
-        else:
 
+        elif image_metadata.get("import_manifest"):
+            # Import an image from the upload API
+            try:
+                import_manifest = ImportManifest.from_json(
+                    image_metadata["import_manifest"]
+                )
+            except AnchoreApiError:
+                raise
```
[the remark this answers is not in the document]
zhill: This case can be removed. That function doesn't raise an AnchoreApiError, it's a copy-paste from a previous usage. I'll fix that up.

anchore_engine/db/entities/common.py
nightfurys: Aah, gotcha. The comment makes sense now

anchore_engine/common/__init__.py
```diff
@@ -2,13 +2,13 @@
...
-    "tag_update",
-    "vuln_update",
-    "repo_update",
-    "analysis_update",
-]
+repository_subscription_types = ["repo_update"]
+
+tag_subscription_types = ["policy_eval", "tag_update", "vuln_update", "analysis_update"]
+
+subscription_types = repository_subscription_types + tag_subscription_types
```
nightfurys, line 9: This aligns with the downstream usage where the subs are extended. Note to self - attribute name changes impact the downstream usage that need to be updated

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -376,78 +376,78 @@ paths:
...
+          in: body
+          schema:
+            $ref: "#/definitions/ImageAnalysisRequest"
+        - $ref: "#/parameters/AsAccountParameter"
       responses:
         200:
           description: Successfully added image to analysis queue
           schema:
-            $ref: "#/definitions/AnchoreImageList"
+            $ref: "#/definitions/AnchoreImage"
```
nightfurys, line 406: Placeholder for checking if this impacts clients

nightfurys: Controller logic backing this API returns a list object. So the api definition *was* aligned with the response prior to this change. Whether the api should return a single object or a list is a different question, one that needs to be addressed across the board

https://github.com/anchore/anchore-engine/blob/0e393e6f13e18cb0e0c0f81b199295da732eb054/anchore_engine/services/apiext/api/controllers/images.py#L1412

wagoodman: the plot thickins...

anchore_engine/services/catalog/api/controllers/default_controller.py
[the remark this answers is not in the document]
zhill: fixed

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+                cmd = None
+            if cmd:
+                dockerfile_contents = dockerfile_contents + cmd + "\n"
```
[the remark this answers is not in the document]
zhill: Another: don't change it, just move it update. A lot of these old utils need a thorough revisiting.

anchore_engine/util/docker.py
zhill: the field is a boolean, so added this for consistency rather than True or None, but the effect is the same.

anchore_engine/util/docker.py
zhill: Oh, I like that :)

anchore_engine/services/apiext/swagger/swagger.yaml
zhill: reverted back to the original

anchore_engine/clients/localanchore_standalone.py
```diff
@@ -725,235 +729,143 @@ def retrying_pull_image(
...
-        dockerfile_mode = "Actual"
+    # Not found in the tar directly, so lookup using the index.json
+    blobdir, blobs = get_blob_list(copydir, cachedir)
+    if not blobdir:
+        raise Exception("No blob directory found to read image configuration")
+
+    dfile = nfile = None
+    with open(os.path.join(copydir, "index.json"), "r") as FH:
+        idata = json.loads(FH.read())
+        d_digest = idata["manifests"][0]["digest"].split(":", 1)[1]
```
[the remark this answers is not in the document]
zhill: yes, that's an OCI index file so it has an expected format. This is old code that I just moved, so not a regression/change here just wrapped in a function. All of these until functions could use an update and testing, but beyond the scope of this PR to fix the legacy code.

zhill:
Thanks for the review feedback. I'll try to address most of it in follow-ups since a lot of them were on code that I moved but didn't re-write or change.

anchore_engine/util/docker.py
```diff
@@ -149,3 +151,238 @@ def parse_dockerimage_string(instr, strict=True):
...
+        return self.tag is not None
+
+    def has_digest(self):
+        return self.digest is not None
+
+    def has_id(self):
+        return self.image_id is not None
+
+    def tag_pullstring(self):
+        raise Exception(
```
alfredodeza, line 184: @wagoodman this should be:

```python
if not any([self.registry, self.repository, self.tag]):
    raise Exception("missing one of registry, repository, or tag to construct the pullstring"
    )
```

wagoodman: I went with
```python
        if not all((self.registry, self.repository, self.tag)):
            raise Exception(
                "missing one of registry, repository or tag to construct the pullstring"
            )
```

PR #781, opened 2020-12-14
Refactor catalog eval lookup logic

### LINKED ISSUE #780: Refactor implementation and separate it from api controllers
Catalog logic for looking up policy evaluations should be accessible to the service internally without having to go through the api controllers. Otherwise the controller has to imported in the invoking code which results in import errors due to the authnz setup

nightfurys:
Resolves #780


dakaneye (approved):
Looks to me like a simple lift and shift, besides moving the code between these 2 files, there's no change right?

PR #788, opened 2020-12-16
Remove all bare "except:" statements to at least scope to Exception type

zhill:
Replaces "except:" clauses with "except Exception:" equivalents

### DIFF
--- a/anchore_engine/analyzers/modules/13_retrieve_files.py
+++ b/anchore_engine/analyzers/modules/13_retrieve_files.py
@@ -57,7 +57,7 @@ with tarfile.open(
         thefile = re.sub("^/+", "", name)
         try:
             member = tfl.getmember(thefile)
-        except:
+        except Exception:
             member = None
 
         if member and member.isreg():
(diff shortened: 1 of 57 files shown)


alfredodeza:
This is definitely a step in the right direction, but in most of the changes, it will not change the underlying problem which is the incorrect usage of catch-all exceptions. For example in this case:

```python
            except Exception:
                logger.debug_exception("Login failed")
                raise
```

Any exception is being caught to be re-raised immediately with the only addition being a log line. Most of these re-raised exceptions are also caught elsewhere in the stack, further obfuscating the underlying problems that cause the originating exception.

I think that the best way to improve these is by enforcing code-review constraints that prevent such handling from getting merged, and to set a cadence for the continuous improvement of modules as this PR is doing.

zhill:
@alfredodeza I don't disagree this isn't the full solution. Is this a useful update, or should we just address these inline as we actually resolve the exception paths themselves? I left this is in draft for that reason.

alfredodeza:
This is good in that bare exceptions are expunged. In addition, we need to prevent them from getting in again. An update to the linter check in CI needs to happen to catch the fixes this PR adds. Relevant [warnings are here](http://pylint.pycqa.org/en/1.6.0/features.html#id8), e.g. broad-except (W0703) (needs double checking though). The relevant part of the code is in scripts/ci/lint, specifically this line:

```python
# This is specifically enabling checks because the output is too large to
# fix in one go. As cleanups are done, more checks can be enabled
pylint -s n --disable=all --enable=C0325 anchore_engine
```

The `--enable` portion needs to include the new checks

zhill:
stale

PR #790, opened 2020-12-16
Fix docker tag validation & epoch to RFC3339 extapi reference

wagoodman:
This is a follow up PR to #779 with a couple adjustments:
- add validation condition around docker tag and digest
- changed the type of timestamps on post to /images to expect RFC3339 strings (was epoch before hand)

cc: @zhill

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -3465,11 +3465,11 @@ definitions:
     description: A metadata detail record for a specific image. Multiple detail records may map a single catalog image.
     properties:
       created_at:
-        # server returning epoch
-        type: number
+        type: string
+        format: date-time
       last_updated:
-        # server returning epoch
-        type: number
+        type: string
+        format: date-time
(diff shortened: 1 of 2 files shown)


alfredodeza (approved):

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -3465,11 +3465,11 @@ definitions:
...
       created_at:
-        # server returning epoch
-        type: number
+        type: string
+        format: date-time
       last_updated:
-        # server returning epoch
-        type: number
+        type: string
+        format: date-time
```
zhill, line 3472: And confirming these are changes to the spec to reflect the formats actually being returned by the service as-is. I confirmed this on a running system.

wagoodman: indeed, I tried this out on master with a re-generated client-go and right away the parsing of the response failed due to the these fields. Not a problem since the RFC3339 is the right way 👍

PR #791, opened 2020-12-16
Add image max size validation

### LINKED ISSUE #786: Add ability to limit addition of image for analysis based upon a maximum allowable size
Create a configuration value that allows a user to limit the ability to add an image for analysis based upon the size of the image. Needs to be able to handle all paths that add image for analysis, including the repo and image watchers

zburstein:
This fixes #786 by creating a config variable called `max_compressed_image_size` that is used to validate whether the image being added is too large. The feature uses compressed size because the system does not determine the full size of the image until it is analyzed and the best way to quickly determine the size is by summing together the size of the layers on the manifest, which represents the compressed size of each layer. Spoke with @zhill about this and decided that compressed will be a good starting point for this feature.

Adds the validation to the following workflows:

- addition of image for analysis through the endpoint in the api service
- addition of image for analysis via the repo watcher handler
- addition of image for analysis via the image watcher handler

In order to test it required the addition of the new variable to the default config


anchore_engine/services/catalog/__init__.py
```diff
@@ -768,6 +768,14 @@ def handle_repo_watcher(*args, **kwargs):
                                 )
                                 raise
 
+                            if not catalog_impl.is_image_valid_size(new_image_info):
+                                localconfig = (
+                                    anchore_engine.configuration.localconfig.get_config()
+                                )
+                                raise Exception(
+                                    f"Image size of {new_image_info['compressed_size']} exceeds configured maximum of {localconfig.get('max_compressed_image_size')}"
```
dakaneye, line 776: minor: We're using several different string interpolation methods across the code base, but I haven't seen many F strings...might want to use format or % for consistency?

anchore_engine/common/images.py
```diff
@@ -75,6 +75,10 @@ def get_image_info(
             image_info["manifest"] = manifest
             image_info["parentmanifest"] = parentmanifest
             image_info["parentdigest"] = parentdigest
+            if manifest and manifest.get("layers"):
+                image_info["compressed_size"] = sum(
```
dakaneye, line 79: +1 this is nice and clean

anchore_engine/services/catalog/__init__.py
```diff
@@ -974,6 +982,12 @@ def handle_image_watcher(*args, **kwargs):
                     )
                     raise
 
+                if not catalog_impl.is_image_valid_size(image_info):
+                    localconfig = anchore_engine.configuration.localconfig.get_config()
+                    raise Exception(
```
dakaneye, line 987: consider raising an AnchoreApiError here, might be a BadRequest?

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2604,3 +2621,18 @@ def upsert_eval(dbsession, userId, record):
 
 
 ################################################################################
+
+# return true or false if image is a valid size based upon max_compressed_image_size specified in config
+def is_image_valid_size(image_info):
+    localconfig = anchore_engine.configuration.localconfig.get_config()
+    max_compressed_image_size = localconfig.get("max_compressed_image_size")
```
dakaneye, line 2628: Would we ever want to add a maximum possible configured value here? IE. max_compressed_image_size can't go over 10GB or something? and it can't be negative?

If the configured value is out of expected bounds perhaps we should do some normalizing. @zhill wdyt?

tests/functional/services/api/images/test_post.py
```diff
@@ -0,0 +1,24 @@
+class TestOversizedImageReturns400:
+    # Expectation for this test is that the image with tag is greater than the value defined in config
+    def test_oversized_image_post(self, make_image_analysis_request):
+        resp = make_image_analysis_request("anchore/test_images:oversized_image")
```
dakaneye, line 4: +1 nice dude!

dakaneye (approved):
LGTM, just some minor comments. Up to you to address them!

anchore_engine/services/catalog/__init__.py
zburstein: Yeah I was actually curious about this one. I initially raised an api exception, but then changed it because this exception is getting raised within a loop of all the detected tags. Its set up in such a way that all errors in this block will be caught and the process will then continue on to the next tag so this exception never gets exposed through the API.

anchore_engine/services/catalog/__init__.py
zburstein: switched to concatenation to match the rest of the file

anchore_engine/services/catalog/catalog_impl.py
zhill: If the configured value is out of acceptable bounds then the system should fail bootstrap with an error indicating invalid configuration found. We could set a max allowed configurable size, but I don't see that as an error case to handle it's just effectively a no-op. A value < 0 should be considered a "disabled" state equivalent to the config value being omitted entirely.

zburstein: @zhill I added a validation that prevents the service from starting if the value is not an integer. The way the code is structured, a negative value provided should always cause the image to pass this validation because the size provided by the manifest should never be less than 0, but I went ahead and added an explicit check on the condition to verify that the config value is "enabled"

anchore_engine/services/catalog/__init__.py
zhill: Generally we want to keep API-specific errors out of code that isn't in the API tier. If invoked via an API then an internal Exception can be mapped to an API exception if that error should be exposed to users.

anchore_engine/configuration/localconfig.py
```diff
@@ -55,6 +55,7 @@
         },
     },
     "policy_bundles_dir": "bundles/",
+    "max_compressed_image_size": None,
```
zhill, line 58: nit: this should default to 0 so the type is always int

zhill (approved):
small non-blocking (nit), but otherwise looks good.

PR #798, opened 2020-12-17
only validate rule if appropariate

dakaneye:

**What this PR does / why we need it**:  During testing in preparation for the 0.9.0 engine release found a bug that makes it impossible to add a rule after a (system_global == true && max_images_per_account != null) rule already exists

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #: (private repo issue)

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/api/controllers/archives.py
+++ b/anchore_engine/services/catalog/api/controllers/archives.py
@@ -177,16 +177,20 @@ def create_analysis_archive_rule(rule):
     try:
         with session_scope() as session:
             # Validate that only one system_global rule has max_images_per_account set
-            qry = session.query(ArchiveTransitionRule).filter(
-                ArchiveTransitionRule.account == ApiRequestContextProxy.namespace(),
-                ArchiveTransitionRule.system_global == True,
-                ArchiveTransitionRule.max_images_per_account != None,
-            )
-            if qry.first() is not None:
-                raise BadRequest(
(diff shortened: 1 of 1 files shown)


anchore_engine/services/catalog/api/controllers/archives.py
```diff
@@ -177,16 +177,20 @@ def create_analysis_archive_rule(rule):
...
-                raise BadRequest(
-                    "A system_global Archive Transition Rule already exists with max_images_per_account set",
-                    {"existingRule": repr(qry.first())},
+            if (
+                rule.get("system_global", False)
+                and rule.get("max_images_per_account", None) is not None
+            ):
+                qry = session.query(ArchiveTransitionRule).filter(
+                    ArchiveTransitionRule.account == ApiRequestContextProxy.namespace(),
+                    ArchiveTransitionRule.system_global.is_(True),
```
zhill, line 186: Why the change from == True to is_(True) here?  The null check on the following line I understand, but is there a postgres specific reason for this?

zhill (approved):
one nit (not blocker) question but looks good.

anchore_engine/services/catalog/api/controllers/archives.py
dakaneye: @zhill i was getting warnings from my IDE saying that technically the previous check was "truthy", which I understood to mean that it was synonymous with the "IS NOT NULL" check

PR #801, opened 2020-12-18
Make parent_manifest field not required in queue message

zhill:
Images that don't have a parent manifest aren't analyzing in the 0.9.0-dev branch properly, due to failure to parse the queue message on the analyzer. This is caused by an over-restrictive schema that requires parent_manifest to be present.

This fix relaxes that and adds unit tests for the parsing.


nightfurys (approved):
lgtm

Issue #787, opened 2020-12-16
Policy validation for expiration value in whitelist rule returns an inaccurate error message


holt:
If an invalid (non RFC 3339) string is submitted, the error message returns:

`could not convert input created_at value ...`

This should read:

`could not convert input expires_on value ...`

PR #802, opened 2020-12-18
fix error message, add field_name parameter to improve error message usefulness

dakaneye:

**What this PR does / why we need it**: Resolves an issue with error messages around failure to convert a date time string into a date time object

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: Fixes https://github.com/anchore/anchore-engine/issues/787

**Special notes**:

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -3274,6 +3274,7 @@ definitions:
       expires_on:
         type: string
         format: "date-time"
+        pattern: ^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:\:\d+)?[A-Z]?$
   PolicyRule:
     type: object
     description: A rule that defines and decision value if the match is found true for a given image.
--- a/anchore_engine/services/policy_engine/api/models.py
+++ b/anchore_engine/services/policy_engine/api/models.py
@@ -1,6 +1,5 @@
 from anchore_engine.common.schemas import Schema, JsonSerializable, RFC3339DateTime
 from marshmallow import fields, post_load
-from anchore_engine.utils import rfc3339str_to_datetime, datetime_to_rfc3339
 
 
 class DistroMapping(JsonSerializable):
--- a/anchore_engine/utils.py
+++ b/anchore_engine/utils.py
@@ -385,7 +385,7 @@ def rfc3339str_to_datetime(rfc3339_str):
 
     if ret is None:
         raise Exception(
-            "could not convert input created_at value ({}) into datetime using formats in {}".format(
+            "could not convert input value ({}) into datetime using formats in {}".format(
                 rfc3339_str, rfc3339_date_input_fmts
             )
         )


dspalmer99 (approved):
LGTM, pending rebase on master.

zhill:
Presumably the "date-time" spec in the swagger spec for the API isn't RFC3339 specific, but we can replace that with a specific regex in the API spec for that field and it will handle the validation at the API layer just like other types of checks and it should return a decent error message in that case. Would that approach suffice or was it tried and did not sufficiently indicate the problem?

dakaneye:
> Presumably the "date-time" spec in the swagger spec for the API isn't RFC3339 specific, but we can replace that with a specific regex in the API spec for that field and it will handle the validation at the API layer just like other types of checks and it should return a decent error message in that case. Would that approach suffice or was it tried and did not sufficiently indicate the problem?

@zhill I didn't consider adding it to the API spec, that's an interesting idea. I pulled this thread and started building/searching for a pre-existing regex for RFC3339 date time strings and there's a few long and complex regexes we could use:

```
"^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\.\\d+)?[A-Z]?(?:[+.-](?:08:\\d{2}|\\d{2}[A-Z]))?$"
```

Presumably, we'd want to match each of the 3 formats we support internally:
```python
rfc3339_date_input_fmts = [
    "%Y-%m-%dT%H:%M:%SZ",
    "%Y-%m-%dT%H:%M:%S.%fZ",
    "%Y-%m-%dT%H:%M:%S:%fZ",
]
```

Should I go ahead and pursue this path? I'm sure it would have value later on for date-time fields.

zhill:
> > Presumably the "date-time" spec in the swagger spec for the API isn't RFC3339 specific, but we can replace that with a specific regex in the API spec for that field and it will handle the validation at the API layer just like other types of checks and it should return a decent error message in that case. Would that approach suffice or was it tried and did not sufficiently indicate the problem?
> 
> @zhill I didn't consider adding it to the API spec, that's an interesting idea. I pulled this thread and started building/searching for a pre-existing regex for RFC3339 date time strings and there's a few long and complex regexes we could use:
> 
> ```
> "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\.\\d+)?[A-Z]?(?:[+.-](?:08:\\d{2}|\\d{2}[A-Z]))?$"
> ```
> 
> Presumably, we'd want to match each of the 3 formats we support internally:
> 
> ```python
> rfc3339_date_input_fmts = [
>     "%Y-%m-%dT%H:%M:%SZ",
>     "%Y-%m-%dT%H:%M:%S.%fZ",
>     "%Y-%m-%dT%H:%M:%S:%fZ",
> ]
> ```
> 
> Should I go ahead and pursue this path? I'm sure it would have value later on for date-time fields.

Yes, let's give that a try since it has less code impact as well as having consistent error responses for input format validation.

dakaneye:
@zhill seems to work! I tested the regex against the 3 formats and it works for those so we should be good!

zhill (approved):
lgtm. I'd like to review making this consistent across the API using a common type definition here (e.g. "rfc3339"), but works for this fix.

PR #811, opened 2020-12-19
Migrate binary analyzer to be in-process

wagoodman:
- Migrates the binary analyzer module to be within the analyzer process
- Adds a functional test to verify the results of the binary analyzer module
- Organizes all analyzers to be run with a single entrypoint under `anchore_engine.analyzers.run()`
- Modifies the busybox evidence collection to not use previous distro information but instead discover the version from the binary.


anchore_engine/analyzers/syft/__init__.py
```diff
@@ -2,22 +2,22 @@
...
     """
     Catalog the given image with syft, keeping only select artifacts in the returned results.
     """
-    all_results = run_syft(image)
+    all_results = run_syft(imagedir)
     return convert_syft_to_engine(all_results, unpackdir)
 
 
-def convert_syft_to_engine(all_results, unpackdir, handle_hints=True):
+def convert_syft_to_engine(all_results, unpackdir):
```
zhill, line 20: This change breaks the image import path that uses the handle_hints=False case. Need to update analyzer/imports.py's invocation and make the content_hints() function able to execute outside of an actual analysis directory structure, or restore the flag to bypass it.

anchore_engine/analyzers/syft/__init__.py
```diff
@@ -43,15 +43,17 @@ def convert_syft_to_engine(all_results, unpackdir, handle_hints=True):
...
-        for engine_entry in content_hints(unpackdir=unpackdir):
-            pkg_type = engine_entry.get("type")
-            if pkg_type:
-                handler = handlers_by_engine_type[pkg_type]
-                handler.save_entry(findings, engine_entry)
+    # apply content hints, overriding values that are there
+    # note: upstream of this processing overwrites the hints file location if
+    # the config does not explicitly enable hints processing, effectively disabling
+    # hints processing.
+    for engine_entry in content_hints(unpackdir=unpackdir):
```
zhill, line 53: Alternatively, to allowing a hints bypass, we could decouple the hints processing from the syft data merge, hosting this capability into the catalog_image() call since this is a post-processing on the merged data itself.

anchore_engine/analyzers/utils.py
```diff
@@ -692,7 +692,7 @@ def _hints_to_binary(pkg):
     pkg_license = anchore_engine.utils.ensure_str(pkg.get("license", ""))
     pkg_origin = anchore_engine.utils.ensure_str(pkg.get("origin", ""))
     pkg_files = pkg.get("files", [])
-    pkg_metadata = json.dumps(pkg.get("metadata", {}))
+    pkg_metadata = pkg.get("metadata", {})
```
zhill, line 695: This change makes sense in isolation, but does it change the API results? It looks like all the other conversions use the string version.

zhill (changes requested):
The approach makes sense, but a few issues arise in the refactor that need to be resolved. Tests are good as well.

anchore_engine/analyzers/syft/__init__.py
wagoodman: Great catch, I didn't see it used in the analysis pass and believed it to be dead code. I'll put that back.

anchore_engine/analyzers/syft/__init__.py
wagoodman: Agreed

anchore_engine/analyzers/syft/__init__.py
wagoodman: re: https://github.com/anchore/anchore-engine/pull/811#discussion_r546637218 , I'll change `convert_syft_to_engine` to not contain the hints processing

anchore_engine/analyzers/utils.py
wagoodman: `metadata` is not included in the content API:
```
$  anchore-cli --json image content docker.io/anchore/test_images:bin binary

        {
            "cpes": [],
            "license": "N/A",
            "licenses": [
                "N/A"
            ],
            "location": "/staged/positive/VERSION",
            "origin": "N/A",
            "package": "go",
            "type": "BINARY",
            "version": "1.15"
        },
        {
            "cpes": [],
            "license": "N/A",
            "licenses": [
                "N/A"
            ],
            "location": "/staged/positive/busybox",
            "origin": "N/A",
            "package": "busybox",
            "type": "BINARY",
            "version": "3.33.3"
        },
        {
            "cpes": [],
            "license": "N/A",
            "licenses": [
                "N/A"
            ],
            "location": "/staged/copy/busybox",
            "origin": "N/A",
            "package": "busybox",
            "type": "BINARY",
            "version": "1.31.1"
        }
...
```

I also didn't see this being referenced in the loaders or other spots. I did see the golang `_hints_to_go` does the same thing (which I can also remove the `json.dumps` I think), but I can't seem to find where this get's ingested.

anchore_engine/analyzers/syft/__init__.py
wagoodman: migrated hints processing to not be within the syft mappers

anchore_engine/analyzers/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+                            repr(serr.strip()),
+                        )
+                    )
+                else:
+                    logger.debug(
+                        "command succeeded: cmd=%s stdout=%s stderr=%s".format(
+                            repr(cmdstr), repr(sout.strip()), repr(serr.strip())
+                        )
+                    )
+            except Exception as err:
```
alfredodeza, line 64: there is no need to capture `err` here

anchore_engine/analyzers/binary.py
```diff
@@ -1,18 +1,133 @@
...
+                logger.exception(
+                    "Unexpected exception evaluating file='{}' for python busybox evidence".format(
+                        member.name
+                    )
+                )
+
+    # write evidence back to a results data structure
+    resultlist = {}
+    version_found_map = {}
+    for runtime in evidence.keys():
```
alfredodeza, line 88: there is no need to do `.keys()` here, it is the default when looping over dictionaries, same suggestion on the below loops

alfredodeza: Actually, I think I am missing the fact that a lot of this is copy/pasta, feel free to ignore :)

alfredodeza (approved):
This looks OK to me

anchore_engine/analyzers/__init__.py
luhring: Just following along and learning Python 😄 — would we not want to include the error in the log message?

anchore_engine/analyzers/binary.py
```diff
@@ -1,18 +1,133 @@
...
+
+binary_package_el = {
+    "name": None,
+    "version": None,
+    "location": None,
+    "type": "binary",
+    "files": [],
+    "license": "N/A",
+    "origin": "N/A",
+    "metadata": {},
```
zhill, line 20: Should this be an empty string?

zhill (commented):
Small suggestion, otherwise looks good.

anchore_engine/analyzers/__init__.py
alfredodeza: @luhring the `logger.exception` takes care of that for you

PR #813, opened 2020-12-22
Rpm handler fixes

Toure:

This fixes an issue that prevented one of the rpm handler functions from being called.


Toure:
moving this against v0.9.0-dev

PR #814, opened 2020-12-22
corrected: uncommented a function within the rpm handler and correcte…

### LINKED ISSUE #812: Image analysis shows "python" as binary package
This behavior was not present in v0.8.2.

(This is also causing more vulnerabilities to be shown than should be.)

Here's the diff for binary content output between `v0.8.2` and `v0.9.0-rc2` for `amazonlinux:2`:

Screen Shot 2020-12-22 at 10 03 26 AM

**Developer notes:**
- Look into the RPM syft mapper... we should be removing binary findings when a given file is owned by a package manager like RPM.

Toure:
 Uncommented a function within the rpm handler and corrected the functional test.

**Which issue this PR fixes**
fixes #812

### DIFF
--- a/tests/functional/clients/standalone/package_list/test_centos.py
+++ b/tests/functional/clients/standalone/package_list/test_centos.py
@@ -4,8 +4,6 @@ import pytest
 
 
 class TestRpm:
-    @pytest.mark.skip(reason="Waiting on completed functionality.")
-    @pytest.mark.parametrize("path", path_params(centos.pkgfiles_all))
     def test_pkgfiles_all(self, analyzed_data, path):
         result = analyzed_data("rpm")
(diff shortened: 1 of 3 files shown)


tests/functional/clients/standalone/package_list/test_centos.py
```diff
@@ -4,8 +4,6 @@
 
 
 class TestRpm:
-    @pytest.mark.skip(reason="Waiting on completed functionality.")
-    @pytest.mark.parametrize("path", path_params(centos.pkgfiles_all))
```
alfredodeza, line 8: this should not get removed

Toure: Thanks, stuck key.

alfredodeza (approved):

PR #816, opened 2020-12-23
Adds image import resource cleanup during account deletion

zhill:
Also fixes issue in account resource cleanup where failure in delete of
a single resource could cause the system to think the cleanup was
complete and remove the account before all resources were removed.

### DIFF
--- a/anchore_engine/services/catalog/__init__.py
+++ b/anchore_engine/services/catalog/__init__.py
@@ -32,6 +32,8 @@ from anchore_engine.db import (
     db_subscriptions,
     db_anchore,
     db_services,
+    db_archivemetadata,
+    db_policyeval,
     AccountStates,
     AccountTypes,
 )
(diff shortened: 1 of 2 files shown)


anchore_engine/services/catalog/__init__.py
```diff
@@ -2512,6 +2524,49 @@ def garbage_collect_imports():
...
+def list_account_import_ops(account: str, session=None, limit: int = 1000):
+    """
+    Return list of import operations for the given account, but only the first limit count to keep memory use reasonable.
+
+    :param account:
+    :param session:
+    :param limit:
+    :return: list of ImageImportOperations objects
+    """
+
```
nightfurys, line 2536: May be add a check for `session` before using

nightfurys (approved):
lgtm overall

anchore_engine/services/catalog/__init__.py
zhill: Yep, good catch.

PR #829, opened 2021-01-06
Fixes broken webhook notification delivery by catalog

zhill:
Fixes issue with notification webhooks not delivered by async process in catalog due to code expecting an API context when none is available. Removed the code (new to this release) that required the API context.

This only affected 0.9.0 dev code, not any previously released versions.


dspalmer99 (approved):
LGTM

PR #839, opened 2021-01-12
Move analyzer loaders out of package root into manager.py

zhill:
Updates the way that analyzer modules are loaded to be more flexible and explicit on search paths


anchore_engine/configuration/localconfig.py
```diff
@@ -101,6 +101,17 @@
...
+ANALYZER_SEARCH_PATHS = ["anchore_engine.analyzers"]
+
+
+def register_analyzers(module_path):
+    global ANALYZER_SEARCH_PATHS
+    ANALYZER_SEARCH_PATHS.append(module_path)
+
+
+def analyzer_paths():
+    return ANALYZER_SEARCH_PATHS
```
alfredodeza, line 113: I'm not following why this needs to be dynamic via functions that need to use `global`.

anchore_engine/analyzers/manager.py
```diff
@@ -0,0 +1,120 @@
+import os
```
alfredodeza, line 1: Is this just a wholesale move? or where there more changes?

alfredodeza (commented):
Looks OK to me except for the use of dynamic variables that require `global`

anchore_engine/configuration/localconfig.py
zhill: so that it can be modified if needed during bootstrap of the service based on external config. the analyzer service uses it but so does the `anchore-manager analyzers list` command that doesn't have a service bootstrap

anchore_engine/analyzers/manager.py
zhill: just a move, it's not strictly necessary for the fix since I decided to use the explicit config rather than just known package paths.

PR #841, opened 2021-01-13
Use MB for max image size config instead of bytes

zburstein:
I originally built the configuration value to be stored in bytes but MB makes more sense from a user perspective. Logging still shows values in bytes for granularity


Dockerfile
```diff
@@ -128,7 +128,7 @@ ENV ANCHORE_CONFIG_DIR=/config \
     ANCHORE_OAUTH_TOKEN_EXPIRATION=3600 \
     ANCHORE_AUTH_ENABLE_HASHED_PASSWORDS=false \
     AUTHLIB_INSECURE_TRANSPORT=true \
-    ANCHORE_MAX_COMPRESSED_IMAGE_SIZE=null
+    ANCHORE_MAX_COMPRESSED_IMAGE_SIZE=-1
```
zhill, line 131: Please add the unit here since it's not bytes, ANCHORE_MAX_COMPRESSED_IMAGE_SIZE_MB, will work. Just to be clear on units in the name.

anchore_engine/configuration/localconfig.py
```diff
@@ -55,7 +55,7 @@
         },
     },
     "policy_bundles_dir": "bundles/",
-    "max_compressed_image_size": None,
+    "max_compressed_image_size": -1,
```
zhill, line 58: Also, here if we can add unit since it isn't bytes: max_compressed_image_size_mb

zhill (changes requested):
Looks good just please add the _mb or _MB suffix to the configs so its clear to users that they should specify megabytes instead of bytes in this value.

dakaneye (approved):
LGTM

zhill (approved):
lgtm

PR #843, opened 2021-01-14
Provide logic which return a CVE id instead of VulnDB ID for a single reference.

Toure:

This change will determine the size of the NVD data structure and if
it is of size one element, the cve_id will be reported instead of the
vulndb_id.

**Which issue this PR fixes** 
fixes #: 442
https://github.com/anchore/enterprise/issues/442

**Special notes**:

### DIFF
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -649,6 +649,10 @@ class NvdMetadata(Base):
         )
 
     @property
+    def normalized_id(self):
+        return self.name
+
+    @property
     def description(self):
         return self.summary if self.summary else ""
 
@@ -802,6 +806,10 @@ class NvdV2Metadata(Base):
             self.__class__, self.name, self.created_at
         )
 
+    @property
+    def normalized_id(self):
+        return self.name
+
     def _get_score(self, metric, score_key):
         if metric:
             score = metric.get(base_metrics_key).get(score_key, -1.0)
@@ -942,6 +950,21 @@ class VulnDBMetadata(Base):
         return "<{} name={}, created_at={}>".format(
             self.__class__, self.name, self.created_at
         )
+
+    @property
+    def normalized_id(self):
+        """normalized_id will inspect the in coming external
+        references and return a cve id in the case of a single
+        match against vulndb information.
+        """
+        res = _get_one_or_none("source", "CVE ID", self.references)
+        if res and res.get("url"):
+            # findall should return a single id list ['2020-11989']
+            cve_id_col = re.findall(r"\=(\d+\-\d+)", res.get("url"))
+            if cve_id_col:
+                return "CVE-" + cve_id_col[0]
+
+        return self.name
 
     def _get_max_cvss_v3_metric_nvd(self):
         cvss_v3 = None
@@ -3033,3 +3056,33 @@ def select_nvd_classes(db=None):
 
     log.debug("selected {}/{} nvd classes".format(_nvd_cls, _cpe_cls))
     return _nvd_cls, _cpe_cls
+
+
+def _get_one_or_none(key, val, collection):
+    """
+    Find a match for the object in the collection using the key-value pair,
+    return the result only if 1 match is found.
+
+    Example instance of collection
+    [
+        {
+          "source": "CVE ID",
+          "url": "http://cve.mitre.org/cgi-bin/cvename.cgi?name=2000-0089"
+        },
+        {
+          "source": "Bugtraq ID",
+          "url": "http://www.securityfocus.com/bid/947"
+        },
+    ]
+    """
+    if not key or not val:
+        return None
+    result = None
+
+    for entry in collection:
+        if entry.get(key) == val:
+            if result:
+                return None
+            else:
+                result = entry
+    return result
--- a/anchore_engine/services/apiext/api/controllers/images.py
+++ b/anchore_engine/services/apiext/api/controllers/images.py
@@ -147,7 +147,6 @@ def make_response_vulnerability(vulnerability_type, vulnerability_data):
         "package_path": "Package_Path",
         "package_version": "Package_Version",
     }
-    id_cves_map = {}
     scan_result = vulnerability_data["legacy_report"]
     try:
         for imageId in list(scan_result.keys()):
@@ -181,8 +180,6 @@ def make_response_vulnerability(vulnerability_type, vulnerability_data):
                     el["vendor_data"] = make_cvss_scores(
                         all_data.get("vendor_data", [])
                     )
-                    for nvd_el in el["nvd_data"]:
-                        id_cves_map[nvd_el.get("id")] = el.get("vuln")
 
     except Exception as err:
         logger.exception("could not prepare query response")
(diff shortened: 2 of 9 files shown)


anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
```diff
@@ -998,8 +997,21 @@ def get_image_vulnerabilities(user_id, image_id, force_refresh=False, vendor_onl
...
 
+                    nvd_data = vulnerability_cpe.parent.get_cvss_data_nvd()
+                    vulndb_id = vulnerability_cpe.vulnerability_id
+
+                    # Check to see if a single CVE id is located in the nvd data
+                    # and present it to allow for easy cross reference from 
+                    # other sources and use public links for vulnerability information.
+                    # issue #442
+
+                    if vulndb_id and len(nvd_data) == 1:
```
nightfurys, line 1008: This condition is a little broad and and results true for nvd vulnerabilities too which is not necessary. I think a check like this, if necessary, should be over the type `isInstance(vulnerability_cpe, VulnDBCpe)` or the feed `vulnerability_cpe.feed == "vulndb"`

nightfurys (changes requested):
These changes would only impact the vulnerability listing API output. Policy evaluation uses vulnerabilities as an input. The underlying set of vulnerabilities on which policy engine operates should be the same but in this case the policy checks continue to receive non-nvd IDs even when an nvd translation is available. This might be the reason why @zhill may have suggested moving this upstream

anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
Toure: Ok, thank you for the feedback.

anchore_engine/db/entities/policy_engine.py
```diff
@@ -943,6 +954,19 @@ def __repr__(self):
...
+        """normalized_id will inspect the in coming external
+        references and return a cve id in the case of a single
+        match against vulndb information.
+        """
+        res = _search_dict("source", "CVE ID", self.references)
+        if len(res) == 2:
+            name = "CVE-" + res["url"].split("name=")[1]
+        else:
+            name = self.name
+        return name
```
nightfurys, line 968: This just returns the first instance of `CVE ID` object in references. Additional logic is required to compute that there is only one CVE reference before returning it as the normalized ID. Change this to something like 

```suggestion
        res = _get_one_or_none("source", "CVE ID", self.references)
        if res and res.get("url"):
            # regex match might be better, going with split for now. also making assumptions about link structure, need to verify
            parts = res.get("url").split("name=", 1)
            if parts and len(parts) > 1:
                return "CVE-" + parts[1]

        return self.name
```

anchore_engine/db/entities/policy_engine.py
```diff
@@ -3033,3 +3058,9 @@ def select_nvd_classes(db=None):
 
     log.debug("selected {}/{} nvd classes".format(_nvd_cls, _cpe_cls))
     return _nvd_cls, _cpe_cls
+
+
+def _search_dict(key, val, collection):
+    for entry in collection:
+        if entry[key] == val:
+            return entry
```
nightfurys, line 3066:
```suggestion
def _get_one_or_none(key, val, collection):
    """
    Find a match for the object in the collection using the key-value pair. Return the result only if 1 match is found

    Example instance of collection
    [
        {
          "source": "CVE ID",
          "url": "http://cve.mitre.org/cgi-bin/cvename.cgi?name=2000-0089"
        },
        {
          "source": "Bugtraq ID",
          "url": "http://www.securityfocus.com/bid/947"
        },
    ]
    """
    if not key or not val:
        return None

    result = None
    for entry in collection:
        if entry.get(key) == val:  # use get operator here instead of [] to avoid KeyError
            if result:
                return None  # > 1 result found, bail out of here and return None
            else:
                result = entry

    return result
```

anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
```diff
@@ -995,11 +994,11 @@ def get_image_vulnerabilities(user_id, image_id, force_refresh=False, vendor_onl
                     link = vulnerability_cpe.parent.link
                     if not link:
                         link = "{}/query/vulnerabilities?id={}".format(
-                            api_endpoint, vulnerability_cpe.vulnerability_id
+                            api_endpoint, vulnerability_cpe.parent.normalized_id
```
nightfurys, line 997: Rollback this change. This is related to the vulnerability query api which will be based off the regular ID, not the normalized ID

nightfurys (changes requested):

anchore_engine/db/entities/policy_engine.py
zhill: alternatively:
```
filtered = [ x for x in collection where x.get(key) == val]
if len(filtered) != 1:
  return None
else:
 return filtered[0]
```

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -447,7 +447,8 @@ def evaluate(self, image_obj, context):
                                     parameter_data[
                                         "link"
                                     ] = "{}/query/vulnerabilities?id={}".format(
-                                        api_endpoint, vulnerability_cpe.vulnerability_id
+                                        api_endpoint,
+                                        vulnerability_cpe.parent.normalized_id,
```
nightfurys, line 451: Rollback please. Refer to https://github.com/anchore/anchore-engine/pull/843/files#r558711037

anchore_engine/db/entities/policy_engine.py
Toure: Thanks for the pointers.

anchore_engine/db/entities/policy_engine.py
```diff
@@ -802,7 +808,12 @@ def __repr__(self):
             self.__class__, self.name, self.created_at
         )
 
-    def _get_score(self, metric, score_key):
+    @property
+    def normalized_id(self):
+        return self.name
+
+    @staticmethod
```
nightfurys, line 815: Why was this changed? Looks like an unrelated change, please roll this back

anchore_engine/db/entities/policy_engine.py
```diff
@@ -365,7 +364,7 @@ def get_nvd_identifiers(self, _nvd_cls, _cpe_cls):
                     if cve_id and cve_id not in cves:
                         cves.append(cve_id)
         except Exception as err:
-            log.warn(
+            log.warning(
```
nightfurys, line 367: log.warn() is used all over the place and needs to be updated. But it should be handled in a different commit as it's unrelated to this change

anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -126,14 +126,16 @@ class VulnerabilityMatchTrigger(BaseTrigger):
     fix_available = BooleanStringParameter(
         name="fix_available",
         example_str="true",
-        description="If present, the fix availability for the vulnerability record must match the value of this parameter.",
+        description="If present, the fix availability for the "
```
nightfurys, line 129: Unrelated change probably caused by pycharm default linter, rollback and use black to format to minimize the changes

nightfurys (changes requested):

nightfurys:
I would also strongly recommend unit tests for the newly introduced methods and may be integration tests in downstream repo to test this functionality

nightfurys:
Touré's rebase pulled commits that have already been merged, probably because the rebase was done against a stale branch without fetching the changes. Cleaned up the rebase and pushed the changes to the PR

nightfurys:
Touré's rebase effort rolled back use of normalized ID for blacklist evaluation. Reintroduced that change back into the PR

anchore_engine/services/apiext/api/controllers/images.py
```diff
@@ -204,7 +201,25 @@ def make_response_vulnerability(vulnerability_type, vulnerability_data):
         "feed_group": "feed_namespace",
     }
     scan_result = vulnerability_data["cpe_report"]
+    # gather a map of package path->CVE IDs referred by vulndb entries, dedup logic copied from policy engine
```
Toure, line 204: that fixed the output issue.

nightfurys (approved):
@zhill, the commits in PR have been finalized, can you give this a quick look

zhill (approved):
lgtm. one suggestion but it is not blocking.

anchore_engine/services/apiext/api/controllers/images.py
Toure: This issue arose once the new logic was introduced as it thought the cve data was presented twice but instead of presenting some information, it prevented the information from surfacing.

nightfurys: @zhill, the change was necessary because of the switch over to CVE IDs. Previous dedup logic collects nvd refs (CVEs) for each vulnerability and filters those from the final list. After the switch over to CVE identifier, the dedup logic filtered out the vulnerabilities that it shouldn't have. In addition, dedup was based purely on the identifier without any consideration for artifact. That was incorrect behaviour and the fix addresses that as well

PR #845, opened 2021-01-14
Require default admin password for initial bootstrap

### LINKED ISSUE #810: System should require explicit admin password set to bootstrap first time
Without an explicit admin password set the system should fail to bootstrap for the first time.

See localconfig.py for default admin password loading logic.

zburstein:
Fixes #810 by removing the default for `ANCHORE_ADMIN_PASSWORD` in dockerfile and `default_admin_password` in localconfig. Raises an exception if `default_admin_password` is not present for the first bootstrap


zhill (approved):
looks good!

PR #857, opened 2021-01-22
If an image has not been analyzed it, do not consider it a candidate …

dakaneye:
…for transition

**What this PR does / why we need it**: Addresses issue where stacktrace is printed when transition rules execute against images that are not done being analyzed

**Special notes**:

### DIFF
--- a/anchore_engine/db/db_catalog_image_docker.py
+++ b/anchore_engine/db/db_catalog_image_docker.py
@@ -242,7 +242,7 @@ def get_tag_histories(session, userId, registries=None, repositories=None, tags=
                 CatalogImageDocker.imageDigest == CatalogImage.imageDigest,
             ),
         )
-        .filter(CatalogImage.userId == userId, CatalogImage.analyzed_at.isnot(None))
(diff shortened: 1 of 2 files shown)


anchore_engine/db/db_catalog_image_docker.py
```diff
@@ -242,7 +242,7 @@ def get_tag_histories(session, userId, registries=None, repositories=None, tags=
                 CatalogImageDocker.imageDigest == CatalogImage.imageDigest,
             ),
         )
-        .filter(CatalogImage.userId == userId)
+        .filter(CatalogImage.userId == userId, CatalogImage.analyzed_at.isnot(None))
```
zhill, line 245: Looking downstream at this, the analysis_status is handled in the actual archiver task. So lets just handle this case in the timestamp comparison code instead of filtering at the db query.

zhill (approved):
this works. I think we can actually push that check into the _is_exclude_expired() call which is the first place it actually cares about the timestamp. This is more efficient but exposes internal logic from the function. But, that's not a correctness issue.

PR #862, opened 2021-01-26
Multiple policy bundle dirs

dspalmer99:
This PR updates localconfig to allow multiple folly-qualified dirs containing policy bundles to be passed in during bootstrap.

### DIFF
--- a/anchore_engine/configuration/localconfig.py
+++ b/anchore_engine/configuration/localconfig.py
@@ -120,6 +120,7 @@ def analyzer_paths():
 def register_policy_bundle_source_dir(source_dir):
     global POLICY_BUNDLE_SOURCE_DIRS
     POLICY_BUNDLE_SOURCE_DIRS.append(source_dir)
+    load_policy_bundle_paths()
 
 
 def policy_bundle_source_dirs():
--- a/anchore_engine/configuration/localconfig.py
+++ b/anchore_engine/configuration/localconfig.py
@@ -101,7 +101,11 @@ CRED_CACHE_TTL = int(os.getenv("ANCHORE_INTERNAL_CRED_CACHE_TTL", 600))
 CRED_CACHE_LOCK_WAIT_SEC = int(os.getenv("ANCHORE_INTERNAL_CRED_CACHE_WAIT_SEC", 3))
 
 ANALYZER_SEARCH_PATHS = ["anchore_engine.analyzers"]
-POLICY_BUNDLE_SOURCE_DIRS = ["conf/bundles/"]
+POLICY_BUNDLE_SOURCE_DIRS = [
+    os.path.join(
+        resource_filename("anchore_engine", "conf/bundles/")
+    )
+]
 
 
 def register_analyzers(module_path):
@@ -188,13 +192,7 @@ def load_policy_bundle_paths(src_dirs=None):
 
     # This value will typically == None, outside of automated tests
     if src_dirs == None:
-        src_dirs = []
-        for policy_bundles_source_dir in POLICY_BUNDLE_SOURCE_DIRS:
-            src_dirs.append(
-                os.path.join(
-                resource_filename("anchore_engine", policy_bundles_source_dir)
-                )
-            )
+        src_dirs = policy_bundle_source_dirs()
 
     try:
         if src_dirs is not None and len(src_dirs) > 0 and policy_bundles_dir is not None:
--- a/anchore_engine/configuration/localconfig.py
+++ b/anchore_engine/configuration/localconfig.py
@@ -101,6 +101,7 @@ CRED_CACHE_TTL = int(os.getenv("ANCHORE_INTERNAL_CRED_CACHE_TTL", 600))
 CRED_CACHE_LOCK_WAIT_SEC = int(os.getenv("ANCHORE_INTERNAL_CRED_CACHE_WAIT_SEC", 3))
 
 ANALYZER_SEARCH_PATHS = ["anchore_engine.analyzers"]
+POLICY_BUNDLE_SOURCE_DIRS = ["conf/bundles/"]
 
 
 def register_analyzers(module_path):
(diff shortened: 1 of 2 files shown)


anchore_engine/configuration/localconfig.py
```diff
@@ -112,6 +115,16 @@ def analyzer_paths():
     return ANALYZER_SEARCH_PATHS
 
 
+def register_policy_bundle_source_dir(source_dir):
+    global POLICY_BUNDLE_SOURCE_DIRS
+    POLICY_BUNDLE_SOURCE_DIRS.append(source_dir)
+    load_policy_bundle_paths()
```
dspalmer99, line 121: I'd like to refactor this to not have this extra call here, but as the config/bootstrapping currently exists this is necessary because in https://github.com/anchore/anchore-engine/blob/e4d21c6f8171027e03b974ed8a0d570dcc20e190/anchore_engine/twisted.py#L217 when the service starts up, the path that leads to `load_policy_bundle_paths()` below is called before the service is initialized, so if the service adds policy bundles to the config object via a handler it will be to late, even if with a pre_config handler.

This is maybe slightly hacky, but multiple calls to load_policy_bundle_paths() will have no effect on existing bundles in the case where subsequent calls are just adding more bundles.

anchore_engine/configuration/localconfig.py
```diff
@@ -168,43 +181,48 @@ def load_defaults(configdir=None):
...
+    if src_dirs == None:
+        src_dirs = policy_bundle_source_dirs()
+
     try:
-        if policy_bundles_dir is not None:
+        if (
+            src_dirs is not None
+            and len(src_dirs) > 0
+            and policy_bundles_dir is not None
+        ):
```
alfredodeza, line 201: in Python a `None` or an empty data structure like a list, will evaluate to `False` so this could be:
```suggestion
        if policy_bundles_dir is not None and src_dirs:
```

anchore_engine/configuration/localconfig.py
```diff
@@ -168,43 +181,48 @@ def load_defaults(configdir=None):
...
+                    try:
+                        file = os.path.join(policy_bundles_dir_full_path, file_name)
+                        policy_bundles.append(
+                            {
+                                "active": file_name == default_bundle_name,
+                                "bundle_path": file,
+                            }
+                        )
+                        copy_config_file(file, file_name, src_dir)
+                    except Exception as e:
```
alfredodeza, line 220: what exception are we catching here?

anchore_engine/configuration/localconfig.py
```diff
@@ -168,43 +181,48 @@ def load_defaults(configdir=None):
...
-                    )
-                    copy_config_file(file, file_name, src_dir)
-                except Exception as e:
-                    logger.warn(
-                        "Policy bundle {} not found, unable to load. Exception: {}".format(
-                            file_name, e
+            for src_dir in src_dirs:
+                for file_name in os.listdir(src_dir):
+                    try:
+                        file = os.path.join(policy_bundles_dir_full_path, file_name)
```
alfredodeza, line 212: this line will probably not bubble up an exception

anchore_engine/configuration/localconfig.py
```diff
@@ -168,43 +181,48 @@ def load_defaults(configdir=None):
...
+                        policy_bundles.append(
+                            {
+                                "active": file_name == default_bundle_name,
+                                "bundle_path": file,
+                            }
+                        )
+                        copy_config_file(file, file_name, src_dir)
+                    except Exception as e:
+                        logger.warn(
```
alfredodeza, line 221: using `logger.warn` eats the traceback. Are we trying to fully obscure the traceback?

tests/unit/anchore_engine/configuration/test_localconfig.py
```diff
@@ -73,30 +73,48 @@ def get_mock_config_with_policy_bundles(dir, bundle_filenames, simulate_exceptio
...
-    load_policy_bundle_paths(src_dir=input_dir.strpath)
+    load_policy_bundle_paths(src_dirs=src_dirs)
 
     # get and validate the relevant config bits
     config = get_config()
     assert config["policy_bundles"] is not None
-    assert len(config["policy_bundles"]) == len(config_filenames)
-    for config_filename in config_filenames:
+    assert len(config["policy_bundles"]) == len(config_filenames_flat)
+    for config_filename in config_filenames_flat:
```
alfredodeza, line 117: I know you aren't adding this change in the PR, but loops in tests are not good because they don't allow to see the failures in each item of the loop

anchore_engine/configuration/localconfig.py
dspalmer99: In that case, couldn't it also further be simplified to `if policy_bundles_dir and src_dirs: ` ?

anchore_engine/configuration/localconfig.py
dspalmer99: From the docs it doesn't look like it does. I'll simplify.

anchore_engine/configuration/localconfig.py
dspalmer99: Looking through the commit history, it looks like it was originally intended to catch an exception in the line you called out above.

tests/unit/anchore_engine/configuration/test_localconfig.py
dspalmer99: I agree that could be refactored but I'm not sure I want to pick that up as part of this PR. Also, when this failed on me (occasionally) during dev it tended to be all or nothing. Not saying that would always be the case though.

dspalmer99:
Thanks for the feedback @alfredodeza . I've made the changes you asked for, and added a little extra logging and another test as well.

anchore_engine/configuration/localconfig.py
alfredodeza: I wasn't sure what `policy_bundles_dir` were (was not clear if that was a list as well). Sometimes the logic may want to really ensure it isn't a `None` so that is valid. If you know that it doesn't need to check for a `None` just if it is empty, then your suggestion works!

anchore_engine/configuration/localconfig.py
alfredodeza: Right. I mean, it is catching _any_ exception that may show up. If this is existing code, feel free to ignore. For reference, this issue captures the problems with broad exceptions: https://github.com/anchore/anchore-engine/issues/867

dspalmer99: Yeah I had looked at 867 this morning. Totally agree that making that more targeted is a good thing to do, but this is existing code.

zhill (approved):
lgtm

zhill:
I'm going to squash-merge this since it's into the release branch and has 10 commits many of which are linting updates.

dspalmer99:
Ack, thanks @zhill

PR #863, opened 2021-01-26
Dont allow selector to be empty if not max_images_per_account rule, a…

### LINKED ISSUE #858: Behavior of max_images_per_account in analysis-archive rule
When creating an archive rule to set a global max_images_per_account like this:

```
HTTP POST /v1/archives/rules
{
  "system_global": true,
  "transition": "archive",
  "max_images_per_account": 2
}
```

the system applies defaults and creates a rule like this
```
{
    "analysis_age_days": -1,
    "created_at": "2021-01-23T22:49:06Z",
    "exclude": {
        "expiration_days": -1,
        "selector": {
            "registry": "",
            "repository": "",
            "tag": ""
        }
    },
    "last_updated": "2021-01-23T22:49:06Z",
    "max_images_per_account": 2,
    "rule_id": "fcb57410360b4f8ba3f4ecc3a3e4c6b1",
    "selector": {
        "registry": "*",
        "repository": "*",
        "tag": "*"
    },
    "system_global": true,
    "tag_versions_newer": -1,
    "transition": "archive"
}
```

The result is that this creates a rule with selectors that default to all and add every single image to the archive. Should this be the expected behavior or should there be changes?

dakaneye:
…nd force empty if it is

**What this PR does / why we need it**: This Change makes our logic around transition rules a bit better. Selector fields can no longer be empty strings unless its a global rule for max_images_per_account

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #858

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/api/controllers/archives.py
+++ b/anchore_engine/services/catalog/api/controllers/archives.py
@@ -177,10 +177,16 @@ def create_analysis_archive_rule(rule):
     try:
         with session_scope() as session:
             # Validate that only one system_global rule has max_images_per_account set
+            selector = rule.get("selector", {})
             if (
                 rule.get("system_global", False)
                 and rule.get("max_images_per_account", None) is not None
             ):
+                if selector or rule.get("exclude", {}):
+                    raise BadRequest(
+                        "Cannot set a selector or exclude on a rule with max_images_per_account",
+                        {},
+                    )
                 qry = session.query(ArchiveTransitionRule).filter(
                     ArchiveTransitionRule.account == ApiRequestContextProxy.namespace(),
                     ArchiveTransitionRule.system_global.is_(True),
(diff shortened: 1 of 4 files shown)


zburstein:
@dakaneye I see how this fixes the issues regarding validation of the users input for the selectors when creating an archive rule, but wouldn't this still create global selectors that match everything when creating a rule that sets `max_images_per_account`?

dakaneye:
> @dakaneye I see how this fixes the issues regarding validation of the users input for the selectors when creating an archive rule, but wouldn't this still create global selectors that match everything when creating a rule that sets `max_images_per_account`?

Ahh yeah I see that my logic is a bit off. Should we set a match_all exclude in this case then, @zhill ?

zburstein:
@dakaneye What if  [catalog/archiver.py:482](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/catalog/archiver.py#L482) - [catalog/archiver.py:519](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/catalog/archiver.py#L519)  was wrapped in a conditional that only runs if max_images_per_account is not set? That could then bypass the need to make any queries that wont be used and would ensure that max_images_per_account operates entirely independent of any other selectors

zburstein (approved):
lgtm

anchore_engine/services/catalog/api/controllers/archives.py
```diff
@@ -177,10 +177,16 @@ def create_analysis_archive_rule(rule):
     try:
         with session_scope() as session:
             # Validate that only one system_global rule has max_images_per_account set
+            selector = rule.get("selector", {})
```
zhill, line 180: good cleanup. this easier to read/understand as well

zhill (approved):
lgtm

PR #866, opened 2021-01-26
Break down image deletion into an extensible list of smaller functions

nightfurys:

### DIFF
--- a/anchore_engine/services/catalog/catalog_impl.py
+++ b/anchore_engine/services/catalog/catalog_impl.py
@@ -2180,13 +2180,10 @@ def _image_deletion_checks_and_prep(userId, image_record, dbsession, force=False
     return dodelete, msgdelete, image_ids, image_fulltags
 
 
-def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_fulltags):
+def _delete_image_artifacts(account_id, image_digest, image_ids, full_tags, db_session):
+    error = None
+    rcs = []
     obj_store = anchore_engine.subsys.object_store.manager.get_manager()
-    imageDigest = image_record["imageDigest"]
-
-    logger.debug("DELETEing image from catalog")
-    rc = db_catalog_image.delete(imageDigest, userId, session=dbsession)
-
     # digest-based archiveId buckets
     for bucket in [
         "analysis_data",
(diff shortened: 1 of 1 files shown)


anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2196,32 +2191,85 @@ def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_ful
         "manifest_data",
         "parent_manifest_data",
     ]:
-        archiveId = imageDigest
-        logger.debug("DELETEing image from archive {}/{}".format(bucket, archiveId))
-        rc = obj_store.delete(userId, bucket, archiveId)
+        try:
+            logger.debug("DELETEing image from archive %s/%s", bucket, image_digest)
+            rc = obj_store.delete(account_id, bucket, image_digest)
+        except Exception:
```
alfredodeza, line 2197: the usage of try/except here could be better. It is useful to capture only what is expected to be failing - even if it is being logged.

In all instances where this try/except/log pattern is used, it is actually not needed: the exception is still caught by the caller (and again, a broad exception). Ideally a granular exception - or no exception at all would be caught. It is up to the caller to decide how to deal with this.

Introducing these patterns makes it hard to debug (can't catch these exceptions anywhere) and impossible to test accurately, because again, all exceptions are caught

alfredodeza (changes requested):
At some point, it would be ideal to revisit the exception catching patterns in Anchore Engine as a whole. These are problematic because makes debugging and testing hard if not impossible. 

I understand that this PR addresses important issues for a release, but I've bubbled up these concerns before other releases and would like this one to get addressed. 

To follow up, and have some accountability on fixing the problem, I've created an issue to track the effort https://github.com/anchore/anchore-engine/issues/867

For this PR in particular, the changes needed to make this better are straightforward.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2196,32 +2191,85 @@ def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_ful
...
 
     # digest/tag-based archiveId buckets
     for bucket in ["vulnerability_scan"]:
-        for fulltag in image_fulltags:
-            archiveId = "{}/{}".format(imageDigest, fulltag)
-            logger.debug("DELETEing image from archive {}/{}".format(bucket, archiveId))
-            rc = obj_store.delete(userId, bucket, archiveId)
+        for full_tag in full_tags:
+            archive_id = "{}/{}".format(image_digest, full_tag)
+            try:
```
Toure, line 2206: The addition of a try statement is not needed as the called function within the obj_store object will raise an exception on failure.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2196,32 +2191,85 @@ def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_ful
...
+                    "Error deleting image from archive %s/%s", bucket, archive_id
+                )
 
-    logger.debug("DELETEing image from policy_engine")
 
+def _delete_image_catalog(account_id, image_digest, image_ids, full_tags, db_session):
+    try:
+        logger.debug("DELETEing image from catalog")
+        rc = db_catalog_image.delete(image_digest, account_id, session=db_session)
+    except Exception:
```
Toure, line 2219: Exception is too broad this should be ValueError or extending the exceptions.py module to include a new Type that is more specific to this kind of error.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2196,32 +2191,85 @@ def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_ful
...
+                "DELETING image from policy engine account_id=%s image_id=%s",
+                account_id,
+                img_id,
             )
-            rc = pe_client.delete_image(user_id=userId, image_id=img_id)
-    except:
-        logger.exception("Failed deleting image from policy engine")
-        raise
+            rc = pe_client.delete_image(user_id=account_id, image_id=img_id)
+    except Exception:
```
Toure, line 2237: Exception is too broad this should be ValueError or extending the exceptions.py module to include a new Type that is more specific to this kind of error.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2196,32 +2191,85 @@ def _delete_image_for_real(userId, image_record, dbsession, image_ids, image_ful
...
+        try:
+            logger.debug("Executing delete for image %s", desc)
+            delete_func(
+                account_id=userId,
+                image_digest=image_digest,
+                image_ids=image_ids,
+                full_tags=image_fulltags,
+                db_session=dbsession,
+            )
+        except Exception:
```
Toure, line 2271: Exception is too broad this should be ValueError or extending the exceptions.py module to include a new Type that is more specific to this kind of error.

Toure (changes requested):

nightfurys:
> At some point, it would be ideal to revisit the exception catching patterns in Anchore Engine as a whole. These are problematic because makes debugging and testing hard if not impossible.
> 
> I understand that this PR addresses important issues for a release, but I've bubbled up these concerns before other releases and would like this one to get addressed.
> 
> To follow up, and have some accountability on fixing the problem, I've created an issue to track the effort #867
> 
> For this PR in particular, the changes needed to make this better are straightforward.

I agree with improving the exception handling but it should be noted a bigger change is necessary to support that. The exception raisers are not consistent through out the system and its down to the invoking function to interpret the errors. Error handling across the board needs improvement to raise well-defined reliable exceptions, overhauling the try-except blocks to handle restrictive errors is not a complete solution

anchore_engine/services/catalog/catalog_impl.py
nightfurys: The called function - obj_store.delete(), does not transform errors into a specific type of exception, it just raises any errors caught. Without changing this function, it is impossible to handle a specific type of exception in the calling function. That's a bigger change well outside the scope of this ticket

nightfurys:
@zhill Merging this following our discussion. This blocks a downstream commit. I can follow it up with more updates if necessary

anchore_engine/services/catalog/catalog_impl.py
alfredodeza: I see what you mean for the handling of exceptions in this particular function - whatever was the last exception will get raised later. However, it is unclear what exception we are talking about here. It certainly can't be any possible exception?

Issue #850, opened 2021-01-16
Anchore Engine 0.9.0 unable to scan images built with buildah


found-it:
**Is this a request for help?**:

No

---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):

BUG REPORT

**Version of Anchore Engine and Anchore CLI if applicable**:

Engine version
```
Engine DB Version: 0.0.14
Engine Code Version: 0.9.0
```

CLI Version
```
anchore-cli, version 0.9.0
```

**What happened**:

Anchore fails to scan images built by `buildah bud`.  I deployed anchore engine using docker-compose and then built an image using buildah. I then pushed the image to my personal docker.io account and scanned it with anchore.

```
buildah bud -f Dockerfile -t jpetersenames/curly:buildah .
buildah push jpetersenames/curly:buildah
anchore-cli image add jpetersenames/curly:buildah
anchore-cli image wait jpetersenames/curly:buildah
```

Will return

```
Image Digest: sha256:38dc15e89cad8c4e1ef58509e0d166f09efd56f4f6e9d513931665539eaf515b
Parent Digest: sha256:38dc15e89cad8c4e1ef58509e0d166f09efd56f4f6e9d513931665539eaf515b
Analysis Status: analysis_failed
Image Type: docker
Analyzed At: None
Image ID: 50fc93eeea603572ca1f7b40bafb4db70dfe0117ea1da427064a1fd694abcb09
Dockerfile Mode: None
Distro: None
Distro Version: None
Size: None
Architecture: None
Layer Count: None

Full Tag: docker.io/jpetersenames/curly:buildah
Tag Detected At: 2021-01-15T23:09:04Z
```

**What did you expect to happen**:

I expected the container to analyze successfully

**Any relevant log output from /var/log/anchore**:

The analyzer log contains

```
[service:worker] 2021-01-15 23:10:06+0000 [-] [Thread-3663] [anchore_engine.utils/run_check()] [ERROR] {"from-lib":"stereoscope","level":"debug","msg":"image: source=OciDirectory location=/analysis_scratch/e8d53133-d0ba-4edd-a086-b73b6c003351/raw","time":"2021-01-15 23:10:06"}
[...]

PR #870, opened 2021-01-26
Ensure correct mediaType is set in manifest

alfredodeza:
Adds a helper function to ensure that each blob in an OCI directory will be of the correct type. If it isn't the helper will rewrite the manifest to ensure that the correct `mediaType` is set.

Fixes https://github.com/anchore/anchore-engine/issues/850

### DIFF
--- a/anchore_engine/clients/skopeo_wrapper.py
+++ b/anchore_engine/clients/skopeo_wrapper.py
@@ -1,3 +1,4 @@
+import gzip
 import json
 import os
 import re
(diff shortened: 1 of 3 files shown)


anchore_engine/clients/skopeo_wrapper.py
```diff
@@ -269,6 +271,49 @@ def get_digest_value(digest_with_algorithm_prefix: str):
...
+def is_gzip(path: str):
+    try:
+        with gzip.open(path) as _f:
+            _f.read(1)
+            return True
+    except Exception:
+        return False
+
+
+def ensure_tar_gzip_media_type(layer: dict):
```
luhring, line 307: I think we can ditch this function now 🗑️

alfredodeza: removed!

luhring (approved):
🏆

Toure (approved):

PR #880, opened 2021-01-27
Improve the message and description for vulnerability_data_unavailable and stale_feed_data triggers

zhill:
The existing messages are bit confusing since engine has multiple feeds used for image vuln scans. Those triggers are specific for distro packages, not all package types. This update makes that more clear.

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
@@ -892,7 +892,7 @@ class VulnerabilityMatchTrigger(BaseTrigger):
 
 class FeedOutOfDateTrigger(BaseTrigger):
     __trigger_name__ = "stale_feed_data"
-    __description__ = "Triggers if the CVE data is older than the window specified by the parameter MAXAGE (unit is number of days)."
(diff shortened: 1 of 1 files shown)


anchore_engine/services/policy_engine/engine/policy/gates/vulnerabilities.py
```diff
@@ -948,15 +948,13 @@ def evaluate(self, image_obj, context):
...
-    )
+    __description__ = "Triggers if vulnerability data is unavailable for the image's distro packages such as rpms or dpkg. Non-OS packages like npms and java are not considered in this evaluation"
 
     def evaluate(self, image_obj, context):
         if not have_vulnerabilities_for(DistroNamespace.for_obj(image_obj)):
             self._fire(
-                msg="Feed data unavailable, cannot perform CVE scan for distro: "
-                + str(image_obj.distro_namespace)
+                msg="Distro-specific feed data not found for distro namespace: %s. Cannot perform CVE scan OS/distro packages"
+                % image_obj.distro_namespace
```
Toure, line 957: Looks good to me, a small nit around string format, should we start to replace the old string % operator with the new style format method.

Toure (approved):

PR #884, opened 2021-01-29
Fix method name to match parent class

### LINKED ISSUE #883: Archiving fails when there is an analysis-archive delete transition
**Is this a request for help?**:
No
---

BUG REPORT

Engine Code Version: 0.9.0

**What happened**:
If an analysis-archive rule with delete transition exists the archive process fails.

That the archive and delete process works without aborting with an error.

```
return self._evaluate_tag_history(rule, tag_histories_qry)
AttributeError: 'ArchivedAnalysisDeleter' object has no attribute '_evaluate_tag_history'
```

**Anything else we need to know:**
The error occurs here:
https://github.com/anchore/anchore-engine/blob/c045794dde9d282a6393c60febb4852fefe20353/anchore_engine/services/catalog/archiver.py#L742
I think it should be:
```
return self._evaluate_tag_history_and_exclude(rule, tag_histories_qry)
```

dakaneye:

**What this PR does / why we need it**: Forgot to rename the method call used by the child class

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #883

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/archiver.py
+++ b/anchore_engine/services/catalog/archiver.py
@@ -739,7 +739,7 @@ class ArchivedAnalysisDeleter(ImageAnalysisArchiver):
                 ArchivedImage.created_at < min_time
             )
 
-        return self._evaluate_tag_history(rule, tag_histories_qry)
+        return self._evaluate_tag_history_and_exclude(rule, tag_histories_qry)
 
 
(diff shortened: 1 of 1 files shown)


zhill (approved):
lgtm

PR #895, opened 2021-02-02
Set the python package location according to the package key

dakaneye:


nightfurys (approved):
lgtm. not blocking - could probably benefit from a few unit tests

anchore_engine/common/helpers.py
```diff
@@ -276,6 +276,7 @@ def extract_python_content(image_data):
         ]["base"]
         for k in list(adata.keys()):
             ret[k] = safe_extract_json_value(adata[k])
+            ret[k]["location"] = k
```
zhill, line 279: this is correct, but can you add a comment here indicating why this is done so we have some future context?
```suggestion
            # Set the location to the module path itself. The Python "Location" field from analysis is the parent install directory so can be the same for many packages, this ensures the location is specific to that package
            ret[k]["location"] = k
```

zhill (changes requested):

zhill (approved):
test looks good. thanks

PR #896, opened 2021-02-02
Update the scanner config method in policy engine for providing overr…

zhill:
Updates the indirection mechanism so it can be re-used elsewhere without a reference to the API layer. Uses the mechanism in the policy evaluation for the vulnerabilities gate's context setup.

### DIFF
--- a/anchore_engine/services/policy_engine/__init__.py
+++ b/anchore_engine/services/policy_engine/__init__.py
@@ -33,9 +33,10 @@ feed_sync_queuename = "feed_sync_tasks"
 system_user_auth = None
 feed_sync_msg = {"task_type": "feed_sync", "enabled": True}
 
+# These are user-configurable but mostly for debugging and testing purposes
 try:
     FEED_SYNC_RETRIES = int(os.getenv("ANCHORE_FEED_SYNC_CHECK_RETRIES", 5))
-except:
+except ValueError:
     logger.exception(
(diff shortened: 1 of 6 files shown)


dakaneye (approved):
Should have the same effect as before but as a singleton pattern

anchore_engine/services/policy_engine/__init__.py
```diff
@@ -33,9 +33,10 @@
 system_user_auth = None
 feed_sync_msg = {"task_type": "feed_sync", "enabled": True}
 
+# These are user-configurable but mostly for debugging and testing purposes
 try:
     FEED_SYNC_RETRIES = int(os.getenv("ANCHORE_FEED_SYNC_CHECK_RETRIES", 5))
-except:
+except ValueError:
```
alfredodeza, line 39: vhpy

anchore_engine/services/policy_engine/engine/scanner.py
```diff
@@ -0,0 +1,37 @@
...
+        return image.vulnerabilities()
+
+    def get_cpe_vulnerabilities(self, image):
+        with timer("Image vulnerability cpe lookups", log_level="debug"):
+            return image.cpe_vulnerabilities(
+                _nvd_cls=self.nvd_cls, _cpe_cls=self.cpe_cls
+            )
+
+
+scanner_type = DefaultVulnScanner
```
alfredodeza, line 26: why is this a variable instead of an argument to `get_scanner()`?

PR #905, opened 2021-02-08
set to cryptography version 3.3.2 to avoid setuptools-rust downstream dependency

dakaneye:

**What this PR does / why we need it**: [Over the weekend, the `cryptography` package released an update](https://cryptography.io/en/latest/changelog.html) that required the `setuptools-rust` dev tool package to be installed, which isn't in our image (for this repo or enterprise), and it's causing the build to fail

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 Authlib==0.15.2
+cryptography==3.3.2
 boto3==1.11.6
 Click==7.0
 connexion==2.5.1


alfredodeza:
We should not try to pin dependencies of dependencies because it is not up to us to define what those constraints are. In this case, Authlib is defining what is acceptable, and we can't tell for sure if this version in particular is safe enough. 

I do understand this may fix the problem, although I was not able to replicate at all.

dakaneye:
@alfredodeza I'm not sure why you couldn't replicate, should be the same regardless of our different laptops unless you have a different environment than me somehow. 

For now, I'm ok w/ closing this b/c of your reasoning but still not sure what the best solution is...

alfredodeza:
I *think* it was tied to the version of `pip` - if you had the latest one you were safe and wouldn't hit this problem

PR #912, opened 2021-02-14
The allOf sections in these objects is breaking the swagger-codegen generated clients, so…

dakaneye:
… remove them, but keep the objects

**What this PR does / why we need it**: The branch name was a bit premature, but essentially the "allOf" logic in the swagger.yaml was affecting the ability of a python swagger codegen client to successfully be used (it resulted in some ImportErrors when the client got imported)

This PR replaces those blocks with the properties of the "inherited" definition. We don't get the allOf simplicity any more, but it allows the clients to still work and allows us to keep the definitions themselves, albeit with a little less reuse.

### DIFF
--- a/anchore_engine/services/apiext/swagger/swagger.yaml
+++ b/anchore_engine/services/apiext/swagger/swagger.yaml
@@ -2924,19 +2924,6 @@ paths:
         500:
           description: "Internal Error"
 definitions:
-  PaginationProperties:
-    type: object
-    properties:
-      page:
-        type: string
-        description: The page number returned (should match the requested page query string param)
-      next_page:
-        type: string
-        description: True if additional pages exist (page + 1) or False if this is the last page
-      returned_count:
-        type: integer
-        description: The number of items sent in this response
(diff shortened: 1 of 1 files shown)


nightfurys:
@dakaneye, I would like us to move in the direction of more reuse. It makes the APIs easier to read and consistent across the board (for instance image definition is slightly different depending on the route although it doesn't have to be). With the upgrade to swagger 3.0 we should be able to use oneOf, anyOf and other keywords for combining schemas with builtin validation. I understand that this could be a pain with codegen client, is there anyway we could patch the codegen to work correctly?

Vijay-P:
@zburstein your changes are failing CI. Please revisit when you get time.

zburstein:
@Vijay-P Turns out that my commits and one of Sam's commits in this branch are already present in master so when I rebased the branch, there is only one commit remaining. @dakaneye I just want to confirm that this PR/commit is still relevant and should still be merged given that the PR is in a pretty different state from what it was when it was first opened.

Issue #236, opened 2019-07-23
tag discovery logic 


armstrongli:
**Is this a request for help?**:
No. This is a feature enhancement
---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):
Feature request

**Version of Anchore Engine and Anchore CLI if applicable**:
All versions
 
**What happened**:
The tag discovery is based on skopeo, it has limitations on tags discovery. docker registry v2 provides the API to get tags directly without the dependency to must have a `lookuptag`. 
> https://\<registry url>/v2/\<repository>/tags/list

**What did you expect to happen**:
It is not needed to add the blasting fuse to trigger repository scan.

nurmi:
Hi @armstrongli - thank you for the FR!  This is something we've looked into, and have considered adding an FR to skopeo itself as a new option to scan a registry/repo for tags without requiring the (default) latest tag to be present.  If we were to add this capability into anchore directly, then the effort would require implementation of a native docker registry client (to handle auth) which is something we've tried to move away from (relying on the skopeo/OCI tooling as standard).

Interested to hear your thoughts on the above idea!

armstrongli:
@nurmi  thank you for your reply. Just from performance perspective, a native implementation will be good enough to handle it. 

all the command line tools to handle image fetch, image manifest check, and tags are all from skopeo command line , and they are all called by wrapping the command line tool to do it. Just from performance perspective, it is not good enough. From function part, it is good. it depends on the requirement.

to fetch tags from registry, there are tens of repos in github, and it is available to make it directly to use other tools rather than only one tool. 

How do you think about adding a new command tool to fetch tags?

armstrongli:
https://github.com/genuinetools/reg/releases

armstrongli:
here are some command examples:

fetch all repository manifest(with debug option `-d`):

```
$ reg-darwin-amd64 manifest -k --username='****' --password='****'  -d hub.tess.io/testuser/alpine
2019/07/24 15:29:22 registry.ping url=https://hub.tess.io/v2/
2019/07/24 15:29:24 registry.manifests uri=https://hub.tess.io/v2/testuser/alpine/manifests/latest repository=testuser/alpine ref=latest
2019/07/24 15:29:26 registry.manifests resp.Status=200 OK, body={
   "schemaVersion": 1,
   "name": "testuser/alpine",
   "tag": "latest",
   "architecture": "amd64",
   "fsLayers": [
      {
[...]

PR #918, opened 2021-02-18
Update Skopeo and use list-tags

anxolerd:
This PR updates Skopeo version in the docker image and utilizes `skopeo list-tags` command to retrieve tags from remote repository. This fixes issues with adding repositories which does not contain `latest` tag to anchore-engine.

References: 
- #236
- #484


zhill:
Hi @anxolerd ! First, thanks for the work on this. It's something we've had on the to-do list for a while. We'll review!

zhill:
With this update, it's also possible to remove: https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/apiext/swagger/swagger.yaml#L1164 since that API parameter is used for handling the old behavior of skopeo.

anxolerd:
@zhill , I've deleted some should-be outdated and no more required references to lookuptag in 06ab80c.

zhill:
Thanks for the update @anxolerd ! Will do some testing and review and update here. This is much appreciated work. Thanks!

zhill (approved):
lgtm

dakaneye (approved):
This looks good to me! Thanks @anxolerd !

It's cool that we don't need a lookup tag anymore as well, which is confusing when you're trying to add a whole "repo"

@zhill I think it'd be really useful to have some integration tests for the skopeo_wrapper just so we can have a bit more confidence around changes here, but I tested these locally and verified the skopeo command (on v1.2.1) works well and should output the correct format used here

zhill:
Thanks @anxolerd and @dakaneye !

PR #929, opened 2021-03-02
Removed schema from delete responses

### LINKED ISSUE #928: Fix delete routes return values in swagger spec
Archive rule deletion route in the swagger spec specifies an incorrect value. Currently it specifies the 200 return to use the rule schema, but the real return value is an empty dictionary. This is causing issues with the generated client when making deletion calls because it expects transition to be set on the returned object. 

There may be additional deletion routes that have similar issues and should be investigated.

zburstein:
Fixes #928 


dakaneye (approved):
LGTM!

PR #933, opened 2021-03-04
Fix the client metadata merge process during oauth init. Fixes #931

### LINKED ISSUE #931: Incorrect merge logic in oauth.py causes invalid client config for upgrade to v0.9.x from v0.8.x if using oauth tokens
The merge update that occurs on the service start is not always creating the correct client metadata dict to be saved in the DB. This results in "invalid_client" errors. The workaround is to delete the row from the database and let it be re-created from scratch (in the _oauth_clients_ table) and the new record created is correct.

The upgrade logic is not setting other properties aside from the "grant_types" correctly.

Affects upgrades from v0.8.x -> v0.9.x.

zhill:


dakaneye (approved):
lgtm

PR #949, opened 2021-03-21
Fix Java pom.properties parsing and lang-pack matching

### LINKED ISSUE #950: Missing GHSA vulnerabilities for java packages due to missing metadata post syft integration
Engine analyzer for java packages supply raw pom properties (`pom.properties`) in the metadata for the java package. This data is used for driving matches against GHSA feeds https://github.com/anchore/anchore-engine/blob/v0.9.0/anchore_engine/db/entities/policy_engine.py#L1775-L1786. 

Syft integration replaced the engine's java analyzer in https://github.com/anchore/anchore-engine/pull/693 that was introduced in 0.9.0. The metadata that engine relied on finding GHSA matches was no longer available after the switch. It resulted in the engine missing GHSA matches for java/maven packages

wagoodman:
This PR achieves a few things:
- Bumps syft to v0.14.0 to pull in https://github.com/anchore/syft/pull/348 to allow for both `:` and `=` delimiters in parsed pom.property files when analyzing java packages.
- https://github.com/anchore/anchore-engine/pull/693 removed the java `metadata_json` data in the policy engine loader, which was used for the lang-pack matching (which is how GHSA matches are discovered). This PR synthesizes a pom.properties file and stores it in the `metadata_json` object.

Fixes #950

### DIFF
--- a/anchore_engine/analyzers/syft/handlers/java.py
+++ b/anchore_engine/analyzers/syft/handlers/java.py
@@ -55,6 +55,18 @@ def translate_and_save_entry(findings, artifact):
     if group_id:
         origin = group_id
 
+    # synthesize a part of the pom.properties
+    pom_artifact_id = dig(artifact, "metadata", "pomProperties", "artifactId")
+    pom_version = dig(artifact, "metadata", "pomProperties", "version")
+
+    pomProperties = """
+groupId={}
+artifactId={}
+version={}
+""".format(
+        group_id, pom_artifact_id, pom_version
+    )
+
     pkg_value = {
         "name": artifact["name"],
         "specification-version": values.get("Specification-Version", "N/A"),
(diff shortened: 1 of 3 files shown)


nightfurys (approved):
lgtm

nightfurys:
Addresses https://github.com/anchore/anchore-engine/issues/950

anchore_engine/analyzers/syft/handlers/java.py
```diff
@@ -55,6 +55,18 @@ def translate_and_save_entry(findings, artifact):
...
+    # synthesize a part of the pom.properties
+    pom_artifact_id = dig(artifact, "metadata", "pomProperties", "artifactId")
+    pom_version = dig(artifact, "metadata", "pomProperties", "version")
+
+    pomProperties = """
+groupId={}
+artifactId={}
+version={}
+""".format(
+        group_id, pom_artifact_id, pom_version
```
thehh1974, line 67: This actually creates a bug if the properties are missing when parsing the properties later. 
For example "version=None" is parsed to `"version": "None"` instead of `"version": None`.
Then, anywhere we check `if version:` we get True instead of false.

PR #957, opened 2021-03-26
Remove CPE hashing

dakaneye:
(it is a remnant of inaccurate de-dup logic from before the improvements added in 3.0)

### DIFF
--- a/anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
+++ b/anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
@@ -978,7 +978,6 @@ def get_image_vulnerabilities(user_id, image_id, force_refresh=False, vendor_onl
 
         cpe_vuln_listing = []
         try:
-            all_cpe_matches = []
             with timer("Image vulnerabilities cpe matches", log_level="debug"):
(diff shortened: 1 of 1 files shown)


nightfurys (approved):
lgtm. Yay for losing redundant dedups!

dakaneye:
All of the removed string methods were not used (I did a string search for those method names). It seemed that only the method from ImageCpe was actually used, and I ported that over to be a static method in the common/helpers.py file with some unit tests. LMK what you think.

nightfurys (changes requested):
Refactored methods are not all the same. There are subtle differences between the functions in ImageCpe and CpeV2Vulnerability/VulnDBCpe

nightfurys:
> All of the removed string methods were not used (I did a string search for those method names). It seemed that only the method from ImageCpe was actually used, and I ported that over to be a static method in the common/helpers.py file with some unit tests. LMK what you think.

get_cpe23string function in CpeV2Vulnerability/VulnDBCpe is used
https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py#L1008

nightfurys (approved):

Issue #867, opened 2021-01-26
broad try/except blocks makes it hard to handle exceptions


alfredodeza:
There is a wide-use of a Python exception handling anti-pattern in the code base. This anti-pattern looks like this:

```python
try:
    logger.info("some info message")
    db.query.images.get(id=1)
except Exception:
    logger.exception("error encountered when querying")
```

This is problematic in many different ways:

1. The `try` part has usually more than one call and can have many lines. A `logger.info` will probably never cause an exception, and should not be caught. Catching exceptions should be done as targeted as possible.
2. There should be a good reason to catch every single exception possible, and in this case (as it is with most cases in the code base) there isn't one. Exception catching should be as granular as possible - not generic and broad. This allows flow control by callers by catching individual exceptions.
3. It is impossible to test this properly because the exception does not bubble up
4. In most cases where this pattern is found, parent callers of the function are _also doing the same catching_ which doesn't make sense for the apparent goal of producing a message when any exception is raised.
5. It makes it extremely hard, if not impossible, to debug. It is becoming useful to remove whole try/except blocks when debugging because of this pattern. 

Exceptions are there to allow behavior changes and make logical decisions on what to do next. Broad exceptions do not allow that and forces this pattern everywhere else to suppress errors.

Although this is technically a _"tech debt"_ problem - it has been raised before, and the issue is significant, so I'm going to add this as a _"bug"_ category/label as well. It is pressing to address this problem.

zhill:
This can probably be broken down into a few different issues we can tackle as part of a milestone to cleanup.
1. Explicit code for the "fail safe" pattern where some code needs to execute and no "reasonable" (e.g. not a RuntimeException) should cause it to exit with an exception that it doesn't explicitly throw. This would help identify these cases and distinguish from lazy handling.
2. Create a better set of specific Exceptions for Engine. There are parts of the code that have their own but far too many cases just raise Exception types with a message. Without this change the handlers must necessarily be generic.
3. Consistent mappers in the API layers to handle and raise AnchoreApiError exceptions rather than returning specific response payloads. This is used in some services but not others, so we have a mix of catch-all and return using 'make_error_response()' and raising the more specific exceptions that the web stack will catch and has proper handlers for marshalling responses into HTTP.

I do not think, however, that this is something that can be done randomly or as part of other feature work/bug fixes. We'll need to divide up the code into sections and work thru it systematically since these changes will certainly cause potential regressions in control flow and user-facing error responses.

PR #961, opened 2021-04-01
Add image code Refactor

dakaneye:
This is an attempt to modularize and unit test the add-image workflow (by tag/digest)
* This should _not_ be modifying the add-from-archive or import-image-content workflow
* This should not be changing _ANY_ functionality


anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -1826,6 +1873,30 @@ def perform_policy_evaluation(
     return curr_evaluation_record, curr_evaluation_result
 
 
+ImageKey = collections.namedtuple("ImageKey", ["tag", "digest"])
```
dakaneye, line 1876: this bit is new, and meant to support other features who _have_ an image tag and digest, and want it to be easy to add an image for analysis

dakaneye: Note: this flow will NOT create any subscriptions for the image, that bit happens currently in the external API service

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -483,201 +482,240 @@ def image(dbsession, request_inputs, bodycontent=None):
...
+                image_info,
+                input_string,
+            )
+    except AnchoreApiError as err:
+        logger.exception("Error processing image request")
+        return_object = anchore_engine.common.helpers.make_response_error(
+            err.message, in_httpcode=err.__response_code__, details=err.detail
+        )
+        httpcode = err.__response_code__
+    except Exception as err:
```
alfredodeza, line 500: I think we should not allow blanket exceptions anymore as part of new additions https://github.com/anchore/anchore-engine/issues/867

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -483,201 +482,240 @@ def image(dbsession, request_inputs, bodycontent=None):
...
-                                detail={
-                                    "digest": image_info["digest"],
-                                    "tag": image_info["fulltag"],
-                                },
-                            )
+    except AnchoreApiError:
+        raise
+    except Exception as err:
+        logger.exception("Error adding image")
+        raise err
```
alfredodeza, line 620: The only thing that this block is adding is a log entry mentioning that there is an error adding an image. The cost is high: indents the code into a large try/except block, makes it hard to debug, and it not necessary. Similar as previous comment, for reference, issue: https://github.com/anchore/anchore-engine/issues/867

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -483,201 +482,240 @@ def image(dbsession, request_inputs, bodycontent=None):
...
+        fail_event = anchore_engine.subsys.events.ImageRegistryLookupFailed(
+            user_id=account_id,
+            image_pull_string=input_string,
+            data=err.__dict__,
+        )
+        try:
+            add_event(fail_event, dbsession)
+        except Exception:
+            logger.warn("Ignoring error creating image registry lookup event")
+        raise err
```
alfredodeza, line 686: similar to other broad exception comments

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -483,201 +482,240 @@ def image(dbsession, request_inputs, bodycontent=None):
...
-    return return_object, httpcode
+
+def get_manifest(image_info):
+    try:
+        if "manifest" in image_info:
+            manifest = json.dumps(image_info["manifest"])
+        else:
+            raise Exception("no manifest from get_image_info")
+    except Exception as err:
+        raise Exception("could not fetch/parse manifest - exception: " + str(err))
```
alfredodeza, line 717: This is similar to the other exception handling, but causes a problem: the original exception is being caught and only added in its string representation to a new error that adds a message, obfuscating the originating exception. This will cause debugging to be really hard

alfredodeza (changes requested):

alfredodeza:
I guess I didn't read the subject of the PR that well that says that nothing should be changing functionality :( Feel free to ignore the exception comments if they don't apply @dakaneye

dakaneye:
> I guess I didn't read the subject of the PR that well that says that nothing should be changing functionality :( Feel free to ignore the exception comments if they don't apply @dakaneye

I think we could address the exception handling possibly, @zhill what do you think?

anchore_engine/services/catalog/catalog_impl.py
Vijay-P: any reason why we don't use dataclasses instead of namedtuples? we're using 3.8 now so compatibility isn't an issue anymore.

Vijay-P: https://medium.com/@jacktator/dataclass-vs-namedtuple-vs-object-for-performance-optimization-in-python-691e234253b9

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -483,201 +482,240 @@ def image(dbsession, request_inputs, bodycontent=None):
                     raise Exception("image not found in DB")
 
         elif method == "POST":
-            timer = time.time()
+            return_object, httpcode = image_post(
```
Vijay-P, line 485: if we're going this far, might as well split the handlers for get and post into two different methods, create a common pre-processing method, and update the calls in anchore_engine/services/catalog/api/controllers/default_controller.py. will be easier to read. having the POST handler tacked on to the end of the images() method feels like an afterthought

zhill:
> > I guess I didn't read the subject of the PR that well that says that nothing should be changing functionality :( Feel free to ignore the exception comments if they don't apply @dakaneye
> 
> I think we could address the exception handling possibly, @zhill what do you think? I was trying to be risk averse here but it seems like a worthwhile improvement.

I think it is good to address the cases in the direct code paths changed here, but we should be careful of scope-creep if we start delving into the methods called. Agreed they all need review, but should be done in small focused updates if we start modifying functions called here but not modified in this PR.

anchore_engine/services/catalog/catalog_impl.py
dakaneye: I didn't even know about data classes! I used this just because I was more familiar. There are some python 3.6 relics left in engine (around integration tests, I believe), so I am reticent to adding this now, but super glad you shared this so I know about it for the future!

anchore_engine/services/catalog/catalog_impl.py
dakaneye: I think specifically, I want to leave this error because there are lots of possibilities of errors to be thrown here, but this should be a target for a later refactor

anchore_engine/services/catalog/catalog_impl.py
dakaneye: What would you think about refactoring this in a subsequent change? Wanted to minimize the impact here as there _is_ some pre-processing that is common for the two methods for now

anchore_engine/services/catalog/catalog_impl.py
dakaneye: This one looks like it's squelching an error when a "fail" event _fails_ to be submitted....and might need some thinking about re-implementation, because removing this WOULD change the functionality here (main goal was to refactor and modularize)

dakaneye:
@alfredodeza @Vijay-P @zhill I addressed some of the comments and updated the branch against master. LMK what you think. 

Some of the exception blocks are hard to remove (not impossible) but I _did_ want to limit the scope of the change here as much as possible. So I did what I could for now, and happy to open more follow up tech debt tasks.

anchore_engine/services/catalog/catalog_impl.py
alfredodeza: Unfortunately that doesn't end up working well for testing, debugging or how the rest of the code-base deals with exceptions. The `image()` function gets called in a couple of places. One is in `list_images()` and the other one in `add_image()`, *both of which* do enormous try/except blocks that are also catching anything that might get raised. 
As I explain in #867, this broad exception catching at multiple levels in the stack makes it impossible to debug, and causes problems for testing.

I'm going to be a -1 on adding more broad-exception catches like this one. There needs to be a solid reason for catching everything, while avoiding multiple broad try/except blocks that are catching exceptions.

dakaneye: @alfredodeza I totally agree with you here, but I have two comments:
* I'm not adding this exception block (it's not new), just ported over as I modularized things
* If I change this logic now...I need to go through each sub method, identify _all possible_ errors, and be 100% sure that expected use cases don't end up missing a more specific block. At this point, that's not a risk I'm looking to take as this is a highly used path in the engine. At this point I am just trying to do the bare bones modularization...but you are 100% right that this is bad and needs to change, I'm just trying to avoid doing that change in this PR.

anchore_engine/services/catalog/catalog_impl.py
Vijay-P: That's fine w me.

Vijay-P (approved):
LGTM, but would like to see more work done to this part of the code in the near future.

alfredodeza (approved):

PR #977, opened 2021-04-21
Add grype_wrapper facade

dspalmer99:
Adds a grype_wrapper facade, along with tests, as a client to use grype for sbom analysis and vulnerability queries. It also installs grype as part of image build.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #:978

**Special notes**:


nightfurys (approved):
I've mainly focussed on the grype-wrapper and it looks good to me. We may want to update the "get checksum" to "get metadata" to provide more metadata

dspalmer99:
Thanks @nightfurys . I'll include that metadata bit in the follow-up PR - as per our earlier conversation I am going to merge this so we can start integrating.

PR #983, opened 2021-04-22
Impose list ordering before result comparisons

nightfurys:
PR fixes an intermittent functional test failure. The test compares a list of expected results to the API response. There are no ordering guarantees from the API and this causes the comparison to fail. The fix orders the lists before comparing them.

### DIFF
--- a/tests/functional/services/policy_engine/conftest.py
+++ b/tests/functional/services/policy_engine/conftest.py
@@ -42,14 +42,14 @@ def read_expected_content(module_path, filename):
 
 
 @pytest.fixture
-def expected_content(request) -> Callable[[str], Dict]:
+def expected_content(request) -> Callable:
     """
     Returns method that will load expected vulnerability response json for a given image_digest
     :rtype: Callable[[str], Dict]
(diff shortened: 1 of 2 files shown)


tests/functional/services/policy_engine/conftest.py
```diff
@@ -42,14 +42,14 @@ def read_expected_content(module_path, filename):
...
 @pytest.fixture
-def expected_content(request) -> Callable[[str], Dict]:
+def expected_content(request) -> Callable:
     """
     Returns method that will load expected vulnerability response json for a given image_digest
     :rtype: Callable[[str], Dict]
     :return: method that loads expected response json
     """
 
-    def get_expected_content(filename) -> Dict:
```
Vijay-P, line 52: Should probably change this to Union[List, Dict]

nightfurys: Yep, I think it might be okay to leave it open since the type enforcement is merely a hint.

Vijay-P (approved):

PR #985, opened 2021-04-23
Adds FIPS support by updating hashlib usage to include usedforsecurity=False, and builds psycopg2 from source

zhill:
Adds FIPS support changes necessary for identifying non-crypto use of md5 hashing and updates psycopg2 to build from source instead of using the psycopg2-binary install so that the install will use versions of ssl that respect FIPS flags.

Need to test if a container build without the FIPS flag set on the host will have the correct options compiled in when run on a FIPS-enabled host.


zhill:
To verify will require testing w/TLS enabled in the services directly as well as for the DB connection and any external storage connections such as S3 or Swift to ensure all those paths are verified as well.

zhill:
Will be following up with an updated approach for a simpler way to do this.

zhill:
Will merge changes from dev/nurmi/fipsbase to get the combo of both.

PR #988, opened 2021-04-26
Grype facade string sbom

dspalmer99:
**What this PR does / why we need it**: Adds the ability to pass sboms to the grype facade as a string or a file path, and consequently updates utils to allow the execution of piped commands. Protects against requests to update the grype db where the new file does not exist. Adds a good number of unit tests.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #: 978

**Special notes**:

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -8,7 +8,7 @@ import tarfile
 
 from anchore_engine.db.entities.common import UtilMixin
 from anchore_engine.subsys import logger
-from anchore_engine.utils import run_check, CommandException
+from anchore_engine.utils import CommandException, run_piped_command, run_check
 from readerwriterlock import rwlock
 from sqlalchemy import Column, String, Integer, ForeignKey
 from sqlalchemy.ext.declarative import declarative_base
@@ -20,19 +20,11 @@ grype_db_session = None
 grype_db_lock = rwlock.RWLockWrite()
 Base = declarative_base()
 
+GRYPE_SUB_CMD = "grype -vv -o json"
 VULNERABILITY_FILE_NAME = "vulnerability.db"
+METADATA_FILE_NAME = "metadata.json"
 VULNERABILITY_TABLE_NAME = "vulnerability"
 VULNERABILITY_METADATA_TABLE_NAME = "vulnerability_metadata"
-
-
-# TODO Implement, once added to Grype
-def get_current_grype_db_checksum():
-    """
-    Return the checksum for the in-use version of grype db
-    """
-    grype_db_checksum = None
-    logger.info("Returning current grype_db checksum: {}".format(grype_db_checksum))
-    return grype_db_checksum
 
 
 def _get_default_grype_db_dir_from_config():
@@ -163,12 +155,6 @@ def _init_grype_db(lastest_grype_db_archive: str, version_name: str):
     Write the db string to file, create the engine, and create the session
     Return the file and session
     """
-    # TODO Further refine this, we should be extracting everything to a subdir, and letting the
-    # replacement logic operate at that level (ie always write and remove the contents of an
-    # archive file to a subdir under '<service-dir>/grype_db/', rather than operating on
-    # the specific vulnerability.db file.
-    # It sounds like we will also be getting the metata.json file, and the further contents
-    # of those archives could change if they needed to.
     latest_grype_db_dir = _move_and_open_grype_db_archive(
         lastest_grype_db_archive, version_name
     )
@@ -231,32 +217,92 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
             write_lock.release()
 
 
-def get_vulnerabilities(grype_sbom: str) -> json:
+def get_current_grype_db_metadata() -> json:
     """
-    Use grype to scan the provided sbom for vulnerabilites.
+    Return the json contents of the metadata file for the in-use version of grype db
     """
     global grype_db_dir
 
+    # Get the path to the latest grype_db metadata file
+    latest_grype_db_metadata_file = os.path.join(grype_db_dir, METADATA_FILE_NAME)
+
+    # Ensure the file exists
+    if not os.path.exists(latest_grype_db_metadata_file):
+        # If not, return None
+        return None
+    else:
+        # Get the contents of the file
+        with open(latest_grype_db_metadata_file) as read_file:
+            json_output = json.load(read_file)
+            return json_output
+
+
+def _get_proc_env():
+    global grype_db_dir
+
+    # Set grype env variables, including the grype db location
+    grype_env = {
+        "GRYPE_CHECK_FOR_APP_UPDATE": "0",
+        "GRYPE_LOG_STRUCTURED": "1",
+        "GRYPE_DB_AUTO_UPDATE": "0",
+        "GRYPE_DB_CACHE_DIR": "{}".format(grype_db_dir),
+    }
+    proc_env = os.environ.copy()
+    proc_env.update(grype_env)
+    return proc_env
+
+
+def get_vulnerabilities_for_sbom(grype_sbom: str) -> json:
+    """
+    Use grype to scan the provided sbom for vulnerabilites.
+    """
     # Get the read lock
     read_lock = grype_db_lock.gen_rlock()
     if read_lock.acquire(blocking=False, timeout=60):
         try:
-            # Set grype env variables, including the grype db location
-            grype_env = {
-                "GRYPE_CHECK_FOR_APP_UPDATE": "0",
-                "GRYPE_LOG_STRUCTURED": "1",
-                "GRYPE_DB_AUTO_UPDATE": "0",
-                "GRYPE_DB_CACHE_DIR": "{}".format(grype_db_dir),
-            }
+            # Get env variables to run the grype scan with
+            proc_env = _get_proc_env()
 
-            proc_env = os.environ.copy()
-            proc_env.update(grype_env)
+            # Format and run the command. Grype supports piping in an sbom string, so we need to this in two steps.
+            # 1) Echo the sbom string to std_out
+            # 2) Pipe that into grype
+            pipe_sub_cmd = 'echo \'{sbom}\''.format(
+                sbom=grype_sbom,
+            )
+            full_cmd = [shlex.split(pipe_sub_cmd), shlex.split(GRYPE_SUB_CMD)]
+
+            logger.debug("Running grype with command: {} | {}".format(pipe_sub_cmd, GRYPE_SUB_CMD))
+
+            stdout = None
+            err = None
+            try:
+                _, stdout, _ = run_piped_command(full_cmd, env=proc_env)
+            except CommandException as exc:
+                logger.error(
+                    "Exception running command: {} | {}, stderr: {}".format(pipe_sub_cmd, GRYPE_SUB_CMD, exc.stderr)
+                )
+                raise exc
+        finally:
+            read_lock.release()
+
+        # Return the output as json
+        return json.loads(stdout.decode("utf-8"))
+
+
+def get_vulnerabilities_for_sbom_file(grype_sbom_file: str) -> json:
+    """
+    Use grype to scan the provided sbom for vulnerabilites.
+    """
+    # Get the read lock
+    read_lock = grype_db_lock.gen_rlock()
+    if read_lock.acquire(blocking=False, timeout=60):
+        try:
+            # Get env variables to run the grype scan with
+            proc_env = _get_proc_env()
 
             # Format and run the command
-            # TODO This is currently expecting an sbom file reference, not the actual sbom string itself
-            # Should we support both, or just the string?
             cmd = "grype -vv -o json sbom:{sbom}".format(
-                sbom=grype_sbom,
+                sbom=grype_sbom_file,
             )
 
             logger.debug("Running grype with command: {}".format(cmd))
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -1,9 +1,7 @@
-import errno
-from json.decoder import JSONDecodeError
-
 import anchore_engine.configuration.localconfig
-import os
+import errno
 import json
+import os
 import shlex
 import shutil
 import sqlalchemy
@@ -12,6 +10,7 @@ import tarfile
 from anchore_engine.db.entities.common import UtilMixin
 from anchore_engine.subsys import logger
 from anchore_engine.utils import CommandException, run_check, run_piped_command_list
+from json.decoder import JSONDecodeError
 from readerwriterlock import rwlock
 from sqlalchemy import Column, String, Integer, ForeignKey
 from sqlalchemy.ext.declarative import declarative_base
@@ -67,7 +66,7 @@ def _move_grype_db_archive(
         raise FileNotFoundError(
             errno.ENOENT,
             ARCHIVE_FILE_NOT_FOUND_ERROR_MESSAGE,
-            grype_db_archive_local_file_location
+            grype_db_archive_local_file_location,
         )
     else:
         # Move the archive file
@@ -255,7 +254,9 @@ def get_current_grype_db_metadata() -> json:
             try:
                 return json.load(read_file)
             except JSONDecodeError:
-                logger.error("Unable to decode grype_db metadata file into json: %s", read_file)
+                logger.error(
+                    "Unable to decode grype_db metadata file into json: %s", read_file
+                )
                 return None
 
 
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -1,3 +1,6 @@
+import errno
+from json.decoder import JSONDecodeError
+
 import anchore_engine.configuration.localconfig
 import os
 import json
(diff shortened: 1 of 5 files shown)


anchore_engine/utils.py
```diff
@@ -218,6 +218,39 @@ def shellcheck(x):
...
+def run_piped_command_list(
+    cmd_lists, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs
+):
+    """
+    Executes a series of subcommands from cmd_lists, the stdout result of each being piped stdin for the next.
+    For example, the command: 'echo {}' | grype
+    cmd_lists should be passed as: [['echo', '{}'], ['grype']]
+    Do not include the actual pipe symbol, it is inferred from the data structure.
+    """
+    if cmd_lists:
```
alfredodeza, line 230: in what situation the `cmd_lists` wouldn't be passed in? why is it useful to allow this?

tests/unit/anchore_engine/clients/test_grype_wrapper.py
```diff
@@ -102,6 +96,30 @@ def test_move_grype_db_archive(tmp_path, grype_db_archive):
     )
 
 
+def test_move_missing_grype_db_archive(tmp_path):
+    # Setup non-existent input archive and real output dir
+    missing_output_archive = "/does/not/exist.tar.gz"
+    output_dir = os.path.join(tmp_path, "output")
```
alfredodeza, line 102: great use of pytest fixtures here!

tests/unit/anchore_engine/clients/test_grype_wrapper.py
```diff
@@ -204,21 +222,71 @@ def test_update_grype_db(grype_db_parent_dir, old_grype_db_dir, grype_db_archive
     assert not os.path.exists(old_grype_db_dir)
 
 
-# TODO This needs to move to a functional test to pass on the ci, as the container with grype
-# is not built before that. Disabling for now. Works locally if you have grype installed.
+def test_get_current_grype_db_metadata(grype_db_dir):
+    # Setup test input
+    grype_wrapper.grype_db_dir = grype_db_dir
```
alfredodeza, line 227: this is exactly the kind of thing that globals will cause. Now this test has effectively tainted the global variable, if this test were to fail, the next test that relies on this value could potentially fail. Figuring out which test did what and when to the global is time-consuming to say the least and all of this can be avoided by removing the global altogether

tests/unit/anchore_engine/test_utils.py
```diff
@@ -92,6 +99,58 @@ def test_parse_dockerimage_string(image, expected):
...
+    # Function under test
+    output_cmd_list = run_sanitize(input_cmd_list)
+
+    # Validate input
+    output_cmd_list == input_cmd_list
+
+
+@pytest.mark.parametrize(
+    "input",
+    [";&<>", ";", "&", "<", ">"],
```
alfredodeza, line 115: thank you for adding these!

tests/unit/anchore_engine/test_utils.py
```diff
@@ -92,6 +99,58 @@ def test_parse_dockerimage_string(image, expected):
...
+)
+def test_run_piped_command(
+    cmds_list, expected_return_code, expected_stdout, expected_stderr
+):
+    # Function under test
+    return_code, stdout, stderr = run_piped_command_list(cmds_list)
+
+    # Binary string returned in different environments can be padded with different amounts of whitespace
+    # So where expected_output != None, convert it to utf-8 and trim it so we get a clean, reliable comparison
+    if stdout is not None:
```
alfredodeza, line 134: I think it is very odd to have the function return `None, None, None` - this type of interaction with that function complicates how to deal with the output. I think it would be best to always return the expected values or raise an exception

tests/unit/anchore_engine/test_utils.py
```diff
@@ -92,6 +99,58 @@ def test_parse_dockerimage_string(image, expected):
...
+    assert stderr == expected_stderr
+
+
+def test_run_invalid_piped_command():
+    # Setup input var
+    first_command = shlex.split("pwd")
+    bad_command = shlex.split("<")
+    cmds_list = [first_command, bad_command]
+
+    with pytest.raises(Exception):
```
alfredodeza, line 149: if this is catching any exception then at least the message should be verified, otherwise it might be too broad - making this test not that useful (false positive)

alfredodeza (changes requested):

anchore_engine/clients/grype_wrapper.py
```diff
@@ -60,16 +52,26 @@ def _move_grype_db_archive(
...
-    logger.info(
-        "Moving the grype_db archive from {} to {}".format(
+    if not os.path.exists(grype_db_archive_local_file_location):
+        logger.warn(
+            "Unable to move grype_db archive from {} to {} because it does not exist".format(
+                grype_db_archive_local_file_location,
+                grype_db_archive_copied_file_location,
+            )
+        )
+        raise FileNotFoundError
```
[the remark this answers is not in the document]
dspalmer99: Will do

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +227,98 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
             write_lock.release()
 
 
-def get_vulnerabilities(grype_sbom: str) -> json:
+def get_current_grype_db_metadata() -> json:
     """
-    Use grype to scan the provided sbom for vulnerabilites.
+    Return the json contents of the metadata file for the in-use version of grype db
     """
     global grype_db_dir
```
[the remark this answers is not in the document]
dspalmer99: I've considered refactoring this whole module into a class so these state variables, currently implemented as globals, can just become data on the class. Not in scope for this change, but still something I would like to do. @nightfurys What do you think? We'd probably have to make sure it was created as a singleton during startup so that all of the users of this facade get the same instance.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +227,98 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
...
+    latest_grype_db_metadata_file = os.path.join(grype_db_dir, METADATA_FILE_NAME)
+
+    # Ensure the file exists
+    if not os.path.exists(latest_grype_db_metadata_file):
+        # If not, return None
+        return None
+    else:
+        # Get the contents of the file
+        with open(latest_grype_db_metadata_file) as read_file:
+            json_output = json.load(read_file)
```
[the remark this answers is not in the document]
dspalmer99: Will do

anchore_engine/utils.py
dspalmer99: As you said below, raising an exception is more apropriate. The only scenario is user error, the whole goal of protecting this block was to prevent an NPE on output.communicate()

dspalmer99: So I will make that change as well

tests/unit/anchore_engine/clients/test_grype_wrapper.py
dspalmer99: Thanks

tests/unit/anchore_engine/test_utils.py
dspalmer99: Np

tests/unit/anchore_engine/test_utils.py
dspalmer99: I will verify the message, but `Exception` is what the function under test raises, and it is used in enough places that I don't want to pull the testing involved on changing that into this change.

nightfurys:
Few comments with reference to global variable access
- reduce the global variable read access by refactoring - create functions such as get_grype_db_session that access the global variable and return it. Would make swapping out with an other implementation if we decide to do that later
- move the global value writes to init functions for the same reason - set grype_db_session in __init_grype_db_engine. This is also consistent with how db connection is instantiated [engine](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/db/entities/common.py#L148-L196) 

Circle CI is failing on the linter. I use black connect plugin in pycharm with linting on save option, makes linting a no-op for me

nightfurys (commented):
Few comments with reference to global variable access
- reduce the global variable read access by refactoring - create functions such as get_grype_db_session that access the global variable and return it. Would make swapping out with an other implementation if we decide to do that later
- move the global value writes to init functions for the same reason - set grype_db_session in __init_grype_db_engine. This is also consistent with how db connection is instantiated [engine](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/db/entities/common.py#L148-L196) 

Circle CI is failing on the linter. I use black connect plugin in pycharm with linting on save option, makes linting a no-op for me

dspalmer99:
@nightfurys @alfredodeza I've made the changes you both requested, anything else in this one before I merge to the feature branch?

tests/unit/anchore_engine/clients/test_grype_wrapper.py
Vijay-P: @dspalmer99 do you think the globals could be refactored to a singleton? Or class variables? I really dislike the use of globals here.

tests/unit/anchore_engine/test_utils.py
alfredodeza: Yeah, if the underlying code raises `Exception` this is going to be more involved. Bummer :(

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +237,111 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
             write_lock.release()
 
 
-def get_vulnerabilities(grype_sbom: str) -> json:
+def _get_grype_db_dir():
+    global grype_db_dir
+    return grype_db_dir
```
alfredodeza, line 242: I'm perhaps not understanding this well, but I'm not sure why `grype_db_dir` needs to be a global. Using a function here is definitely an improvement, but I can't reason why `global` is still being required

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +237,111 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
...
+        # If not, return None
+        return None
+    else:
+        # Get the contents of the file
+        with open(latest_grype_db_metadata_file) as read_file:
+            try:
+                return json.load(read_file)
+            except JSONDecodeError:
+                logger.error(
+                    "Unable to decode grype_db metadata file into json: %s", read_file
```
alfredodeza, line 270: excellent!

anchore_engine/utils.py
```diff
@@ -213,11 +215,44 @@ def shellcheck(x):
...
+):
+    """
+    Executes a series of subcommands from cmd_lists, the stdout result of each being piped stdin for the next.
+    For example, the command: 'echo {}' | grype
+    cmd_lists should be passed as: [['echo', '{}'], ['grype']]
+    Do not include the actual pipe symbol, it is inferred from the data structure.
+    """
+    if not cmd_lists:
+        raise ValueError(PIPED_CMD_VALUE_ERROR_MESSAGE)
+    else:
```
alfredodeza, line 234: This `else` is no longer needed here

anchore_engine/utils.py
```diff
@@ -213,11 +215,44 @@ def shellcheck(x):
...
+        output = None
+        for cmd_list in cmd_lists:
```
alfredodeza, line 236: This is a nice way of writing a readable for-loop. `for book in books` is the usual example I give, and this implements it well. Thanks for writing it this way

nightfurys (approved):

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +237,111 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
...
-            }
+            # Get env variables to run the grype scan with
+            proc_env = _get_proc_env()
 
-            proc_env = os.environ.copy()
-            proc_env.update(grype_env)
+            # Format and run the command. Grype supports piping in an sbom string, so we need to this in two steps.
+            # 1) Echo the sbom string to std_out
+            # 2) Pipe that into grype
+            pipe_sub_cmd = "echo '{sbom}'".format(
```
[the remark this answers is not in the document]
dspalmer99: I had originally had some code in here (https://github.com/anchore/anchore-engine/blob/83cc719ada1469cab438d433f9b6e5796995418b/anchore_engine/clients/grype_wrapper.py#L255) that was going down the road of  allowing grype to be run more broadly, but took it out, primarily because I don't want to add maintenance and testing burden to the team for code we aren't actively using, but also because our use case here is that engine will build and manage grype db. If we also allow it to be run with latest grype db it pulls down itself that has a good chance of introducing confusion.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -231,32 +237,111 @@ def update_grype_db(grype_db_archive_local_file_location: str, version_name: str
...
+    # Get the path to the latest grype_db metadata file
+    latest_grype_db_metadata_file = os.path.join(
+        _get_grype_db_dir(), METADATA_FILE_NAME
+    )
+
+    # Ensure the file exists
+    if not os.path.exists(latest_grype_db_metadata_file):
+        # If not, return None
+        return None
+    else:
```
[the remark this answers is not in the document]
dspalmer99: I think this is a personal preference; to me it is more readable this way.

anchore_engine/clients/grype_wrapper.py
dspalmer99: Possibly I am not understanding the `global` correctly - my intention here (and previously when these variables were accessed elsewhere) was that when accessing these we always want the module-scoped `grype_db_dir` and `grype_db_session` declared above. This module differs from the other wrappers in that it needs to maintain state to be applied on all get/query vulnerabilities requests, and allow that state to be updated externally.

Coming from a Java background I also find allowing module-external, direct variable access to be not great, and would rather refactor this whole thing into a class with a method that returns a singleton instance of the class. But 1) I'd need to discuss the startup/instantiation implications of that with the rest of our feature group first, 2) that will break the pattern of the other wrappers (but may be worth it), and 3) that is really outside the scope of this change.

tests/unit/anchore_engine/clients/test_grype_wrapper.py
dspalmer99: Maybe, but as I said in another comment, I'd rather do something like make the whole module a singleton class in that case. Which I am not opposed to doing (and would probably prefer) but I'd want to discuss it with the group in our sync first, and in any case it is outside the scope of this pr.

anchore_engine/clients/grype_wrapper.py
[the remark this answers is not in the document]
alfredodeza: I'll make this a nit then: whenever a method or function is doing many different things, or directly responsible for many things, that creates a problem: it is harder to maintain, harder to reason about, and harder to write tests. 
Writing tests, or writing code that needs tests - and tests in general - is _not a burden_. The more tested code (regardless of amount of usage) the better.

anchore_engine/clients/grype_wrapper.py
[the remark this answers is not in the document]
alfredodeza: I'll defer this one to other reviewers, but in general, most (if not all) non-error-producing comments can be labeled as "personal preference". You might be right that this in particular is a preference to me, but this has some core reasons why that is so:

Adding code that might make it obvious _at the expense of_ making it redundant causes others to double check the intent. In this particular case: "why is there an `else` here? isn't there something else that the next block  shouldn't do? is there a return?". I had to pause and check those because it wasn't clear to me. The `else:` here is redundant, and it doesn't make it any more clear.

anchore_engine/clients/grype_wrapper.py
alfredodeza: I understand the concerns, but I'll have to push back on the reasoning that this is out of scope. This PR is introducing `globals()`, and I'm not finding a reasonable explanation to why that is a requirement. 
I'm not implying that a complete new mechanism needs to exist to handle singletons or configuration values. There are many different ways to declare variables, keep state, and let that state be shared on all consumers _besides_ using `global()`.

This could be:

1. a top-level variable in a module
2. a read-only implementation of a dictionary (or other data structure), to use as a variable
3. a value that gets passed in from function to function
4. an environment variable

I would gladly welcome any code that breaks away from current production code that implements anti-patterns like `globals()` instead of any other option - including singletons, module-level vars, or any other mechanism similar to what I've suggested. 

This  type of addition is going to be a -1 for me.

dspalmer99:
@alfredodeza Why don't we move this to a Zoom call? There seems to be some disagreement about the scope of this PR, and I think it would be faster to resolve this with a f2f discussion.

alfredodeza:
After some offline discussion with @dspalmer99, it was clear that some sort of mutable variable is required because the destination directory keeps changing. Instead of using `global()` or a function that returns a variable that sets it to `global()` I believe that it is better to have a module level variable that has the same effect.

Regardless of that implementation, the underlying cause to force the use of a variable that keeps changing is because the destination directory is always different. I wasn't part of the discussion on how to implement/architect this, but I think that the global state tracking could be avoided by declaring well-known directory paths for the database destination.

All of my other comments were already addressed except for the problematic `globals()` which I still don't agree is a good path forward - but I'll defer to others closer to this functionality that may have a better understanding to continue with the review/approval/merge process.

dspalmer99:
@alfredodeza as per our discussion on Friday, the two most-apparent options to eliminate the use of globals are to either refactor this into a class and make those class variables, or as you said above to keep a consistent db dir name, and rename the new dir with that name, after removing the old (although this adds an additional point of potential failure). For this PR I have moved all references within the wrapper module to getters and setters (only the setters require those variables to use the global modifier). As this PR did not introduce those I do not want to expand the scope on that any further here. Our group working on this discussed this on Friday, and I am going to merge this today.

PR #993, opened 2021-04-27
Fixes error in malware gate execution causing evaluation error of whole policy

### LINKED ISSUE #992: Malware policy gate can fail causing error in policy eval when malware not enabled and other rules precede malware rule
When a malware scan rule is in a policy but has another rule preceding it that uses the execution context for data storage, there can be cases where the malware rule errors out due causing policy failure when it should just return an empty result.

zhill:
Fixes #992 by handling pre-existing context in the trigger executions when there are no scans for the image.


anchore_engine/services/policy_engine/engine/policy/gates/malware.py
```diff
@@ -50,11 +54,13 @@ class ScanNotRunTrigger(BaseTrigger):
...
     def evaluate(self, image_obj, context):
-        if not hasattr(context, "data"):
-            self._fire(msg="No malware scans found for image")
-            return
 
-        if len(context.data) == 0 or not context.data.items():
+        try:
+            scans = context.data.get(MALWARE_CONTEXT_KEY)
+        except AttributeError:
+            scans = None
```
alfredodeza, line 61: This is an excellent way of checking and falling-back from missing attributes

anchore_engine/services/policy_engine/engine/policy/gates/malware.py
```diff
@@ -80,8 +86,8 @@ def prepare_context(self, image_obj, context):
             AnalysisArtifact.analyzer_type == "base",
         ).all()
 
-        # Don't use a dict comprehension since context.data may have other content (not usually, but possible)
-        for scanner in scans:
-            context.data[scanner.artifact_key] = scanner.json_value
+        context.data[MALWARE_CONTEXT_KEY] = {
+            scanner.artifact_key: scanner.json_value for scanner in scans
+        }
```
alfredodeza, line 91: is `.data` going to always be available at this point? I thought there was a chance that this could be missing (hence the `AttributeError` check?

zhill: This is a good point, the old code was a bit redundant. The data property is part of the context object and should always be present. There is a bigger review/update to the gates/context code needed here to address it more systematically, but I can fix it in this one.

zhill:
I will update the code but also make this for 1.0/master branch merge since we missed the window for patch v0.9.4

zhill:
Updated merge base to master, but will re-base the branch as well.

PR #995, opened 2021-04-28
Vulnerabilities refactor

nightfurys:
This PR is the beginning of a refactor effort for all things vulnerabilities. It introduces a provider abstraction and refactors current logic spread out across policy-engine into legacy implementation. The abstraction provides a subsystem, a single landing spot for image vulnerabilities for now. There are place holders for other vulnerabilities related logic which follow in subsequent PRs


zhill (commented):
there are other aspects of the scanner that will need to be accounted for, such as the querying distro support, the distro mappings, and the sync process, but this is a step in the right direction for the scanning functionality itself.

PR #996, opened 2021-04-28
Add config and set timeout on site object

dakaneye:
https://anchore.atlassian.net/browse/ENTERPRISE-253


dspalmer99 (approved):

Toure (approved):
LGTM

zburstein:
Does a default need to be defined in `anchore_engine/configuration/localconfig.py`? Im not fully versed on the usage of the defaults in there so it might not actually be needed but just something I thought I would surface.

Vijay-P:
Is it worth writing tests for the timeout? Manual testing for this config change seems tedious and unnecessary.

dakaneye:
> Does a default need to be defined in anchore_engine/configuration/localconfig.py? Im not fully versed on the usage of the defaults in there so it might not actually be needed but just something I thought I would surface.

I left the default in the get method on the dict in the retrieval of the value

> Is it worth writing tests for the timeout? Is that in scope? Manual testing for this config change seems tedious and unnecessary.

@rbrady is working on an E2E test for this!

PR #1000, opened 2021-05-04
Vulnerabilities refactor - Part II

nightfurys:
This PR is part-II of the vulnerabilities refactor effort https://github.com/anchore/anchore-engine/issues/1001
- Gut the existing image vulnerabilities format. Create a richer report-like format that captures existing vulnerability results and new data. The new report format has placeholders for recording metadata for the report generator that may not be currently used by the legacy provider
- Decouple policy evaluation logic from vulnerabilities persistence model and streamlines the consumption of vulnerability results internally within policy engine. Updates policy evaluation vulnerabilities gate to operate on an instance of the new vulnerabilities report class. Gets rid of repetitive logic for os and non-os artifacts
- Update external API consumption (parsing) of image vulnerabilities from policy-engine
- Refactor (move) the get-vulnerabilities and get-images-by-vulnerability logic out of the controller and into the legacy provider

### DIFF
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -1290,6 +1290,7 @@ class VulnDBMetadata(Base):
 
         if self.vendor_cvss_v2:
             for cvss_v2_item in self.vendor_cvss_v2:
+                # create a new record for every single score as there could be a different number of v2 and v3 scores and its not clear which belong as a pair
                 results.append(
                     {
                         "id": self.name,
@@ -1314,6 +1315,7 @@ class VulnDBMetadata(Base):
 
         if self.vendor_cvss_v3:
             for cvss_v3_item in self.vendor_cvss_v3:
+                # create a new record for every single score as there could be a different number of v2 and v3 scores and its not clear which belong as a pair
                 results.append(
                     {
                         "id": self.name,
@@ -2599,14 +2601,14 @@ class ImagePackageVulnerability(Base):
 
         return fixed_in
 
-    def fixed_in(self):
+    def fixed_in(self, fixed_in: FixedArtifact = None):
         """
         Return the fixed_in version string value given a package matched (in case there are multiple packages specified in the vuln.
 
-        :param package: package to find a fix version for, if available
         :return: the fixed in version string if any or None if not found
         """
-        fixed_in = self.fixed_artifact()
+        if not fixed_in:
+            fixed_in = self.fixed_artifact()
         fix_available_in = (
             fixed_in.version if fixed_in and fixed_in.version != "None" else None
         )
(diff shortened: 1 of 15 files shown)


anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -1602,11 +1602,12 @@ def perform_vulnerability_scan(
...
-            vdiff = anchore_utils.process_cve_status(
-                old_cves_result=last_vuln_result["legacy_report"],
-                new_cves_result=curr_vuln_result["legacy_report"],
-            )
+        # TODO implement the differ correctly for the new format
+        # if last_vuln_result and curr_vuln_result:
+        #     vdiff = anchore_utils.process_cve_status(
+        #         old_cves_result=last_vuln_result["legacy_report"],
+        #         new_cves_result=curr_vuln_result["legacy_report"],
+        #     )
```
nightfurys, line 1610: Tracked by https://github.com/anchore/anchore-engine/issues/1002. Didn't want to block the merge on account of vuln_update notification

dspalmer99:
@nightfurys Looks good from just reading it, but I think this is sufficiently complex that an overview of it on Zoom would be helpful. I have some general questions about if the new model code (which is really nice) could be reused elsewhere.

nightfurys:
> @nightfurys Looks good from just reading it, but I think this is sufficiently complex that an overview of it on Zoom would be helpful. I have some general questions about if the new model code (which is really nice) could be reused elsewhere.

Thanks @dspalmer99, we can go over it in today's sync meeting.

anchore_engine/db/entities/policy_engine.py
```diff
@@ -2599,14 +2601,14 @@ def fixed_artifact(self):
 
         return fixed_in
 
-    def fixed_in(self):
+    def fixed_in(self, fixed_in: FixedArtifact = None):
```
zhill, line 2604: This code makes sense, but we should review later the inclusion of these functions in the db entity types. I think we're going to have to decouple them, but let's make a tech-debt issue for that and track that for all policy engine entities.

zhill: Shouldn't this formatting just be put on the FixedArtifact class to simplify the flow?

zhill (commented):
One question on code location, but overall makes sense.

anchore_engine/db/entities/policy_engine.py
nightfurys: I agree that this should not be a part of the ImagePackageVulnerability entity code but this will all go away when we get rid of those persistence models as we deprecate the legacy vulnerability matching in policy-engine. Addition of the new argument is to avoid computing the fixed artifact multiple times, I know it's hacky

PR #1003, opened 2021-05-04
If package type is not found, do nothing when applying hints at the e…

dakaneye:
This should allow hints to provide content for go and binary content types

### DIFF
--- a/anchore_engine/analyzers/manager.py
+++ b/anchore_engine/analyzers/manager.py
@@ -33,7 +33,8 @@ def apply_hints(analyzer_report, unpackdir):
         if pkg_type and pkg_type in syft.modules_by_engine_type:
             handler = syft.modules_by_engine_type[pkg_type]
             handler.save_entry(analyzer_report, engine_entry)
-
+        else:
+            logger.info("Current package type: %s is not processed by syft", pkg_type)
 
 def _run_analyzer_modules(analyzer_report, configdir, imageId, unpackdir, outputdir):
     for f in list_modules():
--- a/anchore_engine/analyzers/manager.py
+++ b/anchore_engine/analyzers/manager.py
@@ -36,6 +36,7 @@ def apply_hints(analyzer_report, unpackdir):
         else:
             logger.info("Current package type: %s is not processed by syft", pkg_type)
 
+
 def _run_analyzer_modules(analyzer_report, configdir, imageId, unpackdir, outputdir):
     for f in list_modules():
         cmdstr = " ".join([f, configdir, imageId, unpackdir, outputdir, unpackdir])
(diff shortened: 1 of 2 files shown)


anchore_engine/analyzers/manager.py
```diff
@@ -56,9 +57,8 @@ def _run_analyzer_modules(analyzer_report, configdir, imageId, unpackdir, output
                     )
                 else:
                     logger.debug(
-                        "command succeeded: cmd=%s stdout=%s stderr=%s".format(
-                            repr(cmdstr), repr(sout.strip()), repr(serr.strip())
-                        )
+                        "command succeeded: cmd=%s stdout=%s stderr=%s"
+                        % (repr(cmdstr), repr(sout.strip()), repr(serr.strip()))
```
alfredodeza, line 61: if you are changing this already, it is safer to skip the `%` altogether and use the rest as arguments (see https://github.com/anchore/enterprise/issues/150 for reference):

```suggestion
                        "command succeeded: cmd=%s stdout=%s stderr=%s",
                        repr(cmdstr), repr(sout.strip()), repr(serr.strip())
```

dakaneye: got it, fixed

Toure (approved):
LGTM

anchore_engine/analyzers/manager.py
```diff
@@ -47,18 +47,18 @@ def _run_analyzer_modules(analyzer_report, configdir, imageId, unpackdir, output
                 serr = anchore_engine.utils.ensure_str(serr)
                 if rc != 0:
                     logger.error(
-                        "command failed: cmd=%s exitcode=%s stdout=%s stderr=%s".format(
-                            repr(cmdstr),
-                            repr(rc),
-                            repr(sout.strip()),
-                            repr(serr.strip()),
-                        )
+                        "command failed: cmd=%s exitcode=%s stdout=%s stderr=%s",
```
zhill, line 50: Good fix. These needed updating anyway :)

anchore_engine/analyzers/manager.py
```diff
@@ -30,7 +30,7 @@ def apply_hints(analyzer_report, unpackdir):
     # hints processing.
     for engine_entry in utils.content_hints(unpackdir=unpackdir):
         pkg_type = engine_entry.get("type")
-        if pkg_type:
+        if pkg_type and pkg_type in syft.modules_by_engine_type:
```
zhill, line 33: I think an info message here would be good to indicate to the user that there are hints entry types not processed by the syft handler.

zhill (changes requested):
Super minor change to add a log line. Otherwise looks good.

anchore_engine/analyzers/manager.py
dakaneye: @zhill that might be a bit confusing. The syft handler is not applying hints for certain package types, but they _are_ getting applied by the analyzers themselves

Toure: @dakaneye I could move it to debug only?

dakaneye:
Duplicated by #1008

PR #1011, opened 2021-05-07
New provider, scanner and cache manager to interact with grype

nightfurys:
This PR builds on the refactor in https://github.com/anchore/anchore-engine/pull/1000  and introduces a new vulnerabilities provider for grype-integration. It also introduces
- a scanner that can generate a grype sbom from engine artifacts and transform grype output into engine report format
- a cache manager to load and retrieve vulnerability reports for images


nightfurys:
Commits in this PR are a subset of https://github.com/anchore/anchore-engine/pull/973. Merging this PR into the integration branch without an explicit approval

PR #1013, opened 2021-05-09
Temporary grype wrapper fixes

nightfurys:
- return the directory name for checksum 
- avoid removing grype active directory


nightfurys:
@dspalmer99, merging this PR to unblock @Vijay-P and @zburstein from testing feeds sync for grype. My fix for returning the checksum is definitely not correct, feel free update that as necessary

PR #1015, opened 2021-05-10
Link back to Syft and Grype from README

luhring:
Signed-off-by: Dan Luhring <dan.luhring@anchore.com>

**What this PR does / why we need it**: Adds links to Syft and Grype in the README (per [Kim W's suggestions](https://anchore.slack.com/archives/C01S5KSJ68G/p1620502250024300?thread_ts=1620318390.016400&cid=C01S5KSJ68G)) to help alert users to these tools' existence.

**Which issue this PR fixes**: n/a

**Special notes**: You'll see other changes to `README.md` than just the block about Syft and Grype... Prettier did that (i.e. autoformatting) for "consistency", but I can turn that off and remove those changes if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,21 @@ For the most up-to-date information on Anchore Engine, Anchore CLI, and other An
 
 The Anchore Engine is an open-source project that provides a centralized service for inspection, analysis, and certification of container images. The Anchore Engine is provided as a Docker container image that can be run standalone or within an orchestration platform such as Kubernetes, Docker Swarm, Rancher, Amazon ECS, and other container orchestration platforms.
 
+In addition, we also have several modular container tools that can be run standalone or integrated into automated workflows such as CI/CD pipelines.
+
+- **[Syft](https://github.com/anchore/syft)**: a CLI tool and library for **generating a Software Bill of Materials** (SBOM) from container images and filesystems
+
+- **[Grype](https://github.com/anchore/grype)**: a **vulnerability scanner** for container images and filesystems
+
 The Anchore Engine can be accessed directly through a RESTful API or via the Anchore [CLI](https://github.com/anchore/anchore-cli).
 
 With a deployment of Anchore Engine running in your environment, container images are downloaded and analyzed from Docker V2 compatible container registries and then evaluated against user-customizable policies to perform security, compliance, and best practices enforcement checks.
 
 Anchore Engine can be used in several ways:
 
-* Standalone or interactively.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,15 +4,21 @@ For the most up-to-date information on Anchore Engine, Anchore CLI, and other An
...
 The Anchore Engine can be accessed directly through a RESTful API or via the Anchore [CLI](https://github.com/anchore/anchore-cli).
 
 With a deployment of Anchore Engine running in your environment, container images are downloaded and analyzed from Docker V2 compatible container registries and then evaluated against user-customizable policies to perform security, compliance, and best practices enforcement checks.
 
 Anchore Engine can be used in several ways:
 
-* Standalone or interactively.
-* As a service integrated with your CI/CD to bring security/compliance/best-practice enforcement to your build pipeline
-* As a component integrated into existing container monitoring and control frameworks via integration with its RESTful API.
+- Standalone or interactively.
```
alfredodeza, line 19: Why is this necessary?

README.md
```diff
@@ -4,15 +4,21 @@ For the most up-to-date information on Anchore Engine, Anchore CLI, and other An
 
 The Anchore Engine is an open-source project that provides a centralized service for inspection, analysis, and certification of container images. The Anchore Engine is provided as a Docker container image that can be run standalone or within an orchestration platform such as Kubernetes, Docker Swarm, Rancher, Amazon ECS, and other container orchestration platforms.
 
+In addition, we also have several modular container tools that can be run standalone or integrated into automated workflows such as CI/CD pipelines.
```
alfredodeza, line 7: We may have several other tools, but this is listing syft and grype. This could generalize instead of quantifying, for example: 

> In addition, we have container tooling [...]

But I would go further and not emphasize "container" because these also work for filesystems - or this could mention both, and omit from repeating it in the descriptions below

alfredodeza: What would be an example of an _"integrated automated workflow"_ that isn't something related to a CI/CD service? I think this makes it wordy

luhring: @alfredodeza These suggestions come from @kimweins, so your questions may be better suited for her. Check out the linked Slack thread in the PR description if you haven't already.

My two cents:

- I'm up for tweaking this.
- For writing, I prefer the word "tools" over "tooling". While the latter is not wrong, it's a bit inflated and less clear than "tools" which works just as well here.
- Re: kinds of automated workflows, there are plenty of automated workflows that aren't CI/CD. Having said that, I agree that this might be a bit wordy, and we could just refer to CI/CD workflows directly here.

PR #1018, opened 2021-05-10
Re-add #988

dspalmer99:
It looks like an issue with a rebase of the feature branch this is targeting overwrote the changes added in https://github.com/anchore/anchore-engine/pull/988. This PR re-adds them.


zburstein:
Added a commit that udates grype sync manager to use the new update function rather than the one that was removed. Also changed the definition of this function to be public

dspalmer99:
@zburstein Thumbs up from me specifically on your commit, that looks good to me, and as per our slack discussion is needed.

Vijay-P (approved):

PR #1019, opened 2021-05-11
Vulnerabilities refactor continued

nightfurys:
This PR re-introduces deduping vulnerabilities that was skipped in the refactor PR #1000 One main difference with this addition is the where the dedup is performed. It was originally in external API and slightly differing variants sprinkled through out the system (policy eval, other services). This update fixes the issue by moving the dedup to get-image-vulnerabilities path - this guarantees every client gets the same information without the dedup know-how. Added some unit tests as well

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/dedup.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/dedup.py
@@ -1,0 +1,174 @@
+from dataclasses import dataclass, field
+from typing import List
+
+from anchore_engine.services.policy_engine.api.models import VulnerabilityMatch
+from anchore_engine.subsys import logger
+
+
+class FeedGroupRank:
+    """
+    Feed groups ranked by an integer value. Rank defaults to pre-defined value if the group is not not explicitly ranked
+
+    This is a very simplistic ranking strategy that handles three categories - nvdv2, github and all others.
+    The strategy translates to any-group is ranked > github >  nvdv2
+    """
+
+    __ranks__ = {"nvdv2": 1, "github": 10}
+    __default__ = 100
+
+    def get(self, feed_group: str):
+        group_prefix = feed_group.split(":", 1)[0]
+
+        return self.__ranks__.get(group_prefix, self.__default__)
+
+
+# eq=True and frozen=True required for making the instance hashable
+@dataclass(eq=True, frozen=True)
+class VulnerabilityIdentity:
+    vuln_id: str
+    pkg_name: str
+    pkg_version: str
+    pkg_type: str
+    pkg_path: str
+
+    @classmethod
+    def from_match(cls, vuln_match: VulnerabilityMatch):
+        """
+        Returns a list of identities from nvd references if available or a single identity with the vulnerability ID otherwise
+        """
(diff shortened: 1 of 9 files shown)


anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -59,29 +54,40 @@ class VulnerabilitiesProvider:
...
+    @abstractmethod
     def load_image(self, **kwargs):
         """
         Ingress the image and compute the vulnerability matches. To be used in the load image path to prime the matches
         """
-        raise NotImplementedError()
+        ...
+
+    @abstractmethod
+    def get_image_vulnerabilities_json(self, **kwargs) -> json:
```
zhill, line 65: I'm not sure 'json' is what you intended here. I think this is using the json module as the type.

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -100,14 +106,26 @@ def load_image(self, image: Image, db_session, cache=False):
...
+        self,
+        image: Image,
+        db_session,
+        vendor_only: bool = True,
+        force_refresh: bool = False,
+        cache: bool = True,
+    ) -> json:
+        return self.get_image_vulnerabilities(
+            image, db_session, vendor_only, force_refresh, cache
+        ).to_json()
```
zhill, line 119: This seems outside the scope of the provider: formatting the output object.

zhill (commented):
Overall the logic makes sense, I just have comments on how to simplify things a bit. The dedup logic may be simplified by using object hashing and identity to support set operations, but that is a nit, not blocking.

anchore_engine/services/policy_engine/engine/vulns/providers.py
nightfurys: That's incorrect, not sure how it made it there

anchore_engine/services/policy_engine/engine/vulns/providers.py
nightfurys: I had the jsonification outside of the provider scope initially but introduced it as an optimization for the grype-integration. When the report is loaded from cache, get_image_vulnerabilities_json() saves the additional steps to deserialize the json into object model and serialize it back again into json again. get_image_vulnerabilities() method will return the object model response, currently it's used by the policy evaluation gate to load vulnerabilities into the context

anchore_engine/services/policy_engine/engine/vulns/dedup.py
```diff
@@ -0,0 +1,158 @@
...
+            "vuln_id",
+            "vuln_namespace",
+            "pkg_name",
+            "pkg_version",
+            "pkg_type",
+            "pkg_path",
+            "rank",
+        ],
+    )
+    VulnerabilityRankMatchTuple = namedtuple(
```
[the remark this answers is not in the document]
nightfurys: Tuples play well with hashed collections like sets and make it easier for comparison. Dataclasses might be able to deliver the same functionality but might need overriding hash and eq functions

PR #1020, opened 2021-05-11
Grype feed group count

### LINKED ISSUE #1021: Return feed and group counts from grype wrapper
Add a function to grype_wrapper to return feed/group counts (along with a last updated timestamp) from vulnerabilities in the currently configured grype db.

dspalmer99:
This pr adds an additional function (along with a test) to return the feed and group vulnerability record counts from grype db.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #1021


anchore_engine/clients/grype_wrapper.py
```diff
@@ -455,3 +455,61 @@ def query_vulnerabilities(
...
+                output.append(record_source)
+
+            # Return the results
+            return output
+
+        finally:
+            read_lock.release()
+
+
+class RecordSource(object):
```
Vijay-P, line 510: this sort of boilerplate is a great candidate for dataclasses
https://docs.python.org/3/library/dataclasses.html

Vijay-P (approved):
LGTM

anchore_engine/clients/grype_wrapper.py
```diff
@@ -455,3 +455,61 @@ def query_vulnerabilities(
...
             read_lock.release()
+
+
+def query_record_source_counts():
+    """
+    Query the current feed group counts for all current vulnerabilities.
+    """
+    # Get and release read locks
+    read_lock = grype_db_lock.gen_rlock()
+    if read_lock.acquire(blocking=False, timeout=60):
```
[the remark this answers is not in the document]
nightfurys: Something like `with lock.acquire()` would make it easier to apply default values for timeouts and other attributes. Would also be helpful to keep the cleanup consistent later if we need to

anchore_engine/clients/grype_wrapper.py
```diff
@@ -455,3 +456,61 @@ def query_vulnerabilities(
...
+        try:
+            logger.debug("Querying grype_db for feed group counts")
+
+            # Get the counts for each record source
+            results = (
+                _get_grype_db_session()
```
nightfurys, line 481: db sessions should also be handled using a context manager. we do this currently for db sessions to our persistence context

anchore_engine/clients/grype_wrapper.py
```diff
@@ -455,3 +456,61 @@ def query_vulnerabilities(
...
+            last_synced = metadata["built"]
+
+            # Transform the results along with the last_synced timestamp for each result
+            output = []
+            for result in results:
+                feed_group = result[0].split(":", 1)
+                if len(feed_group) != 2:
+                    logger.error(
+                        "Unable to process feed/group for record_source {}. Omitting from the response".format()
+                    )
```
nightfurys, line 501: Missing argument to be plugged into log statement. Also switch to safe logging format
```suggestion
                    logger.error(
                        "Unable to process feed/group for record_source %s. Omitting from the response", feed_group
                    )

File: anchore_engine/clients/grype_wrapper.py [hunk 5]
```diff
@@ -455,3 +456,61 @@ def query_vulnerabilities(
...
+            )
+
+            # Get the timestamp from the current metadata file
+            metadata = get_current_grype_db_metadata()
+            last_synced = metadata["built"]
+
+            # Transform the results along with the last_synced timestamp for each result
+            output = []
+            for result in results:
+                feed_group = result[0].split(":", 1)
```
nightfurys, line 497: What does result[0] contain out of curiosity

nightfurys (commented):
Overall looks good, I know you have some class-fying work landing here pretty soon. Just have a few questions and some general feedback

anchore_engine/clients/grype_wrapper.py
dspalmer99: Thanks, will fix the missing argument. In the next PR I've already switched the logging format, so I'd rather just leave that for there though.

anchore_engine/clients/grype_wrapper.py
dspalmer99: @nightfurys One the first point, I looked into doing that when I first pulled in that library, and due to the way the authors did the inheritance structure, you can't parameterize the the acquire call via a context manager through call to the __enter__(). I'd need to add something extra (Vijay also suggested that) to wrap the calls to the lock in order to use a context manager with it.

On the second point, should I be closing and recreating the session object on every query then? I.e. the SessionMaker is what gets created and persisted when the grypeDB changes? I was thinking since we are using grype db as a read-only db it was ok to leave that open (since the underlying data being queried won't change), but it occurs to me now that there's probably a performance cost to leaving that open.

anchore_engine/clients/grype_wrapper.py
dspalmer99: GrypeVulnerability.record_source - The query above just returns a tuple, I didn't see a way to get it to return named fields. Is that possible when you are pulling back fields instead of the object?

dspalmer99: Probably I should add a little extra protection here in the event that GrypeVulnerability.record_source doesn't return a string.

dspalmer99: I've updated this.

anchore_engine/clients/grype_wrapper.py
dspalmer99: If that is the case I'm happy to make that change, but I think I'd rather do that in [https://github.com/anchore/anchore-engine/pull/1029](https://github.com/anchore/anchore-engine/pull/1029) too. I'm already changing how that variable is handled there, so it will be easier to change what its storing there too (rather than doing it here and then having a messier rebase of that branch on this one).

dspalmer99: @nightfurys I thought more about what you and @Vijay-P were saying here about context managers for the lock and the db session. I've got implementations of each I am testing locally, but I'm going to target them for #1029 rather than this PR.

dspalmer99:
@nightfurys @Vijay-P I've addressed all feedback on this, either here or in https://github.com/anchore/anchore-engine/pull/1029 as made sense merge-ease-wise. (My plan is to merge this one into the feature branch first, then rebase 1029 on that and class-ify anything new added here). Are you good for me to merge this one now?

nightfurys:
Thanks for being patient with the feedback @dspalmer99! Merging this into integration branch, if there's any remaining work a new PR can be opened

PR #1024, opened 2021-05-12
merge v0.9.4-dev to master

dakaneye:


zhill (approved):
lgtm

anchore_engine/analyzers/manager.py
```diff
@@ -29,10 +30,24 @@ def apply_hints(analyzer_report, unpackdir):
...
+        pkg_type = pkg_type.lower() if pkg_type else None
+        if (
+            pkg_type
+            and pkg_type in syft.modules_by_engine_type
+            and pkg_type in hints.hints_by_type
+        ):
             handler = syft.modules_by_engine_type[pkg_type]
-            handler.save_entry(analyzer_report, engine_entry)
+            try:
+                hint = hints.hints_by_type[pkg_type](engine_entry).to_dict()
```
Toure, line 44: If we fail on fail on processing the hints, shouldn't we bail instead of walking the remaining entries?

Toure (changes requested):

anchore_engine/analyzers/manager.py
Toure: Other than that question above it looks good.

dakaneye: It's a single hint, we don't want one hint to cause others to fail. Also, this is just a PR to sync the branches, so wouldn't be fixing any of the underlying fixes here. If you wanna discuss in more detail about the hints let me know!

Toure (approved):

PR #1027, opened 2021-05-13
Improve the docker image info parsing flow

dakaneye:
Now, uses DockerImageReference and refines the tag parsing to its own class with unit tests

Root Cause:
* (Theory, but not sure): Addresses an upstream (enterprise) issue where the digest is getting reset even when an image input string is specified

Other:
* The parse method was long and hard to read and understand all the input cases and expected output cases
* The parse method was hard to test as it was

**Special notes**:
* Relates to Watch Inventory enterprise feature


zhill:
@dakaneye is this still needed to fix an issue or just tech-debt cleanup and testing now?

zhill:
@dakaneye I resolved some conflicts and updated for linters to see if we can get this merged.

anchore_engine/util/docker.py
```diff
@@ -153,6 +155,110 @@ def parse_dockerimage_string(instr, strict=True):
     return ret
 
 
+class DockerImageTag:
```
zhill, line 158: Since this supports tag and digest as the last component, I think a more descriptive name would be "DockerPullReference" so that it's clear this supports the various ways docker can pull the image as opposed to a raw digest or ImageID (which is handled by the DockerImageReference object).

zhill (changes requested):
Looks good, just a naming update request to make the purpose of the new class a bit more clear. I'm happy to make that update if needed.

PR #1014, opened 2021-05-10
[DO NOT MERGE] feeds hackery

nightfurys:
- adds a new configuration section for vulnerabilities provider and associated data feeds
- update feed sync process to download the grype-db from the open source url by default (configurable)
- minor fixes to grype_wrapper for returning the checksums and not deleting active db directory 
- circle ci change to push the container images from this branch to dockerhub

PR #1028, opened 2021-05-13
Grype DB Download

Vijay-P:
This PR updates the download logic in the feeds sync task to also support downloading Grype DB from the toolbox service (or corresponding enterprise API).

**Special notes**:
* Logic implemented corresponds directly to changes in https://github.com/anchore/anchore-engine/pull/1014

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,7 +132,8 @@ ENV ANCHORE_CONFIG_DIR=/config \
     ANCHORE_AUTH_ENABLE_HASHED_PASSWORDS=false \
     AUTHLIB_INSECURE_TRANSPORT=true \
     ANCHORE_MAX_COMPRESSED_IMAGE_SIZE_MB=-1 \
-    ANCHORE_GLOBAL_SERVER_REQUEST_TIMEOUT_SEC=180
+    ANCHORE_GLOBAL_SERVER_REQUEST_TIMEOUT_SEC=180 \
+    ANCHORE_GRYPE_DB_URL="https://toolbox-data.anchore.io/grype/databases/listing.json"
 
 # Insecure transport required in case for things like tls sidecars
 
--- a/anchore_engine/services/policy_engine/engine/feeds/client.py
+++ b/anchore_engine/services/policy_engine/engine/feeds/client.py
@@ -4,7 +4,7 @@ import datetime
 import json
 from dataclasses import dataclass, field
 from io import BytesIO
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 
 import ijson
 import requests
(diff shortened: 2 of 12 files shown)


anchore_engine/services/policy_engine/engine/feeds/sync.py
```diff
@@ -120,6 +120,72 @@ def get_selected_feeds_to_sync(config):
...
+                    "token_url": "https://ancho.re/oauth/token",
+                },
+            },
+        },
+    }
+
+    Returns [grypedb]
+
+    """
+    logger.debug("Searching for data feeds to be synced in {}".format(provider_config))
```
zburstein, line 177: Should the logging use the percent format that @nightfurys  sent us: 

```
logger.debug("Searching for data feeds to be synced in %s" % provider_config)
```

nightfurys: I think this was a copy paste of my own code and I still tend to use .format() out of habit. We can clear these up before the merge into master

anchore_engine/services/policy_engine/engine/feeds/client.py
```diff
@@ -333,193 +589,66 @@ def get_client(
...
-
-    def __repr__(self):
-        return "{} - exception: {}".format(self.msg, self.cause)
+    unauthenticated_http_client = HTTPBasicAuthClient(
+        username=None,
+        password=None,
+        connect_timeout=conn_timeout,
+        read_timeout=read_timeout,
+        verify=verify,
+    )
```
nightfurys, line 649: reordering of ops might be benefit this block. Fetch feeds_config inside the outer if block. grype_db_url could still be null dereferenced

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -119,7 +119,10 @@ def _check_sync_necessary(cls):
         local_grypedb_checksum = cls._get_local_grypedb_checksum()
 
         sync_needed = local_grypedb_checksum != active_grypedb.checksum
-        logger.info("No grypedb sync needed at this time")
+        if not sync_needed:
+            logger.info("No Grype DB sync needed at this time")
+        else:
+            logger.info("Grype DB sync is required.")
 
         return cls.SyncNecessaryResp(sync_needed, active_grypedb)
```
nightfurys, line 127: Not related to this commit but this violates separation of concerns. If the intent is to check whether a sync is needed, the method should just return a boolean. I see that it's an optmization but the intent does not match the response.

conf/default_config.yaml
```diff
@@ -188,6 +188,30 @@ services:
...
+            anonymous_user_password: pbiU2RYZ2XrmYQ
+            url: ${ANCHORE_FEEDS_URL}
+            client_url: ${ANCHORE_FEEDS_CLIENT_URL}
+            token_url: ${ANCHORE_FEEDS_TOKEN_URL}
+          github: # remove after tests are updated
+            enabled: true # remove after tests are updated
+          vulnerabilities: # remove after tests are updated
+            enabled: true # remove after tests are updated
+          nvdv2: # remove after tests are updated
+            enabled: true # remove after tests are updated
```
nightfurys, line 215: The last 3 feeds shouldn't be here if the provider is grypedb. they are all included a part of grypedb.

Dockerfile
```diff
@@ -132,6 +132,7 @@ ENV ANCHORE_CONFIG_DIR=/config \
     ANCHORE_AUTH_ENABLE_HASHED_PASSWORDS=false \
     AUTHLIB_INSECURE_TRANSPORT=true \
     ANCHORE_MAX_COMPRESSED_IMAGE_SIZE_MB=-1 \
+    ANCHORE_GRYPE_DB_URL="https://toolbox-data.anchore.io/grype/databases/listing.json" \
```
nightfurys, line 135: (nit) add new variables at the end of the current list, makes it a tiny bit easy for folks to find new additions

nightfurys (commented):
The flow looks good overall. I have some suggestions about around feed sync

conf/default_config.yaml
Vijay-P: it's just there to get the tests to pass until zane gets to updating them, hence the comments

Dockerfile
Vijay-P: yeah got reordered during merge conflict resolution

anchore_engine/services/policy_engine/engine/feeds/client.py
```diff
@@ -247,7 +423,87 @@ def _extract_response_data(self, response_text):
...
+        self,
+    ) -> Tuple[Dict, HTTPClientResponse]:
+        logger.info("Downloading grypedb listing.json from {}".format(self.feed_url))
+        listing_response = self.http_client.execute_request(
+            requests.get, self.feed_url, retries=self.RETRY_COUNT
+        )
+        if not listing_response.success:
+            raise HTTPStatusException(listing_response)
+        listings_json = json.loads(listing_response.content.decode("utf-8"))
+        db_listing = listings_json.get("available").get("2")[0]
```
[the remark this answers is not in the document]
Vijay-P: @wagoodman @alfredodeza is this something that could be updated in grype?

anchore_engine/services/policy_engine/engine/feeds/client.py
```diff
@@ -247,7 +423,87 @@ def _extract_response_data(self, response_text):
...
+        return FeedGroupList(
+            groups=[
+                FeedAPIGroupRecord(
+                    name="grypedb:vulnerabilities",
+                    description="grypedb:vulnerabilities group",
+                    access_tier="0",
+                )
+            ]
+        )
+
```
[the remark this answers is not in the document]
Vijay-P: Isn't the goal to have feed and feed group information drawn from Grype itself? @zhill brought that up during a short discussion. The policy engine endpoints for feed and feed group information relies on FeedMetadata and FeedGroupMetadata DB rows populated in anchore_engine.services.policy_engine.engine.feeds.sync.DataFeeds.sync_metadata(). That method calls these two methods that we're talking about here, anchore_engine.services.policy_engine.engine.feeds.client.GrypeDBServiceClient.list_feeds() and list_feed_groups(). I'm not sure how to implement this without breaking the legacy flow.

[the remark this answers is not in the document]
Vijay-P: We could make the api controller at anchore_engine.services.policy_engine.api.controllers.feeds.list_feeds() read the config and generate FeedGroupMetadata and FeedMetadata records using the grype wrapper, and then glue those records together with the legacy ones in the database. This would let us surface grype info without duplicating data from Grype DB in Anchore DB feeds tables.

Dockerfile
Vijay-P: fixed

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
Vijay-P: fixed

anchore_engine/services/policy_engine/engine/feeds/client.py
Vijay-P: fixed

anchore_engine/services/policy_engine/engine/feeds/client.py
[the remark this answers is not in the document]
luhring: I'm sure I'm missing context here. But why should the schema version be configurable? Is the Grype version configurable?

anchore_engine/services/policy_engine/engine/feeds/sync.py
alfredodeza: I'm not sure why we there is still a mix of various different formats for logging being discussed. Using `%` or `.format()` should not be preferred *for logging purposes* as [described here](https://github.com/anchore/enterprise/issues/150). To reiterate a bit the message:

> It is safer, and recommended, to not format string directly like that because it can cause applications to break when variables can be substituted. Logging offers the following solution:

```python
logger.info('some statement about %s', 'app')
```

> If string substitution doesn't work, the logging module will _log_ an error, showing what failed, preventing application breakage.

anchore_engine/services/policy_engine/engine/feeds/client.py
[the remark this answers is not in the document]
nightfurys: Not user configurable. The download logic should take the schema version as an input rather than hardcoding it. An enhancement would be asking grype for required db version so the download downloads the correct database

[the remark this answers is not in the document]
luhring: The required schema version is exposed by the `grype version` command. Is the problem the output format?

```
$ grype version
Application:          grype
Version:              0.11.0
BuildDate:            2021-04-22T21:51:14Z
GitCommit:            0a46e54aae837135f73d2ff18a84061189ee6f28
GitTreeState:         clean
Platform:             darwin/amd64
GoVersion:            go1.16.3
Compiler:             gc
Supported DB Schema:  2
```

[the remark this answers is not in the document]
Vijay-P: thanks! we didn't know that. that's what we needed

anchore_engine/services/policy_engine/engine/feeds/sync.py
zhill: This is important to understand :point_up:

anchore_engine/services/policy_engine/engine/feeds/sync.py
```diff
@@ -120,6 +120,72 @@ def get_selected_feeds_to_sync(config):
...
+                    "enabled": True,
+                    "url": "https://toolbox-data.anchore.io/grype/databases/listing.json",
+                },
+                "packages": {
+                    "enabled": False,
+                    "anonymous_user_username": "anon@ancho.re",
+                    "anonymous_user_password": "pbiU2RYZ2XrmYQ",
+                    "url": "https://ancho.re/v1/service/feeds",
+                    "client_url": "https://ancho.re/v1/account/users",
+                    "token_url": "https://ancho.re/oauth/token",
```
zhill, line 168: Only the url field is required. The others are all unused legacy fields.

anchore_engine/services/policy_engine/engine/feeds/sync.py
nightfurys: @alfredodeza there's no discussion about multiple formats. I shared the safe logging format with the grype-integration team but accidentally left the % operator for substitution instead of separate arguments. Thank you for explaining the format again!

anchore_engine/services/policy_engine/engine/feeds/sync.py
Vijay-P: thanks! will adjust in following work

anchore_engine/services/policy_engine/engine/feeds/client.py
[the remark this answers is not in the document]
Vijay-P: https://github.com/anchore/anchore-engine/issues/1036

PR #1029, opened 2021-05-13
Grype wrapper class refactor

dspalmer99:
Update grype wrapper into a singleton class for better encapsulation and null checking of its state variables.

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -82,18 +82,18 @@ class GrypeWrapperSingleton(object):
         return cls._grype_wrapper_instance
 
     @classmethod
-    def get_instance(self):
+    def get_instance(cls):
         """
         Returns the singleton instance of this class.
         """
         return GrypeWrapperSingleton()
 
     @classmethod
-    def _get_test_instance(self):
+    def _get_test_instance(cls):
         """
         Returns a new instance of this class. This method is not intended for use outside of tests.
         """
-        self._grype_wrapper_instance = None
+        cls._grype_wrapper_instance = None
         return GrypeWrapperSingleton()
 
     @property
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -75,11 +75,13 @@ class GrypeWrapperSingleton(object):
     )
 
     def __new__(cls):
+        # If the singleton has not been initialized yet, do so with the instance variables below
         if cls._grype_wrapper_instance is None:
             cls._grype_wrapper_instance = super(GrypeWrapperSingleton, cls).__new__(cls)
             cls._grype_db_dir_internal = None
             cls._grype_db_session_maker_internal = None
             cls._grype_db_lock = rwlock.RWLockWrite()
+        # Return the singleton instance
         return cls._grype_wrapper_instance
 
     @classmethod
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -63,6 +63,7 @@ class RecordSource:
 class GrypeWrapperSingleton(object):
     _grype_wrapper_instance = None
 
+    # These values should be treated as constants, and will not be changed by the functions below
     LOCK_READ_ACCESS_TIMEOUT = 60
     LOCK_WRITE_ACCESS_TIMEOUT = 60
     GRYPE_SUB_CMD = "grype -vv -o json"
@@ -79,10 +80,16 @@ class GrypeWrapperSingleton(object):
     def __new__(cls):
         # If the singleton has not been initialized yet, do so with the instance variables below
         if cls._grype_wrapper_instance is None:
+            # The singleton instance, only instantiated once outside of testing
             cls._grype_wrapper_instance = super(GrypeWrapperSingleton, cls).__new__(cls)
+
+            # These variables are mutable, their state can be changed when grype_db is updated
             cls._grype_db_dir_internal = None
             cls._grype_db_session_maker_internal = None
+
+            # The reader-writer lock for this class
             cls._grype_db_lock = rwlock.RWLockWrite()
+
         # Return the singleton instance
         return cls._grype_wrapper_instance
 
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -63,6 +63,8 @@ class RecordSource:
 class GrypeWrapperSingleton(object):
     _grype_wrapper_instance = None
 
+    LOCK_READ_ACCESS_TIMEOUT = 60
+    LOCK_WRITE_ACCESS_TIMEOUT = 60
     GRYPE_SUB_CMD = "grype -vv -o json"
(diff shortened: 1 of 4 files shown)


dspalmer99:
Commit history on the target branch has changed. Fixing this branch to match.

dspalmer99:
I've fixed the commit history on this branch now.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -16,442 +16,467 @@
...
+    def __new__(cls):
+        if cls._grype_wrapper_instance is None:
+            cls._grype_wrapper_instance = super(GrypeWrapperSingleton, cls).__new__(cls)
+            cls._grype_db_dir_internal = None
+            cls._grype_db_session_internal = None
+            cls._grype_db_lock = rwlock.RWLockWrite()
+        return cls._grype_wrapper_instance
+
+    @classmethod
+    def get_instance(self):
```
Vijay-P, line 75: first argument for a classmethod should be cls, not self

anchore_engine/clients/grype_wrapper.py
```diff
@@ -16,442 +16,467 @@
...
+        """
+        Returns the singleton instance of this class.
+        """
+        return GrypeWrapperSingleton()
+
+    @classmethod
+    def _get_test_instance(self):
```
Vijay-P, line 82: same thing here (cls)

anchore_engine/clients/grype_wrapper.py
Vijay-P, line 75: not super clear as to why this method is necessary, but it doesn't do any harm either.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -16,442 +16,467 @@
...
-    Remove old the local grype db file
-    """
-    if os.path.exists(grype_db_dir):
-        logger.info("Removing old grype_db at {}".format(grype_db_dir))
-        shutil.rmtree(grype_db_dir)
-    else:
-        logger.error(
-            "Failed to remove grype db at {} as it cannot be found.".format(
-                grype_db_dir
+    def __new__(cls):
```
Vijay-P, line 66: nice

anchore_engine/clients/grype_wrapper.py
```diff
@@ -16,442 +16,467 @@
...
+    description = Column(String)
+    cvss_v2 = Column(String)
+    cvss_v3 = Column(String)
 
-    # Return the full path to the grype_db dir
-    logger.info("Returning the unpacked grype_db dir at {}".format(output_dir))
-    return output_dir
 
+class GrypeWrapperSingleton(object):
+    _grype_wrapper_instance = None
```
Vijay-P, line 53: I see that there are other class variables initialized in the __new__ method. Would it enhance readability to also declare them here?

anchore_engine/clients/grype_wrapper.py
```diff
@@ -16,442 +16,467 @@
...
+            return self._grype_db_dir_internal
+
+    @_grype_db_dir.setter
+    def _grype_db_dir(self, _grype_db_dir_internal):
+        self._grype_db_dir_internal = _grype_db_dir_internal
+
+    @property
+    def _grype_db_session(self):
+        if self._grype_db_session_internal is None:
+            raise ValueError(self.MISSING_GRYPE_DB_SESSION_ERROR_MESSAGE)
```
Vijay-P, line 103: Another pattern you might consider is using the None type check to initialize the variable rather than raising a ValueError. So you would potentially call _move_and_open_grype_db_archive, but hiding that behind a getter might be more confusing than not.

Vijay-P (changes requested):

anchore_engine/clients/grype_wrapper.py
Vijay-P, line 82: might be cleaner to introduce a subclass with only this method just for tests.

Vijay-P:
Overall, I'm really liking the switch to singleton. Nice work!

anchore_engine/clients/grype_wrapper.py
dspalmer99: Thanks, will fix.

dspalmer99: Mostly just for code clarity, to make it very obvious to the caller what they are getting (vs the constructor)

anchore_engine/clients/grype_wrapper.py
dspalmer99: I was trying to make clear to the reader that those are instance variables. Maybe a comment on them?

Vijay-P: I see what you intend. It's sort of funny because the way that singletons are implemented, there are no instances, only the class. The pattern you are using takes advantage of the fact that there is only ever one instance of a class variable. To avoid accidental re-initialization of instance variables on calls to the constructor, you are not defining an `__init__` method and not using instance variables. So really, these are not instance variables, and making them appear as such may be more confusing than not.

anchore_engine/clients/grype_wrapper.py
dspalmer99: I'd prefer to keep this a dumb client - the caller is responsible for the logic for kicking off the reinitialization og grype_db if this is called before those variables are set.

anchore_engine/clients/grype_wrapper.py
dspalmer99: Sorry, I don't agree. Those other variables would essentially be constants if python supported such a thing. These are mutable.

anchore_engine/clients/grype_wrapper.py
dspalmer99: Done, great idea.

anchore_engine/clients/grype_wrapper.py
Vijay-P: Sorry, let me clarify. I think that calling them instance variables is confusing. However, I think that your implementation here is fine. I also notice now that you have added more constant style class variables that were not there before. The logical separation feels more apparent than it did during my first review.

Vijay-P: That's more of a nitpick.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -10,322 +10,433 @@
...
+            raise ValueError(self.MISSING_GRYPE_DB_SESSION_MAKER_ERROR_MESSAGE)
+        else:
+            return self._grype_db_session_maker_internal
+
+    @_grype_db_session_maker.setter
+    def _grype_db_session_maker(self, _grype_db_session_maker_internal):
+        self._grype_db_session_maker_internal = _grype_db_session_maker_internal
+
+    @contextmanager
+    def read_lock_access(self):
```
Vijay-P, line 119: This is a nice enhancement. If you had implemented these as a class, it would give you the option of parametrizing the context manager through the constructor for the class, which would mean that you could modify the timeouts for different situations. But if you don't need such a granular level of control, then this is fine.

Vijay-P (approved):
Awesome improvement to the grype-wrapper overall. Glad you spent time working on this.

PR #1033, opened 2021-05-14
Restructure mappers for transforming data between engine and grype

nightfurys:
Improves the oo model for mappers for better testability. Still has a bunch of todos waiting on grype data. Unit tests to follow

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -16,38 +16,452 @@ from anchore_engine.services.policy_engine.api.models import (
 )
 
 
+class DistroMapper:
+    engine_distro = None
+    grype_os = None
+    grype_like_os = None
+
+    @staticmethod
+    def for_engine(distro: str):
+        return ENGINE_DISTRO_MAPPERS.get(distro)
+
+    @staticmethod
+    def for_grype(distro: str):
+        return GRYPE_DISTRO_MAPPERS.get(distro)
+
+    def to_grype_distro(self, version):
+        return {
+            "name": self.grype_os,
+            "version": version,
+            "idLike": self.grype_like_os,
+        }
+
+
+class RhelMapper(DistroMapper):
+    engine_distro = "rhel"
+    grype_os = "redhat"
+    grype_like_os = "fedora"
+
+
+class DebianMapper(DistroMapper):
+    engine_distro = "debian"
+    grype_os = "debian"
+    grype_like_os = "debian"
+
+
+class UbuntuMapper(DistroMapper):
+    engine_distro = "ubuntu"
+    grype_os = "ubuntu"
+    grype_like_os = "debian"
+
+
+class AlpineMapper(DistroMapper):
+    engine_distro = "alpine"
+    grype_os = "alpine"
+    grype_like_os = "alpine"
+
+
+class CentOSMapper(DistroMapper):
+    engine_distro = "centos"
+    grype_os = "centos"
+    grype_like_os = "fedora"  # TODO check with toolbox
+
+
+class OracleLinuxMapper(DistroMapper):
+    engine_distro = "ol"
+    engine_like_distro = "fedora"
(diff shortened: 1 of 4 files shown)


anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -157,180 +524,29 @@ def transform_matches_to_vulnerabilities(self, grype_response):
...
-            elif vuln_id.startswith("VULNDB"):  # this is absolutely WRONG!!!
-                feed = "vulndb"
-                feed_group = "vulndb:vulnerabilities"
-            else:
-                feed = "nvdv2"
-                feed_group = "nvdv2:cves"
-
-            vuln_tuple = (vuln_id, feed_group, pkg_name, pkg_version, pkg_path)
-            if vuln_tuple in unique_results:
+            pkg_mapper_class = GRYPE_PACKAGE_MAPPERS.get(artifact.get("type"))
```
zburstein, line 532: Will artifact ever be `None`? If so the get on line 532 could throw an error

zburstein (commented):
Not sure I can review to the specifics of how each mapping is expected to look. Would be good to also get some unit tests in for these mappers but I know that will come later.

anchore_engine/services/policy_engine/engine/vulns/mappers.py
nightfurys: If an attribute is not present in a dictionary, `.get()` will simply return `None` Further ops on null dereferenced variable will raise errors. And yes, I do agree this needs a lot of unit testing and it's still waiting on grype updates

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -16,38 +16,452 @@
...
+
+    def to_grype_distro(self, version):
+        return {
+            "name": self.grype_os,
+            "version": version,
+            "idLike": self.grype_like_os,
+        }
+
+
+class RhelMapper(DistroMapper):
```
zhill, line 40: These don't need to be classes if they don't have different logic. They should be the same type, just with parameters to the constructor. A child class expects to have different behavior. The state of an object is expected to differ per instance.

An alternative, and cleaner approach is something like:
mappers = {
"rhel": DistroMapper(engine_distro="rhel", grype_os="redhat", grype_like_os="fedora),
"debian": ...
}

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -56,13 +470,27 @@ def transform_image_to_sbom(
...
+        if not distro_mapper_class:
+            log.error(
+                "No distro mapper found for %s. Cannot generate sbom", image.distro_name
+            )
+            raise ValueError(
+                "No distro mapper found for {}. Cannot generate sbom".format(
+                    image.distro_name
+                )
+            )
+        distro_mapper = distro_mapper_class()
```
zhill, line 483: An approach with the lookup map including instances instead of classes makes this simpler and faster too, no need to instantiate over and over unless the instances behave differently.

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -16,38 +16,452 @@
...
+                    ]
+                )
+            )
+            for item in location_cpes_dict.get(image_package.pkg_path)
+        ]
+
+        return artifact
+
+
+class PythonMapper(CPEMapper):
```
zhill, line 371: Same comment here. These all seem to be the same class, just with different input values. And I'm not clear on why the to_grype_artifact() method is overridden in each subclass since it looks to always be the same and always a call to super() for the same method.

anchore_engine/services/policy_engine/engine/vulns/mappers.py
nightfurys: @zhill between rhel debian and apkg, mappers will have different  will have different logic for mapping source package to grype's artifact metadata

PR #1037, opened 2021-05-18
Provider configuration

nightfurys:
This PR 
- updates feed configuration - remove old config and replace it with `vulnerabilities` section nested under policy_engine
- refactor config loading and defaults to a new file so that it can be reused and tested independently of sync
- updated input to feed sync task to use objects that contain sync configuration instead of a list of names. reduces the config lookup from different places and streamlines the process a bit more 
- updated grype sync handler to check for the provider before executing the sync task


anchore_engine/services/policy_engine/engine/feeds/config.py
```diff
@@ -0,0 +1,284 @@
...
+        logger.warn(
+            "%s is not a supported provider. Defaulting to %s",
+            provider,
+            DEFAULT_VULNERABILITIES_PROVIDER,
+        )
+        provider = DEFAULT_VULNERABILITIES_PROVIDER
+
+    connection_timeout_seconds = sync_config.get("connection_timeout_seconds", 3)
+    read_timeout_seconds = sync_config.get("read_timeout_seconds", 60)
+
```
Vijay-P, line 144: I think these numbers should be defined at the top of the file as constants.

Vijay-P: also would be nice if 142 and 142 were moved up to between 131 and 132, keeping the provider logic together

anchore_engine/services/policy_engine/engine/feeds/config.py
```diff
@@ -0,0 +1,284 @@
...
+            provider,
+            data_config,
+        )
+        to_be_synced = dict()
+        for feed_name, sync_config_dict in data_config.items():
+            # get the default config for both verifying feed_name is a valid data feed and accessing url if necessary
+            default_sync_config = DEFAULT_PROVIDER_SYNC_CONFIG_MAP.get(provider).get(
+                feed_name
+            )
+            if default_sync_config:
```
Vijay-P, line 157: because of the earlier block defaulting the provider, this is guaranteed to not be None. Also would reduce nesting if you checked the negative (! default_sync_config) and unindented the true case.

anchore_engine/services/policy_engine/engine/feeds/config.py
```diff
@@ -0,0 +1,284 @@
...
+                logger.warn(
+                    "%s data feed is unrecognized for %s provider, skipping",
+                    feed_name,
+                    provider,
+                )
+                continue
+
+    else:
+        logger.debug(
+            "Configuration found for data feeds not found, falling back to %s provider defaults",
```
Vijay-P, line 180: Remove first instance of "found" in this log message.

anchore_engine/services/policy_engine/engine/feeds/config.py
```diff
@@ -0,0 +1,284 @@
...
+          }
+       }
+    }
+    """
+    if not vulnerabilities_config:
+        return DEFAULT_VULNERABILITIES_PROVIDER
+
+    try:
+        provider = SupportedProvider(vulnerabilities_config.get("provider"))
+        provider = provider.value
```
Vijay-P, line 275: i feel like this defeats the purpose of using an enum. String comparisons are brittle, so you should return the enum and check against the enum in the caller.

anchore_engine/services/policy_engine/engine/feeds/config.py
```diff
@@ -0,0 +1,284 @@
...
+             }
+          }
+       }
+    }
+    :return: True or False
+    """
+    if not vulnerabilities_config:
+        return True
+
+    sync_enabled = vulnerabilities_config.get("sync", {}).get("enabled")
```
Vijay-P, line 234: can't you use `.get("enabled", True)` here and get rid of the if/else?

anchore_engine/services/policy_engine/__init__.py
```diff
@@ -376,15 +330,22 @@ def handle_grypedb_sync(*args, **kwargs):
...
-        try:
-            result = GrypeDBSyncTask().execute()
-
-            if result:
-                logger.info("Grype DB synced to local instance via handler")
-        #         TODO narrow scope of exceptions in handlers. see https://github.com/anchore/anchore-engine/issues/1005
-        except Exception:
-            logger.exception(
-                "Error encountered when running GrypeDBSyncTask from async monitor"
+        provider = get_provider(get_vulnerabilities_config())
```
zburstein, line 333: I had started doing some work on this and I decided to put this logic in the task. While not needed now, I thought it might be nice to have a handler for the local sync that checks the configuration. If a process wants to check the config before doing the sync, then it could call the task but if it wants to skip that it can just call the sync manager class directly. Thoughts?

anchore_engine/services/policy_engine/__init__.py
```diff
@@ -376,15 +330,22 @@ def handle_grypedb_sync(*args, **kwargs):
...
+        if provider.lower() == "grype":
```
zburstein, line 334: Should this use the `SupportedProvider` enum from the config file to verify the provider as opposed to just a string check? Makes it easier to update in the future and more clear where these checks happen

anchore_engine/services/policy_engine/engine/feeds/config.py
nightfurys: That wouldn't work for the case if `enabled=""` Empty string throws off the get with defaults. If sync_enabled is blank, it would technically evaluate to False without the if-else check

anchore_engine/services/policy_engine/engine/tasks.py
```diff
@@ -228,22 +233,36 @@ def execute(self) -> List[FeedSyncResult]:
...
 
             updated_data_feeds = list()
-            if "grypedb" in self.feeds:
-                updated_data_feeds += DataFeeds.sync(
-                    get_grype_db_client(),
-                    to_sync=["grypedb"],
-                    full_flush=self.full_flush,
-                    catalog_client=catalog_client,
-                    operation_id=self.uuid,
+            grype_db_sync_config = self.sync_configs.get("grype")
```
zburstein, line 236: Should this also use the `SupportedProvider` enum value as opposed to a string

anchore_engine/services/policy_engine/engine/feeds/config.py
nightfurys: The purpose of using an enum here is 
1. to keep a collection of normalized providers in one place and avoid sprinkling literal strings through out the system. When we add a new provider, that enum should be updated along 
2. to validate the supplied provider

 I understand this doesn't have to be an enum necessarily but defining a constant for each provider and then create a list with all of them was additional steps that didn't seem necessary

anchore_engine/services/policy_engine/engine/feeds/config.py
nightfurys: `.get(feed_name)` can result in `None` if `feed_name` is not a supported feed for the provider. So default_sync_config is *not* guaranteed to be not None

anchore_engine/services/policy_engine/__init__.py
nightfurys: I would rather keep this out of the task. Tasks by definition are do-this-one-thing-as-you-are-told. The burden of deciding whether to fire a task or not is on the caller.

zhill (commented):
I see some issues with the provider interface, but they are not introduced in this PR. Some nits/suggestions, but assuming comments are addressed and testing is passing. I'm ok. Since this looks like a breaking configuration change we need to make a big note in the release notes and ensure that docker-compose and charts are updated.

anchore_engine/services/policy_engine/engine/feeds/config.py
Vijay-P: Yep, i'm saying you should return the enum, not the value, and compare against the enum in the caller.

anchore_engine/services/policy_engine/engine/feeds/config.py
Vijay-P: i'm referring to lines 134-140

Vijay-P:
```
    if not provider or provider not in DEFAULT_PROVIDER_SYNC_CONFIG_MAP:
        logger.warn(
            "%s is not a supported provider. Defaulting to %s",
            provider,
            DEFAULT_VULNERABILITIES_PROVIDER,
        )
        provider = DEFAULT_VULNERABILITIES_PROVIDER
```

PR #1039, opened 2021-05-18
Grype wrapper checksum fix

dspalmer99:
Fixes a bug on the method that returns the current grype db checksum that is currently causing one of the functional tests to fail, and adds some unit tests on the method. We are replacing that method in the very near future, but the extra tests won't hurt in the interim.

I also improved a few comments on the existing tests.


dspalmer99:
Closing this, DCO found older commits (not from this branch) in the commit history that are not signed, and its suggested fix messed up the commit history. It will be fastest to just create a new branch.

PR #1040, opened 2021-05-18
Fix bug in grype_wrapper on get_current_grype_db_checksum(), add some…

dspalmer99:
Fixes a bug on the method that returns the current grype db checksum that is currently causing one of the functional tests to fail, and adds some unit tests on the method. We are replacing that method in the very near future, but the extra tests won't hurt in the interim.

I also improved a few comments on the existing tests.

### DIFF
--- a/tests/unit/anchore_engine/clients/test_grype_wrapper.py
+++ b/tests/unit/anchore_engine/clients/test_grype_wrapper.py
@@ -89,6 +89,7 @@ def test_get_missing_grype_db_dir():
         # Function under test
         grype_wrapper_singleton._grype_db_dir
 
+    # Validate error message
     assert str(error.value) is GrypeWrapperSingleton.MISSING_GRYPE_DB_DIR_ERROR_MESSAGE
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/anchore_engine/clients/test_grype_wrapper.py
```diff
@@ -100,12 +101,53 @@ def test_get_missing_grype_db_session():
...
+def test_get_current_grype_db_checksum_missing_db_dir():
+    # Create grype_wrapper_singleton instance
+    grype_wrapper_singleton = TestGrypeWrapperSingleton.get_instance()
+
+    with pytest.raises(ValueError) as error:
+        # Function under test
+        result = grype_wrapper_singleton.get_current_grype_db_checksum()
+
+    # Validate error message
+    assert str(error.value) == GrypeWrapperSingleton.MISSING_GRYPE_DB_DIR_ERROR_MESSAGE
```
Vijay-P, line 148: pytest.raises will actually throw an exception if the expected exception isn't raised. Is the second assertion necessary? if you're writing these checks, it may be a sign that it's time to implement custom exception classes for the grype wrapper.

https://docs.pytest.org/en/6.2.x/reference.html#pytest-raises

tests/unit/anchore_engine/clients/test_grype_wrapper.py
```diff
@@ -100,12 +101,53 @@ def test_get_missing_grype_db_session():
...
+    grype_wrapper_singleton = TestGrypeWrapperSingleton.get_instance()
+
+    # Setup test input with a non-existant path
+    grype_wrapper_singleton._grype_db_dir = "/does/not/exist"
+
+    # Function under test
+    result = grype_wrapper_singleton.get_current_grype_db_checksum()
+
+    # Validate error message
+    assert result is None
```
Vijay-P, line 136: should this be an exception case too?

Vijay-P (approved):
I understand that this is temporary, but may want to consider changes to exception handling in subsequent PR.

tests/unit/anchore_engine/clients/test_grype_wrapper.py
dspalmer99: No. The exception is raised if the varliable is not set. The current (and soon-to-be-replaced) logic only returns None if the variable is set but the dir for some reason doesn't actually exist. Should never happen.

tests/unit/anchore_engine/clients/test_grype_wrapper.py
dspalmer99: More of a nicety to confirm the exception message.

tests/unit/anchore_engine/clients/test_grype_wrapper.py
dspalmer99: Ah I see the match parameter in raises() now, I didn't know about that one. Thanks!

dspalmer99:
Looks like there may be a second, previously-masked issue with the functional test itself that needs to also be fixed.

zburstein (approved):
lgtm

Issue #1036, opened 2021-05-17
Add support for `grype version` command in grype-wrapper


Vijay-P:
**Is this a request for help?**:

NO
---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one):

FEATURE REQUEST

**Version of Anchore Engine and Anchore CLI if applicable**:
N/A grype-integration branch

**What happened**:
Schema version of grype DB download should be keyed off of grype installation. (We should download the version that is supported by the installed grype version). This information can be retrieved from the `grype version` command. See linked discussion: https://github.com/anchore/anchore-engine/pull/1028#discussion_r632270861

**What did you expect to happen**:

`grype-wrapper` should have a method for calling `grype version` and serialize the response into a dataclass.

**Any relevant log output from /var/log/anchore**:

**What docker images are you using**:

**How to reproduce the issue**:

**Anything else we need to know**:

dspalmer99:
Done

PR #1042, opened 2021-05-18
Grypedb version temporary fix

Vijay-P:
This is a temporary fix until the grype-wrapper is updated to support the `grype version` command and the installed version of grype is upgraded to the latest available version.

https://github.com/anchore/anchore-engine/issues/1036


zburstein (approved):
lgtm

nightfurys (approved):
Looks good for now, should be followed up with the commits for grype-db-wrapper driving the db version and setting up the db directory correctly

PR #1045, opened 2021-05-19
Merge prep

nightfurys:
This PR
- adds an environment variable for setting vulnerabilities provider and defaults it to legacy provider. selecting grype as the provider would be possible by setting this variable in docker-compose
- improves the mappers based on feedback from #1033

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/cache_managers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/cache_managers.py
@@ -1,22 +1,18 @@
-import dataclasses
 import datetime
 import enum
-import hashlib
-import json
+from dataclasses import dataclass
+from typing import Dict
 
-from anchore_engine import utils
+from anchore_engine.apis.context import ApiRequestContextProxy
 from anchore_engine.clients.services import internal_client_for, catalog
 from anchore_engine.db import (
     Image,
     get_thread_scoped_session as get_session,
     CachedVulnerabilities,
 )
-from anchore_engine.services.policy_engine.engine.feeds.db import get_all_feeds
-from anchore_engine.subsys import logger as log, metrics
-from anchore_engine.apis.context import ApiRequestContextProxy
-from dataclasses import dataclass
-from typing import Dict, List
 from anchore_engine.services.policy_engine.api.models import ImageVulnerabilitiesReport
+from anchore_engine.subsys import logger as log
+from anchore_engine.clients.grype_wrapper import GrypeWrapperSingleton
 
 # Disabled by default, can be set in config file. Seconds for connection to cache
 DEFAULT_CACHE_CONN_TIMEOUT = -1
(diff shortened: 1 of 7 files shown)


anchore_engine/services/policy_engine/engine/vulns/cache_managers.py
```diff
@@ -150,7 +147,11 @@ def _get_current_grype_metadata():
         TODO Invoke the grype_wrapper and generate a grype metadata object, work with dspalmer
         """
 
-        return GrypeMetadata("foo", "bar", "sha256:blah")
+        # return GrypeMetadata("", "", "")
+        archive_checksum = (
+            GrypeWrapperSingleton.get_instance().get_current_grype_db_checksum()
+        )
+        return {"checksum": archive_checksum}
```
zburstein, line 154: Will this ever be called before the grype wrapper has initialized the db? Im guessing not but just wanted to double check because in my code we recently had to handle a ValueError that gets thrown if the grype_db_dir has not yet been initialized.

conf/default_config.yaml
```diff
@@ -161,23 +161,22 @@ services:
...
         read_timeout_seconds: 60
         data:
-          # Uncomment if the provider is set to grype. If packages feed was enabled and necessary leave it uncommented.
-          # All remaining feeds can be commented out
-          #grypedb:
-          #  enabled: true
-          #  url: ${ANCHORE_GRYPE_DB_URL}
-          # The following feeds except packages are required only for legacy provider. Comment them out if the provider is set to grype
+          # grypedb feed is synced if the provider is set to grype. All the remaining feeds except for packages are ignored even if they are enabled
+          grypedb:
```
Vijay-P, line 176: you uncommented this. intentional?

nightfurys: Yep, allowed feeds are maintained in the provider, shouldn't matter what the config has

anchore_engine/services/policy_engine/engine/vulns/cache_managers.py
nightfurys: I noticed. When correctly implemented this won't use the wrapper. It'll instead use the active db record to decide whether the cache is valid or not. That's one of the reasons why the grype-db checksum (not just the archive) must be stored

conf/default_config.yaml
zburstein: I think it is tied to the provider now where it will do an either or based upon what provider is set. See providers [default config](https://github.com/anchore/anchore-engine/blob/55bb293f9d8edb54f82c6825d36f30238fb029bf/anchore_engine/services/policy_engine/engine/vulns/providers.py#L748-L748) and how that it is used [in the config](https://github.com/anchore/anchore-engine/blob/55bb293f9d8edb54f82c6825d36f30238fb029bf/anchore_engine/services/policy_engine/engine/feeds/config.py#L124-L147)

PR #1046, opened 2021-05-19
Update docs to have search engine and analytics

zhill:
Adds custom search id and analytics back into the docs configuration for engine.anchore.io to make the docs easier to use.


zhill:
I think I need to use a different GA token here. So will hold off on merge of this until that change is in.

DaneWeber:
Can this be tested locally? The comments seem to indicate that this will enable a site-search capability and not just tracking stat collection.

zhill:
> Can this be tested locally? The comments seem to indicate that this will enable a site-search capability and not just tracking stat collection.

No, unfortunately these are only enabled in production builds since they would actually send analytics data. They are standard features of the docsy theme, however, so they are not custom functionality we're implementing.

Vijay-P (approved):

PR #1049, opened 2021-05-19
Suppress and log NoActiveGrypeDB on system start

Vijay-P:
GrypeDBSyncTask had some tracebacks logged on startup because of the NoActiveGrypeDBException. The exception raising behavior is correct, but the exception needed to be handled appropriately. I added a try/except in the GrypeDBSyncTask to solve this issue. Alternative location for this is anchore_engine.services.policy_engine.handle_grypedb_sync. Such placement would allow the exception to get propagated to other callers of GrypeDBSyncTask. Currently, there are no other consumers of GrypeDBSyncTask, so the decision is still somewhat arbitrary. Those who need the functionality in GrypeDBSyncTask currently use the upstream code in GrypeDBSyncManager directly.

### DIFF
--- a/anchore_engine/services/policy_engine/engine/tasks.py
+++ b/anchore_engine/services/policy_engine/engine/tasks.py
@@ -27,17 +27,18 @@ from anchore_engine.services.policy_engine.engine.feeds.client import (
     get_feeds_client,
     get_grype_db_client,
 )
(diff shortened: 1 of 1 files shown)


zburstein (approved):
lgtm

nightfurys (approved):
lgtm

PR #1055, opened 2021-05-19
Merge grype integration

nightfurys:
At a high level this PR adds

- configurability around choosing a provider for vulnerabilities - supported values are legacy and grype, defaults to legacy (to avoid breaking existing tests, grype related tests coming soon)
- extension of feeds mechanism for downloading grype db from an API (listing.json) endpoint and syncing it across all policy engines
- a read-write safe wrapper for grype
- a new provider, scanner and cache manager that currently only supports image vulnerabilities. Support for other vulnerability related queries coming soon


Vijay-P (approved):
Built and ran. DB synced and image scan was successful. LGTM

zburstein (approved):

PR #1058, opened 2021-05-21
Fix grype sbom string analysis

dspalmer99:
**What this PR does / why we need it**: Fixes a bug in grype wrapper where special characters in sboms that we normally filter on for subcommand execution ("[;&<>]" were causing those sboms to not be analyzable. utils.run_piped_command_list() can now be optionally parameterized to skip that filtering, which the grype-wrapper now uses. The default for the function elsewhere is still to do the filtering.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #:1056

**Special notes**:

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -568,7 +568,9 @@ class GrypeWrapperSingleton(object):
             stdout = None
             err = None
             try:
-                _, stdout, _ = run_piped_command_list(full_cmd, env=proc_env)
+                _, stdout, _ = run_piped_command_list(
+                    full_cmd, env=proc_env, sanitize_input=False
+                )
             except CommandException as exc:
                 logger.error(
                     "Exception running command: %s | %s, stderr: %s",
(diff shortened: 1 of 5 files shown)


dspalmer99:
Ignore this for the moment please - I branched this from grype-integration instead of master (forgot the merge had already happened) so the commit history is not right. Will fix.

dspalmer99:
Fixed

nightfurys (commented):
This change makes sense for fixing the issue. My only concern is do we have any other checks in place to ensure that the input is somewhat safe?

dspalmer99:
No, but the method being made optional is somewhat misleadingly named. We are not really sanitizing the input, we're just looking for characters that would normally cause shell commands to fail or otherwise not be executed in the way the user expects.  Since the base commands we run though there aren't dynamically generated by the system (ie users can't execute arbitrary commands) this is primarily a protection at development time.

I also only made it optional only on the flavor of that method that runs piped commands. Nothing else uses that (yet), and I tried to make it clear with a big comment that that option was only really useful for commands with big arbitrary strings (like sboms) in them being passed as arguments to the command.

So overall I think it is very minimal risk, but I also think that code overall could be target for refactoring in a future tech debt push. I wanted to streamline those methods when I added the pipelined commands bit, but the existing methods were used in enough places that the scope to do that was not small. Still something I'd like to circle back on in a future PR though.

Vijay-P (approved):
lgtm

PR #1060, opened 2021-05-21
Grype db data model corrections

Vijay-P:

- Rename GrypeDBMetadata to GrypeDBFeedRecord
- Include the database checksum and listing checksum
- Rename date_generated to built_at
- New field synced_at (set when the db is synced in grype-db sync task)
- Remove refs to group and feed (irrelevant)
- Remove foreign key constraints

**Special notes**:

### DIFF
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -146,9 +146,9 @@ class GrypeDBMetadata(Base):
 
     __tablename__ = "grype_db_metadata"
 
-    checksum = Column(String, primary_key=True)
+    archive_checksum = Column(String, primary_key=True)
     schema_version = Column(String, nullable=False)
-    feed_name = Column(String, ForeignKey(FeedMetadata.name), nullable=False)
+    feed_name = Column(String, nullable=False)
     group_name = Column(String, nullable=False)
     date_generated = Column(DateTime, nullable=False)
     object_url = Column(String, nullable=False)
@@ -156,13 +156,6 @@ class GrypeDBMetadata(Base):
     created_at = Column(DateTime, default=datetime.datetime.utcnow)
     last_update = Column(
         DateTime, default=datetime.datetime.utcnow, onupdate=datetime.datetime.utcnow
-    )
-
-    __table_args__ = (
-        ForeignKeyConstraint(
-            columns=(feed_name, group_name),
-            refcolumns=(FeedGroupMetadata.feed_name, FeedGroupMetadata.name),
-        ),
     )
 
 
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -28,7 +28,7 @@ from sqlalchemy import (
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import joinedload, relationship, synonym
 
-from anchore_engine.db.entities.common import anchore_uuid
+from anchore_engine.db.entities.common import anchore_now_datetime, anchore_uuid
 from anchore_engine.util.apk import compare_versions as apkg_compare_versions
 from anchore_engine.util.deb import compare_versions as dpkg_compare_versions
 from anchore_engine.util.langpack import compare_versions as langpack_compare_versions
@@ -147,14 +147,20 @@ class GrypeDBFeedMetadata(Base):
 
     __tablename__ = "grype_db_feed_metadata"
 
-    uuid = Column(String, nullable=False, primary_key=True, default=anchore_uuid)
+    uuid = Column(String, primary_key=True, default=anchore_uuid)
     archive_checksum = Column(String, nullable=False, index=True)
     metadata_checksum = Column(String, nullable=True, index=True)
     schema_version = Column(String, nullable=False)
     object_url = Column(String, nullable=False)
     active = Column(Boolean, nullable=False)
     built_at = Column(DateTime, nullable=False)
-    created_at = Column(DateTime, default=datetime.datetime.utcnow)
+    created_at = Column(DateTime, default=anchore_now_datetime, nullable=False)
+    last_updated = Column(
+        DateTime,
+        default=anchore_now_datetime,
+        onupdate=anchore_now_datetime,
+        nullable=False,
+    )
     synced_at = Column(DateTime, nullable=True)
     groups = Column(JSONB)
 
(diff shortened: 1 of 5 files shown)


anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -820,26 +820,26 @@ def _switch_active_grypedb(
...
+            archive_checksum=checksum,
+            metadata_checksum=None,
             schema_version=record.metadata["version"],
-            feed_name=GrypeDBFeed.__feed_name__,
-            group_name=group_download_result.group,
-            date_generated=date_generated,
             object_url=object_url,
             active=True,
+            built_at=built_at,
+            synced_at=None,
```
zburstein, line 842: Why set synced_at to None here?

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -820,26 +820,26 @@ def _switch_active_grypedb(
...
         object_url = catalog_client.create_raw_object(
             group_download_result.group, checksum, record.data
         )
-        date_generated = rfc3339str_to_datetime(record.metadata["built"])
-        grypedb_meta = GrypeDBMetadata(
-            checksum=checksum,
+        built_at = rfc3339str_to_datetime(record.metadata["built"])
+        grypedb_meta = GrypeDBFeedMetadata(
+            archive_checksum=checksum,
+            metadata_checksum=None,
```
zburstein, line 837: Maybe when we implement the change that @nightfurys was talking about last week of having the feeds update process pass the db to the wrapper to do a temp build to verify that there are no issues with it, the grype wrapper can return the metadata checksum. This would also allow us to make the metadata_checksum non nullable

zburstein (commented):
Might also be good to do a global string search for `GrypeDBMetadata` because there are some docstrings that need to be updated

anchore_engine/services/policy_engine/engine/feeds/feeds.py
Vijay-P: yep, that's the idea. the nullable change is temporary

Vijay-P:
> Might also be good to do a global string search for `GrypeDBMetadata` because there are some docstrings that need to be updated

done. we should be good

Vijay-P:
FYI just rebased this off of master

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -726,16 +726,16 @@ def _find_match(
         :return: sqlalchemy query object
         :rtype: Query
         """
-        results = db.query(GrypeDBMetadata)
+        results = db.query(GrypeDBFeedMetadata)
         if checksum:
-            results = results.filter(GrypeDBMetadata.checksum == checksum)
+            results = results.filter(GrypeDBFeedMetadata.archive_checksum == checksum)
         if not isinstance(active, type(None)):
```
nightfurys, line 732: Check for a boolean here is more accurate
```
if isinstance(active, bool)
```

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -726,16 +726,16 @@ def _find_match(
...
-            results = results.filter(GrypeDBMetadata.active == active)
+            results = results.filter(GrypeDBFeedMetadata.active == active)
         return results
```
nightfurys, line 734: Return the results here, not the query. Operate on the returned objects using the session as necessary. Details in the below comment

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -820,26 +820,25 @@ def _switch_active_grypedb(
...
             catalog_client.delete_document(
-                inactive_record.group_name, inactive_record.checksum
+                inactive_record.group_name, inactive_record.archive_checksum
             )
         inactive_records.delete(synchronize_session="evaluate")
         # search for active and mark inactive
         self._find_match(db, active=True).update(
-            {GrypeDBMetadata.active: False}, synchronize_session="evaluate"
+            {GrypeDBFeedMetadata.active: False}, synchronize_session="evaluate"
         )
```
nightfurys, line 829: The use of synchronization strategy is not correct here. Evaluate as a synchronization strategy means locate objects with in the session but the intent here is to lookup all the active/inactive db records.  https://docs.sqlalchemy.org/en/14/orm/session_basics.html#selecting-a-synchronization-strategy sqlalchemy generally recommends refraining from using the sync strategy

The recommended way of operating on db objects is by directly updating them (sqlalchemy recognizes the changes and updates the record when the session is committed) or by using the session to delete the object. As I indicated above find_match should return the results, not the query (the naming is confusing here)

```
        inactive_records = self._find_match(db, active=False)
        for inactive_record in inactive_records:
            catalog_client.delete_document(
                inactive_record.group_name, inactive_record.archive_checksum
            )
           db.delete(inactive_record)

       for active_record in self._find_match(db, active=True):
          active_record.active = False
```

nightfurys (changes requested):
Looks good for the most part. Just a few suggestions around sqlalchemy and persistence context usage

anchore_engine/db/entities/policy_engine.py
```diff
@@ -144,26 +145,18 @@ class GrypeDBMetadata(Base):
     There should only ever be a single active record. More than one indicates an error in the system
     """
 
-    __tablename__ = "grype_db_metadata"
+    __tablename__ = "grype_db_feed_metadata"
 
-    checksum = Column(String, primary_key=True)
+    uuid = Column(String, nullable=False, primary_key=True, default=anchore_uuid)
+    archive_checksum = Column(String, nullable=False, index=True)
+    metadata_checksum = Column(String, nullable=True, index=True)
```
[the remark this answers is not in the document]
Vijay-P: it's a temporary thing until we re-work the flow of feed sync to verify the db before marking a new file active

anchore_engine/services/policy_engine/engine/feeds/feeds.py
Vijay-P: thanks! i did the sync strategy stuff when i was still relatively new to sqlalchemy. docs definitely make more sense now than they did then. will fix. (i think zane might do this since he's working on that method in this branch right now)

nightfurys (approved):

PR #1061, opened 2021-05-21
Handle too many active dbs

zburstein:
This PR does two separate things:

1. Changes GrypeDBSyncManager to gracefully handle situations in which there are TooManyActiveDBs. It does this by querying all the active dbs and determines the true active to be the most rececntly created. It then set all the other active dbs to active=False. By doing this, when the next sync runs, the sync task will cleanup all inactive dbs. This new public getter is called `resolve_and_get_active_grypedb`
2. Changes how sessions are managed in the class by having db sessions get passed by the caller

### DIFF
--- a/anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
+++ b/anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
@@ -2,12 +2,16 @@ import threading
 from types import TracebackType
 from typing import Optional, Type
 
-import sqlalchemy
+from sqlalchemy.orm import Session
 
 from anchore_engine.clients.grype_wrapper import GrypeWrapperSingleton
 from anchore_engine.clients.services import internal_client_for
 from anchore_engine.clients.services.catalog import CatalogClient
-from anchore_engine.db import GrypeDBFeedMetadata, get_thread_scoped_session
+from anchore_engine.db import GrypeDBFeedMetadata
+from anchore_engine.db.db_grype_db_feed_metadata import (
+    NoActiveGrypeDB,
+    get_most_recent_active_grypedb,
+)
 from anchore_engine.services.policy_engine.engine.feeds.storage import (
     GrypeDBFile,
     GrypeDBStorage,
@@ -21,18 +25,9 @@ class GrypeDBSyncError(Exception):
     pass
 
 
-class TooManyActiveGrypeDBs(GrypeDBSyncError):
+class NoActiveDBSyncError(GrypeDBSyncError):
     def __init__(self):
-        super().__init__(
-            "Could not determine correct grypedb to sync because too many active dbs found in database"
-        )
-
-
-class NoActiveGrypeDB(GrypeDBSyncError):
-    def __init__(self):
-        super().__init__(
-            "Could not determine correct grypedb to sync because no active db found in the database"
-        )
+        super().__init__("Local sync failed because no active db found in the database")
 
 
 class GrypeDBSyncLockAquisitionTimeout(GrypeDBSyncError):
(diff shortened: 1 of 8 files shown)


anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -73,41 +73,82 @@ class GrypeDBSyncManager:
...
+        Commits to session object
+        """
+        # get all active dbs and select the most recent one to be the correct active db
+        active_dbs = (
+            session.query(GrypeDBFeedMetadata)
+            .filter(GrypeDBFeedMetadata.active == True)
+            .order_by(GrypeDBFeedMetadata.created_at.desc())
+            .all()
+        )
+        resolved_active = active_dbs.pop(0)
```
nightfurys, line 140: This will fail when active the list is empty. I understand that might not be the case when this funciton is called but a slightly safer and efficient way of doing this is in python is list slicing  
```
if active_dbs and len(active_dbs) > 1:
  for outdated_db in active_dbs[1:]:
    outdated_db.active = False
```

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -73,41 +73,82 @@ class GrypeDBSyncManager:
...
+        """
+        Resolves active db by querying all active ones sorted by created date.
+        Sets all older ones to inactive
+        Commits to session object
+        """
+        # get all active dbs and select the most recent one to be the correct active db
+        active_dbs = (
+            session.query(GrypeDBFeedMetadata)
+            .filter(GrypeDBFeedMetadata.active == True)
+            .order_by(GrypeDBFeedMetadata.created_at.desc())
```
nightfurys, line 137: Use the last_updated timestamp as that gets updated when the record is modified

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -73,41 +73,82 @@ class GrypeDBSyncManager:
...
+
+        logger.info(
+            "Resolving grype db with checksum of %s to be active db",
+            resolved_active.archive_checksum,
+        )
+
+        # Set the remainder to inactive
+        for outdated_db in active_dbs:
+            outdated_db.active = False
+        session.commit()
```
nightfurys, line 150: Is it okay to commit the session here? I think there may be ops requiring session after this function is called

nightfurys (commented):
get active grype db path seems a little convoluted. Followed up the discussion and I think we are on the same page. Pausing on this review until further changes

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -73,40 +66,45 @@ class GrypeDBSyncManager:
...
-        return: Instance of GrypeDBFeedMetadata or None
-        rtype: GrypeDBFeedMetadata
-        """
-        db = get_thread_scoped_session()
-        return (
-            db.query(GrypeDBFeedMetadata)
-            .filter(GrypeDBFeedMetadata.active == True)
-            .one_or_none()
-        )
+            session.commit()
```
nightfurys, line 105: @zburstein committing the session closes it, i don't that's correct here

nightfurys (changes requested):
Question about the session.commit and other improvements discussed over slack

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
zburstein: I thought commit closes the transaction, not the session

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
```diff
@@ -73,41 +73,82 @@ class GrypeDBSyncManager:
...
 
-        return: Instance of GrypeDBFeedMetadata or None
-        rtype: GrypeDBFeedMetadata
+        return: List of active GrypeDBFeedMetadata
+        rtype: List[GrypeDBFeedMetadata]
         """
-        db = get_thread_scoped_session()
         return (
-            db.query(GrypeDBFeedMetadata)
+            session.query(GrypeDBFeedMetadata)
```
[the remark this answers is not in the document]
zburstein: Discussed and decided to use creation time since that will match most recent upstream. Moving query to dao

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
zburstein: Using created at.  See comment in previous comment

anchore_engine/services/policy_engine/engine/feeds/grypedb_sync.py
zburstein: No longer resolving through actually editing records

anchore_engine/services/policy_engine/engine/tasks.py
```diff
@@ -671,8 +671,9 @@ def execute(self) -> bool:
         rtype: bool
         """
         try:
-            return GrypeDBSyncManager.run_grypedb_sync(self.grypedb_file_path)
-        except NoActiveGrypeDB:
+            session = get_session()
+            return GrypeDBSyncManager.run_grypedb_sync(session, self.grypedb_file_path)
```
nightfurys, line 675: Any reason why context manager can't be used here? That'll handle transaction commit/rollback and close the session 
`with session_scope() as session:`

nightfurys (approved):
lgtm!

PR #1063, opened 2021-05-22
Updates to grype provider

nightfurys:
This PR updates the grype provider with
- a store mechanism (instead of a cache) that saves one report per image at any given time
- a configurable model for overriding report status computation connected to the above store
- mechanism for maintaining state of image vulnerabilities by transferring them from an existing report if any


anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -1,57 +1,108 @@
...
+        db_checksum = report_key.get("db_checksum") if report_key else None
+        if db_checksum:
+            return GrypeDBKey(db_checksum=db_checksum)
+        else:
+            raise ValueError(
+                "Invalid or unexpected report_key format {}".format(report_key)
+            )
+
+    @classmethod
+    def from_report(cls, report: ImageVulnerabilitiesReport):
```
zhill, line 63: I think the db checksum that generated it should be included in the report itself as metadata and set when the report is generated. It can be used here as a key, but seems like its creation and lifecycle should live with the report itself.

anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -1,57 +1,108 @@
...
+        db_checksum = (
+            GrypeWrapperSingleton.get_instance().get_current_grype_db_checksum()
         )
+        return GrypeDBKey(db_checksum=db_checksum)
 
     def to_dict(self):
         return self.__dict__
 
+    @classmethod
+    def get_report_status(cls, report_key: Dict):
```
zhill, line 74: Is this really a class function of the Key class? Seems like a function that just takes the key as input but has no special knowledge of the class or state for the class. Binding these functions to the key class itself makes it harder to, for example, swap key implementations (but with a consistent interface) in the future.

anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -119,40 +172,12 @@ def _lookup(self):
...
-        TODO Invoke the grype_wrapper and generate a grype metadata object, work with dspalmer
-        """
-
-        # return GrypeMetadata("", "", "")
-        archive_checksum = (
-            GrypeWrapperSingleton.get_instance().get_current_grype_db_checksum()
-        )
-        return {"checksum": archive_checksum}
-
     def _delete_entry(self, entry):
```
zhill, line 181: Is this still needed?

anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -1,57 +1,108 @@
...
+            status = Status.invalid
 
-class GrypeCacheManager:
+        return status
 
-    __cache_bucket__ = "policy-engine-vulns-cache"
+
+class ImageVulnerabilitiesStore:
+
+    __store_bucket__ = "policy-engine-vulns-store"
```
zhill, line 104: Is this still necessary? seems like a DB-approach no longer needs a bucket

anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -169,62 +194,57 @@ def _delete_entry(self, entry):
...
+        Flush all report entries for the given image
         :return:
         """
         session = get_session()
-        for entry in session.query(CachedVulnerabilities).filter_by(
+        for entry in session.query(DbImageVulnerabilities).filter_by(
             account_id=self.image.user_id, image_digest=self.image.digest
         ):
             try:
                 self._delete_entry(entry)
```
zhill, line 220: Is this still necessary? Looks like a no-op

anchore_engine/services/policy_engine/engine/vulns/dedup.py
```diff
@@ -172,3 +172,52 @@ def execute(
...
+def _transform_vuln_match_list_to_map(
+    vuln_matches: List[VulnerabilityMatch],
+) -> Dict[str, VulnerabilityMatch]:
+    """
+    Returns a dict from a list of VulnerabilityMatch objects where the key is a tuple representation of VulnerabilityMatch
+    Choosing a tuple over other data structures for the key as it to be the quickest in terms of instantiation
+
+    """
+    if vuln_matches:
+        return {
```
zhill, line 211: If the VulnerabilityMatch object has a key function or identifier, it should be used here as well if possible to avoid duplication of this logic in other places. Perhaps make this tuple a function, identity_tuple() of that class since this seems to be a property of the match itself that should be opaque to other code as long as it serves as an identity (hash or __eq__ stuff may also be useful for that class, but not strictly required for this use-case).

zhill (commented):
Looks like there are some dead code paths and other tweaks that can clean up the transition from obj store to db usage. The main question is if the store should be required to be successful in order for a result to be returned. If other parts of the system expect that store to be consistent with the image set then yes (e.g. timestamp merges), and we should review and ensure things are blocking on failures rather than the cache-like behavior of swallowing errors since they are just performance issues.

anchore_engine/services/policy_engine/engine/vulns/stores.py
nightfurys: db checksum will be  accessed from the report but it's not currently available from grype. It will be with the next grype update. That's why the argument to the function is a report model object. This is a temporary workaround with a TODO right below :)

anchore_engine/services/policy_engine/engine/vulns/stores.py
nightfurys: To be able to swap out the key implementation is the purpose of having the logic encapsulated like this. I made it a class method initially since it was returning an instance of the class. Will fix it

anchore_engine/services/policy_engine/engine/vulns/stores.py
nightfurys: How's it a no-op? It still deletes the existing record before creating a new one in the save() workflow

anchore_engine/services/policy_engine/engine/vulns/stores.py
zhill: Ah, ok it wasn't clear that ImageVulnerabilitiesReport should be raw and not include any wrappers or metadata envelope. I think it could with a metadata property and then the "matches" or similar. It would just be a slightly higher level abstraction, but if it's expected soon from Grype itself then can add it then.

anchore_engine/services/policy_engine/engine/vulns/stores.py
zhill: Yeah I get that, not a big deal, but its a pattern I'm seeing that limits function re-use in cases where others may find a function useful but not need the broader class context since the function doesn't rely on it. And testing also requires initializing the class which can involve other dependencies. Not an issue here, but we should be careful with the pattern of making everything class methods since it can lead to a lot of code duplication or cross-class invocations that shouldn't actually exist.

PR #1062, opened 2021-05-21
Grype stage db

dspalmer99:
Updates grype wrapper to stage grype db updates before applying them, so they can be validated upstream.

PR #1065, opened 2021-05-24
Better account for db version in path to grype dir, fix and add some tests.

dspalmer99:
Fixes a bug @zburstein found in the grype wrapper where the new file path with the grype db version in it isn't always resolved correctly. Adds and updates some tests around the same, and also fixes a few cosmetic test var and function name fixes I noticed.


dspalmer99:
Note that this overlaps slightly in a few places with the change I'm making in https://github.com/anchore/anchore-engine/pull/1062/files as I needed one of the improvements from there to do this. This is a smaller change that fixes a current bug so I want to get this in first - I will rebase 1062 after this is merged.

zburstein (approved):

PR #1066, opened 2021-05-24
Bump to UBI 8.4

robertp:


nightfurys (commented):
lgtm. Just curious if there's been an ask to bump the ubi tag? And does it introduce any dependency updates as well?

nightfurys:
Other upstream repositories might also need a similar update

zhill (commented):
lgtm

robertp:
> lgtm. Just curious if there's been an ask to bump the ubi tag? And does it introduce any dependency updates as well?



> lgtm. Just curious if there's been an ask to bump the ubi tag? And does it introduce any dependency updates as well?

@nightfurys IronBank is mandating UBI 8.4 now so it seems reasonable to bump our version rather than hand-crafting a release w/different upstreams for them.

PR #1071, opened 2021-05-25
Enable GrypeDB to sync without creating FeedGroupMetadata

Vijay-P:

The feed sync implementation was heavily tied to individual FeedGroupMetadata records. Meaning, if one wanted to sync a DataFeed, one had to first create at least one FeedGroupMetadata for that DataFeed. This is what we were doing for Grype DB. As a result, there was a meaningless group record created called `grypedb:vulnerabilities`. The objective is to expose feed group information to the user that reflects the contents of the Grype DB sqlite file. As such, this workflow needed to change. The changes in this PR allow GrypeDBFeed to sync without being tied to or creating a new FeedGroupMetadata record in the database. All of the logic that changes depending on whether the provider is set to grype has been abstracted through the VulnerabilityProvider implementations. There, a new get_sync_utils() method returns an implementation of SyncUtilProvider, of which there is one for Grype and one for Legacy. These SyncUtilProviders contain the key differences between the two workflows, and the DataFeeds.sync() method has been updated to use them.

-  Update GrypeDBServiceClient so that list_feed_groups returns listing information. (Also add new serializable data model to FeedAPIGroupRecord for GrypeDBListing)
-  Split sync_metadata logic so that FeedGroupMetadata sync can be turned off. Update tasks.FeedsUpdateTask and sync.sync() so that all logic differences between legacy and grype flows are implemented through the providers.VulnerabilityProvider abstraction. This allows Grype feed sync to proceed without the creation of  FeedGroupMetadata objects.
-  Make GrypeDBFeed a subclass of DataFeeds instead. Move common logic shared across all feeds from AnchoreServiceFeed to DataFeeds. Remove all use of FeedGroupMetadata from GrypeDBFeed.
-  Update delete_feed, delete_group, toggle_feed_enabled, and toggle_group_enabled api controllers to return HTTP 501 if feed is grypedb feed. 
    - GrypeDB groups will be tied to a single copy of the Grype database, and cannot be deleted, but the whole feed can.
    - GrypeDB feed can be deleted and resynced, but the implementation doesn't yet have enough safety controls around the operation to be confident in the functionality.
    - Enabling/disabling the grypedb feed when grype is the provider will remove the main source of vuln info and may cause unexpected behavior.
- Updated integration tests to adjust for API changes.
- Added docstrings and correct type hints.
- Add new unit tests for SyncUtilProviders

### DIFF
--- a/anchore_engine/apis/exceptions.py
+++ b/anchore_engine/apis/exceptions.py
@@ -64,6 +64,10 @@ class BadGatewayError(AnchoreApiError):
     __response_code__ = 502
 
 
+class UnprocessableEntityError(AnchoreApiError):
+    __response_code__ = 422
+
+
 class MissingRequiredField(BadRequest):
     def __init__(self, required_property, because_of_properties=None):
         if because_of_properties:
--- a/anchore_engine/apis/exceptions.py
+++ b/anchore_engine/apis/exceptions.py
@@ -64,8 +64,8 @@ class BadGatewayError(AnchoreApiError):
     __response_code__ = 502
 
 
-class UnprocessableEntityError(AnchoreApiError):
-    __response_code__ = 422
+class HTTPNotImplementedError(AnchoreApiError):
+    __response_code__ = 501
 
 
 class MissingRequiredField(BadRequest):
--- a/anchore_engine/services/policy_engine/api/controllers/feeds.py
+++ b/anchore_engine/services/policy_engine/api/controllers/feeds.py
@@ -24,6 +24,9 @@ from anchore_engine.common.models.policy_engine import (
     FeedMetadata,
 )
 from anchore_engine.services.policy_engine.engine.feeds import db, sync
+from anchore_engine.services.policy_engine.engine.feeds.sync_utils import (
+    GRYPE_DB_FEED_NAME,
+)
 from anchore_engine.services.policy_engine.engine.tasks import FeedsUpdateTask
 from anchore_engine.subsys import logger as log
 
@@ -214,7 +217,7 @@ def delete_feed(feed):
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
 def delete_group(feed, group):
     session = db.get_session()
-    if feed == "grypedb":
+    if feed == GRYPE_DB_FEED_NAME:
         raise UnprocessableEntityError(
             message="Cannot delete individual groups for grypedb feed. Must delete entire feed.",
             detail={},
--- a/anchore_engine/services/policy_engine/api/controllers/feeds.py
+++ b/anchore_engine/services/policy_engine/api/controllers/feeds.py
@@ -131,11 +131,6 @@ def sync_feeds(sync=True, force_flush=False):
 def toggle_feed_enabled(feed, enabled):
     if type(enabled) != bool:
         raise BadRequest(message="state must be a boolean", detail={"value": enabled})
-    if feed == GRYPE_DB_FEED_NAME:
-        raise HTTPNotImplementedError(
-            message="Cannot disable/enable grypedb feed.",
-            detail={},
-        )
     session = db.get_session()
     try:
         f = db.set_feed_enabled(session, feed, enabled)
@@ -162,7 +157,7 @@ def toggle_group_enabled(feed, group, enabled):
         raise BadRequest(message="state must be a boolean", detail={"value": enabled})
     if feed == GRYPE_DB_FEED_NAME:
         raise HTTPNotImplementedError(
-            message="Cannot disable/enable groups for grypedb feed.",
+            message="Enabling and disabling groups for grypedb feed is not currently supported.",
             detail={},
         )
     session = db.get_session()
@@ -190,7 +185,7 @@ def toggle_group_enabled(feed, group, enabled):
 def delete_feed(feed):
     if feed == GRYPE_DB_FEED_NAME:
         raise HTTPNotImplementedError(
-            message="Cannot delete grypedb feed.",
+            message="Deleting the grypedb feed is not yet supported.",
             detail={},
         )
     session = db.get_session()
@@ -226,12 +221,12 @@ def delete_feed(feed):
 
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
 def delete_group(feed, group):
-    session = db.get_session()
     if feed == GRYPE_DB_FEED_NAME:
         raise HTTPNotImplementedError(
-            message="Cannot delete individual groups for grypedb feed.",
+            message="Deleting individual groups for the grypedb feed is not yet supported.",
             detail={},
         )
+    session = db.get_session()
     try:
         f = db.lookup_feed_group(db_session=session, feed_name=feed, group_name=group)
         if not f:
--- a/anchore_engine/services/policy_engine/api/controllers/feeds.py
+++ b/anchore_engine/services/policy_engine/api/controllers/feeds.py
@@ -9,6 +9,7 @@ from anchore_engine.apis.exceptions import (
     BadRequest,
     ConflictingRequest,
     ResourceNotFound,
+    UnprocessableEntityError,
 )
 from anchore_engine.clients.services.simplequeue import (
     LeaseAcquisitionFailedError,
@@ -34,7 +35,6 @@ def list_feeds(refresh_counts=False):
     """
     GET /feeds
 
-    :param include_counts (ignored since counts are handled in the record now)
     :param refresh_counts: forcibly update the group counts (not normally necessary)
     :return:
     """
@@ -214,6 +214,11 @@ def delete_feed(feed):
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
 def delete_group(feed, group):
     session = db.get_session()
+    if feed == "grypedb":
+        raise UnprocessableEntityError(
+            message="Cannot delete individual groups for grypedb feed. Must delete entire feed.",
+            detail={},
+        )
     try:
         f = db.lookup_feed_group(db_session=session, feed_name=feed, group_name=group)
         if not f:
--- a/anchore_engine/services/policy_engine/api/controllers/feeds.py
+++ b/anchore_engine/services/policy_engine/api/controllers/feeds.py
@@ -8,8 +8,8 @@ from anchore_engine.apis.exceptions import (
     AnchoreApiError,
     BadRequest,
     ConflictingRequest,
+    HTTPNotImplementedError,
     ResourceNotFound,
-    UnprocessableEntityError,
 )
 from anchore_engine.clients.services.simplequeue import (
     LeaseAcquisitionFailedError,
@@ -17,12 +17,9 @@ from anchore_engine.clients.services.simplequeue import (
 )
 from anchore_engine.common.errors import AnchoreError
 from anchore_engine.common.helpers import make_response_error
+from anchore_engine.common.models.policy_engine import FeedGroupMetadata, FeedMetadata
 from anchore_engine.db import FeedGroupMetadata as DbFeedGroupMetadata
 from anchore_engine.db import FeedMetadata as DbFeedMetadata
-from anchore_engine.common.models.policy_engine import (
-    FeedGroupMetadata,
-    FeedMetadata,
-)
 from anchore_engine.services.policy_engine.engine.feeds import db, sync
 from anchore_engine.services.policy_engine.engine.feeds.sync_utils import (
     GRYPE_DB_FEED_NAME,
@@ -218,7 +215,7 @@ def delete_feed(feed):
 def delete_group(feed, group):
     session = db.get_session()
     if feed == GRYPE_DB_FEED_NAME:
-        raise UnprocessableEntityError(
+        raise HTTPNotImplementedError(
             message="Cannot delete individual groups for grypedb feed. Must delete entire feed.",
             detail={},
         )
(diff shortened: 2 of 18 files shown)


anchore_engine/services/policy_engine/engine/feeds/sync_utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+from anchore_engine.db import FeedGroupMetadata
+from anchore_engine.services.policy_engine.engine.feeds.client import (
+    get_feeds_client,
+    get_grype_db_client,
+)
+from anchore_engine.services.policy_engine.engine.feeds.sync import DataFeeds
+from anchore_engine.subsys import logger
+
+GRYPE_DB_FEED_NAME = "grypedb"
```
zburstein, line 11: This string also exists on `GrypeDBFeed` as a class variable called __feed_name__. We might want to establish a single source of truth for the name of this feed that we can reference wherever needed as opposed to using static strings in multiple places

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -215,6 +214,11 @@ def delete_feed(feed):
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
 def delete_group(feed, group):
     session = db.get_session()
+    if feed == "grypedb":
```
zburstein, line 217: See my other comment about having a single source of truth for the name of the grype db feed that we can reference wherever needed

zburstein (commented):
Have not finished the review yet but wanted to note these two things that I noticed on a first pass

anchore_engine/services/policy_engine/engine/feeds/sync_utils.py
Vijay-P: You are right.

Vijay-P: Pushed a commit to fix this. I'm still using that constant, `GRYPE_DB_FEED_NAME`, but I'm setting it to the value of the `GrypeDBFeed.__feed_name__`.

anchore_engine/services/policy_engine/api/controllers/feeds.py
Vijay-P: This now uses the `GRYPE_DB_FEED_NAME` constant from sync_utils.py

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -947,6 +957,96 @@ def _process_group_file_records(
...
+                        "Performing data flush prior to sync as requested",
+                    )
+                )
+                self.flush_all()
+
+            # Iterate thru the records and commit
+
+            logger.info(
+                self._log_context.format_msg(
+                    "Syncing {} total update records into db".format(
```
zburstein, line 1000: Some more format logs in this method

anchore_engine/services/policy_engine/engine/feeds/client.py
```diff
@@ -472,13 +478,33 @@ def list_feeds(self) -> FeedList:
...
+            requests.get, self.feed_url, retries=self.RETRY_COUNT
+        )
+        if not listing_response.success:
+            raise HTTPStatusException(listing_response)
+        listings_json = json.loads(listing_response.content.decode("utf-8"))
+        required_grype_db_version = self._get_supported_grype_db_version()
+        available_dbs = listings_json.get("available").get(required_grype_db_version)
+        if not available_dbs:
+            raise GrypeDBUnavailable(required_grype_db_version)
+        raw_db_listing = available_dbs[0]
```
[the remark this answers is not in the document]
Vijay-P: yep. confirmed with toolbox team

anchore_engine/apis/exceptions.py
```diff
@@ -64,6 +64,10 @@ class BadGatewayError(AnchoreApiError):
     __response_code__ = 502
 
 
+class UnprocessableEntityError(AnchoreApiError):
+    __response_code__ = 422
```
Vijay-P, line 68: change this to 501 Not Implemented

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -214,6 +217,11 @@ def delete_feed(feed):
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
 def delete_group(feed, group):
     session = db.get_session()
+    if feed == GRYPE_DB_FEED_NAME:
```
Vijay-P, line 220: also update toggle_group_enabled

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -214,6 +217,11 @@ def delete_feed(feed):
...
+        raise UnprocessableEntityError(
```
Vijay-P, line 221: this logic needs to be moved into the providers

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -214,6 +217,11 @@ def delete_feed(feed):
...
+            message="Cannot delete individual groups for grypedb feed. Must delete entire feed.",
+            detail={},
```
Vijay-P, line 223: list_feeds and sync_feeds to be updated by zane

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -214,6 +217,11 @@ def delete_feed(feed):
...
+        )
     try:
         f = db.lookup_feed_group(db_session=session, feed_name=feed, group_name=group)
```
Vijay-P, line 226: don't allow grypedb feed to be disabled (don't support most ops for grype)

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -214,6 +217,11 @@ def delete_feed(feed):
...
         if not f:
```
Vijay-P, line 227: disable delete feeds too

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -114,9 +114,54 @@ class DataFeed(ABC):
         MapperFactory
     ] = None  # A dict/map of group names to mapper objects for translating group data into db types
 
-    def __init__(self, metadata: FeedMetadata):
+    def __init__(self, metadata: Optional[FeedMetadata] = None):
+        if not metadata:
```
zhill, line 118: These changes make this class no longer an interface definition, but also some implementation, so either need to separate those out or update the doc string for the class to be a regular base class.

tests/unit/anchore_engine/services/policy_engine/engine/feeds/test_sync_utils.py
```diff
@@ -0,0 +1,271 @@
...
+        else:
+            vulnerabilities_provider = GrypeProvider()
+        sync_configs = compute_selected_configs_to_sync(
+            provider=vulnerabilities_provider.get_config_name(),
+            vulnerabilities_config=get_config_for_params(provider, feed_configurations),
+            default_provider_sync_config=vulnerabilities_provider.get_default_sync_config(),
+        )
+        assert set(sync_configs.keys()) == set(expected_to_sync_after_compute)
+        sync_utils_provider = vulnerabilities_provider.get_sync_utils(sync_configs)
+        filtered_configs = sync_utils_provider._get_filtered_sync_configs(sync_configs)
```
zhill, line 186: This should probably be 2 different tests since it is testing 2 different cases and function outputs.

zhill (commented):
Some suggestions and comments for clarity in the code. Overall I think we need to push the abstraction a bit further in the future to group the feeds and data that the provider itself manages so that there is a clear boundary at the provider layer that provides the functionality for the sync, API queries, and data management. But this is a step in the right direction.

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -131,7 +131,11 @@ def sync_feeds(sync=True, force_flush=False):
 def toggle_feed_enabled(feed, enabled):
     if type(enabled) != bool:
         raise BadRequest(message="state must be a boolean", detail={"value": enabled})
-
+    if feed == GRYPE_DB_FEED_NAME:
```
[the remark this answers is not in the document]
Vijay-P: If the enabled flag is not used anywhere other than the download logic, I don't see why not. Can @nightfurys confirm?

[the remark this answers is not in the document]
Vijay-P: The logic required to make this work is just a couple lines diff.

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -793,7 +830,8 @@ def _enqueue_refresh_tasks(db: Session) -> None:
                 errors.append((result.id, err))
         if len(errors) > 0:
             logger.error(
-                f"Failed to create/enqueue {len(errors)}/{len(image_ids)} refresh tasks."
+                f"Failed to create/enqueue %d/%d refresh tasks.",
+                (len(errors), len(image_ids)),
```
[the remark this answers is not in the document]
Vijay-P: thanks for catching this. was re-doing a bunch of log statements to this interpolation style and must have sort of lost track along the way

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -921,21 +956,19 @@ def _process_group_file_records(
                 total_records_updated += 1
                 logger.info(
                     self._log_context.format_msg(
-                        "DB Update Progress: {}/{}".format(
-                            total_records_updated,
-                            group_download_result.total_records,
-                        ),
-                    )
+                        "DB Update Progress: %d/%d"
+                        % (total_records_updated, group_download_result.total_records)
```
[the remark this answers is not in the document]
Vijay-P: I'm really confused. I changed the string format here to the % interpolation style because of feedback from @zburstein that using .format was unsafe. There has been a lot of back-and-forth between folks on the team as to how to format strings, and I don't understand what is preferred or why. Would really appreciate some clarity.

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -748,29 +754,60 @@ def record_count(self, group_name: str, db: Session) -> int:
...
+
+        :return: db record containing feed metadata
+        :rtype: FeedMetadata
+        """
+        db = get_session()
+        try:
+            catalog_client = self._catalog_client
+            records = self._find_match(db)
+            for record in records:
+                catalog_client.delete_document(
```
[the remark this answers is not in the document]
Vijay-P: Not 100% sure I understand the verbiage, but if what you are asking is whether or not we have a way of rolling back the catalog deletion, the answer is no. I can change this to cache the deleted records so that they may be added back on failure.

anchore_engine/services/policy_engine/engine/feeds/feeds.py
Vijay-P: Considering how we aren't expecting folks to extend engine with their own DataFeeds, i think making this a regular base class rather than an interface is fine.

tests/unit/anchore_engine/services/policy_engine/engine/feeds/test_sync_utils.py
Vijay-P: Yeah you're right. I sort of wanted to keep it like this because this test was also supposed to indirectly serve as documentation by sort of mirroring the order of ops that happen when the sync starts. I recognize that's not really ideal. I'll split it up and leave the big docstring in.

anchore_engine/services/policy_engine/api/controllers/feeds.py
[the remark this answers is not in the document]
Vijay-P: Implemented this last night.

anchore_engine/services/policy_engine/engine/feeds/feeds.py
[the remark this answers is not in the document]
zburstein: Yeah we had a discussion with @nightfurys about how we are doing logging and if Im not mistaken she said that using the percent format was the best way to interpolate with our logger. Is there an alternative way that you suggest?

anchore_engine/services/policy_engine/engine/feeds/feeds.py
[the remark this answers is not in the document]
Vijay-P:
```
    def flush_all(self) -> FeedMetadata:
        """
        Flush all groups for the feed, unset the last full sync timestamp, but leave metadata records for feed.

        :return: db record containing feed metadata
        :rtype: FeedMetadata
        """
        db = get_session()
        deleted_records = []
        try:
            records = self._get_db_metadata_records(db)
            for record in records:
                deleted_records.append(record.archive_checksum)
                db.delete(record)
            db.refresh(self.metadata)

            # Remove all groups
            self.metadata.groups = []
            self.metadata.last_full_sync = None
            db.commit()
        except Exception:
            db.rollback()
            raise
        catalog_client = self._catalog_client
        for archive_checksum in deleted_records:
            catalog_client.delete_document(self.__feed_name__, archive_checksum)
        return self.metadata
```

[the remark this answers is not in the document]
Vijay-P: is this better?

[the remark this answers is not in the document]
Vijay-P: committed that change

zhill (approved):

PR #1072, opened 2021-05-26
Support vendor_only flag in grype provider

nightfurys:
This PR adds the logic for filtering based on the vendor_only flag. When this flag is enabled, a vulnerability match will be dropped from the results if the vendor has indicated that the issue won't be fixed

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/providers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/providers.py
@@ -4,7 +4,7 @@ import json
 import time
 import uuid
 from abc import ABC, abstractmethod
-from typing import Dict
+from typing import Dict, List
 
 from sqlalchemy import asc, func, orm
 
@@ -25,6 +25,7 @@ from anchore_engine.services.policy_engine.api.models import (
     Vulnerability as VulnerabilityModel,
     VulnerabilityMatch,
     Artifact,
+    VendorAdvisory,
     ImageVulnerabilitiesReport,
     VulnerabilitiesReportMetadata,
     CvssCombined,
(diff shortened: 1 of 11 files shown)


dspalmer99 (approved):
One nit, but LGTM

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -782,14 +798,35 @@ def get_image_vulnerabilities(
...
+            report = self._create_new_report(image, db_session, use_store)
         else:
             report = self._try_load_report_from_store(image, db_session, use_store)
-            if isinstance(report, ImageVulnerabilitiesReport):
-                return report
-            else:
-                report = ImageVulnerabilitiesReport.from_json(report)
-                return report
+
+        if not isinstance(report, ImageVulnerabilitiesReport):
```
Vijay-P, line 805: why not check that it's a dict instead of checking that it's not a ImageVulnerabilitiesReport

tests/unit/anchore_engine/services/policy_engine/engine/feeds/test_config.py
```diff
@@ -124,8 +124,8 @@ def test_get_selected_configs_to_sync_defaults(provider, test_config, expected):
         pytest.param(
             LegacyProvider,
             {"provider": "legacy", "sync": {"data": {"vulndb": {"enabled": True}}}},
-            {"vulndb"},
-            id="valid-legacy-vulndb",
+            set(),
```
Vijay-P, line 127: why is this invalid now? not blocking comment, just don't understand why this would change.

Vijay-P (approved):
I don't have the strongest grasp around the vuln scanning bit of the code, so my review is sort of discount. Having said that, LGTM.

anchore_engine/services/policy_engine/engine/vulns/providers.py
nightfurys: It's the same result at the end, this flow made more sense in my head: if it's not a report object, then create one first before operating on it

tests/unit/anchore_engine/services/policy_engine/engine/feeds/test_config.py
nightfurys: Because I've removed it from the default config here https://github.com/anchore/anchore-engine/pull/1072/files#diff-407ba06f155493ba97d77fe4c33a2a7cd64efac42dfb1f25d03ec9bb3e01fb60L133

It'll be added to config in the downstream provider that actually support the feed

PR #1076, opened 2021-05-26
Installs Anchore CLI into a virtual environment in /anchore-cli

### LINKED ISSUE #1074: Install the anchore-cli in a virtual env to avoid dependency conflicts with engine
Since Anchore CLI is python, when installed into the system root, it creates dependency conflicts with engine if not managed properly. This creates issues when either engine or cli need to have dependencies updated that can only be detected at image build and test time rather than during individual project testing. This dependency should be broken to simplify testing.

Proposed solution: Anchore CLI will be installed using a virtual environment at /anchore-cli and the docker entry point will setup the paths accordingly to run the anchore-cli script from that location.

zhill:
Fixes #1074 by creating a virtual env in /anchore-cli that isolates the cli requirements from those of anchore-engine so they cannot conflict.

This PR does not change the operation of Anchore Engine itself since the CLI is never invoked by any service. The CLI is included for convenience only so users can exec into a running container and use it to debug.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -181,7 +181,8 @@ RUN set -ex && \
 RUN set -ex && \
     python3 -m venv /anchore-cli && \
     source /anchore-cli/bin/activate && \
-    pip3 install --no-index --find-links=./ /build_output/cli_wheels/*.whl
+    pip3 install --no-index --find-links=./ /build_output/cli_wheels/*.whl && \
+    deactivate
 
 # Perform the anchore-engine build and install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY . /buildsource
 WORKDIR /buildsource
 
 RUN set -ex && \
-    mkdir -p /build_output /build_output/deps /build_output/configs /build_output/wheels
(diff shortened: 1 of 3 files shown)


zhill:
Oops, missed the deactivate call in the dockerfile. I'll update that.

zhill:
I'll look into the CI job failure. Looks like a CI setup issue

robertp (commented):
@zhill could you please make the following change in `scripts/ci/test-functional`? I think this should resolve the CI issue.

```
-        anchore-cli --u admin --p foobar --url http://engine-api:8228/v1 system wait --feedsready '' && \
-        anchore-cli --u admin --p foobar --url http://engine-api:8228/v1 system status && \
+        /anchore-cli/bin/anchore-cli --u admin --p foobar --url http://engine-api:8228/v1 system wait --feedsready '' && \
+        /anchore-cli/bin/anchore-cli --u admin --p foobar --url http://engine-api:8228/v1 system status && \
```

Vijay-P (approved):
Built and tested. Works. We will need to update the docs to reflect the env change.

PR #1077, opened 2021-05-26
Move policy engine models into a common location

dspalmer99:
This PR moves (and renames) the policy engine models into a new `models` dir under common, and moves `common/schemas.py` into that same dir. Other model modules remain unmoved, but in the future would ideally also be centralized in a future update in order to promote code reusability and readability. This is needed for the upcoming release due to some of the changes made to the legacy matching code @nightfurys made during the grype integration.

This PR just moves those two modules; it does not introduce any new code. The other changed files are just updates to the import paths that reference those two modules.


nightfurys (approved):
lgtm! Fine by me to merge if its working and doesn't break anything

PR #1078, opened 2021-05-26
Update grype scanner to use string sbom grype_wrapper function

nightfurys:
PR updates the input to grype wrapper for scanning an image. The scanner now supplies a string sbom to the wrapper instead of a reference to a file


dspalmer99 (approved):
LGTM

PR #1081, opened 2021-05-28
Update group list and creation

zburstein:
This PR makes 3 changes:

1. The feeds update process now updates the GrypeDBFeedMetadata record with the groups contained within the synced db.
2. Uses the vuln provider to build data for the policy_engine list_feeds endpoint. This is needed in order to support grypedb not persisting groups in the same manner as legacy did
3. Adds some tests and restructures the policy engine functional tests for feed sync to be compatible with grype being set as the vuln provider. The default behavior remains testing the legacy workflow, but using the `VULN+PROVIDER` env variable allows the test runner to test the grype path


anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -906,26 +907,84 @@ def _switch_active_grypedb(
         )
         db.add(self.grypedb_meta)
 
-    def _process_group_file_records(
+    @staticmethod
+    def _run_grypedb_sync_task(
```
Vijay-P, line 911: this method got carried over from a rebase. it was deleted in one of my PRs. you should remove it.

anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -69,13 +86,18 @@ def _marshall_feed_response(feed: DbFeedMetadata) -> FeedMetadata:
     i.enabled = feed.enabled
     i.groups = []
 
-    for group in feed.groups:
+    groups = provider.get_feed_groups_detached(feed)
```
Vijay-P, line 89: so you're detaching this because FeedGroupMetadata backpopulates FeedMetadata?

tests/functional/services/policy_engine/feeds_data_tests/test_feed_sync.py
```diff
@@ -102,52 +121,172 @@ def test_feeds_get_schema(self, sync_feeds, schema_validator):
...
+        assert actual_group["status"] == "success"
+        assert actual_group["updated_record_count"] == expected_count
+        assert actual_group["total_time_seconds"] > 0
+
+    @pytest.mark.parametrize(
+        "expected_feed, expected_group, expected_count",
+        build_feed_sync_test_matrix(),
+        ids=idfn,
+    )
+    def test_expected_feed_sync_get_feeds(
```
Vijay-P, line 160: these tests are getting sort of complex. i think each test method would benefit from a short docstring that explains the premise of the test.

Vijay-P (changes requested):
overall LGTM. nit: docstrings in tests (would be nice for next person), probably should delete that extra method in GrypeDBFeed

nightfurys (commented):
Looks good for the most part. Just a few comments about using the abstraction meaningfully

anchore_engine/services/policy_engine/api/controllers/feeds.py
zburstein: I detatch this for easier session management. The calling context only needs to read the attributes so they can be detatched as opposed to passing sessions around. I modeled this off the existing structure that used detatched records

anchore_engine/services/policy_engine/engine/feeds/feeds.py
zburstein: Oh yeah good catch, this got lumped in with one of my changes when I resolved conflicts

anchore_engine/db/entities/policy_engine.py
```diff
@@ -75,8 +75,8 @@ class FeedMetadata(Base, UtilMixin):
     enabled = Column(Boolean, default=True)
 
     @classmethod
-    def get_by_name(cls, name):
-        return FeedMetadata.query.filter(name=name).scalar()
+    def get_by_name(cls, session, name):
+        return session.query(cls).filter(cls.name == name).scalar()
```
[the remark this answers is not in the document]
zburstein: I did that because its a class variable and the class is available in the function but the code was referencing the class by name as opposed to using the cls reference that was available. Should behave the same as what it was before, but I can change it back if you want

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -906,26 +907,84 @@ def _switch_active_grypedb(
...
+    def _set_group_counts(self, db, source_counts: List[RecordSource]) -> None:
+        """
+        Uses grype wrapper to query the source counts in the grypedb
+        casts these source counts to FeedGroupMetadata and saves it to GrypeDBFeedMetadata groups column
+        certain timestamps like the last_sync are not set in this process the first time around because it has not yet
+        been set on the GrypeDBFeedMetadata object
+        """
+        groups = []
+
+        for source in source_counts:
```
Vijay-P, line 937: Perhaps it would be nice to log this information for debug purposes so that one can see the sort of group info that we are populating during the sync?

anchore_engine/db/entities/policy_engine.py
```diff
@@ -163,7 +163,7 @@ class GrypeDBFeedMetadata(Base):
         nullable=False,
     )
     synced_at = Column(DateTime, nullable=True)
-    groups = Column(JSONB)
+    groups = Column(JSONB, default=[])
```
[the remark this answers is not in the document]
zburstein: Since the expectation is that groups will be an array and will need to be iterated over, I figured it would be easier to default it on the db level so that potential loops dont error out if its `None`.  Thought the db level might be a safer and more reliable location in case there are any errors that occur in processes and it centralizes that attribute being set to one location. Want me to do it at object creation?

Vijay-P (approved):

anchore_engine/db/entities/policy_engine.py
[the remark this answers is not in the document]
nightfurys: That would make it a little clearer. SQLAlchemy classes are special in that sense that they carry a lot of context. Reading this code it's not clear to which context you are querying

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -126,6 +142,35 @@ def get_sync_utils(self, sync_configs: Dict[str, SyncConfig]) -> SyncUtilProvide
...
+
+    def get_feed_groups_detached(
+        self, feed: FeedMetadata
+    ) -> Iterable[FeedGroupMetadata]:
+        """
+        Gets a list of all groups consumed by Legacy provider from specified feed in a detached state
+        """
+        if feed.name not in self.__default_sync_config__.keys():
+            raise ValueError(
+                "Incompatible feed passed to Grype provider get_feed_groups_detached function"
```
nightfurys, line 162: (nit) lose the hardcoded "Grype" in the error message, you can use `self.__config__name__` to reflect the accurate name. Generally when I raise these errors, I would like to see what the offending and allowed values are. So for instance 
```
raise ValueError("%s feed is not supported by %s provider, supported feeds are %s", feed.name, self.__config__name, list(self.__default_sync_config__.keys()))
```

PR #1084, opened 2021-05-30
Integrate with unreleased grype binary

nightfurys:
This PR 
- fills all the placeholders waiting for information from grype. It's been tested against a dev binary provided by the toolbox team
- policy-engine vulnerability model changes to better represent NVD references and CVSS scores and utilize the new model in external API
- adds error handling in grype provider
- adds a bunch of unit tests and modifies existing function tests

PR #1086, opened 2021-06-01
Fixes issues caused by policy-engine model update 

nightfurys:
This PR fixes issues introduced by policy engine model update in #1084 
- cvss score access in vulnerabilities gate
- missing nvd references for nvd vulnerabilities. This is redundant but necessary because of logic that operates on nvd references (such as vulnerabilities gate, external API etc)


dakaneye (approved):
changes lgtm visually, didn't test to reproduce

PR #1092, opened 2021-06-02
Update grype wrapper methods and tests to work with the new grype db schema.

dspalmer99:
Updates methods in the grype wrapper to use the new grype v3 schema. Also updates corresponding tests.


anchore_engine/clients/grype_wrapper.py
```diff
@@ -866,19 +869,10 @@ def query_record_source_counts(self, use_staging: bool = False):
...
-                                feed_group
-                            )
-                        )
-                        continue
-
                     record_source = RecordSource(
                         count=result[1],
-                        feed=feed_group[0],
-                        group=feed_group[1],
+                        feed=VULNERABILITIES,
```
dspalmer99, line 874: @nightfurys  Before this change we were getting this value from the fist part of the record_source field, which AFAICT was always `vulnerabilities`. I've set that to this constant here to keep the functionality from changing, but is there something else we want to set this to? `grype`?

I'm assuming once we remove the legacy scanner this field will be removed, so maybe it doesn't matter so much right now.

Vijay-P: Curious what's in the record_source field now. I see that it's still in the table definition.

Vijay-P (approved):
LGTM overall, would like @nightfurys to verify the use of namespace column over record_source.

anchore_engine/clients/grype_wrapper.py
```diff
@@ -866,19 +869,10 @@ def query_record_source_counts(self, use_staging: bool = False):
                 # Transform the results along with the last_synced timestamp for each result
                 output = []
                 for result in results:
```
nightfurys, line 871: (nit) unpacking the tuple here makes for easier reading. It would also raise exceptions sooner if the expected number of elements are not available
```
for group, count in results:
```

anchore_engine/clients/grype_wrapper.py
nightfurys: Let's leave this at vulnerabilities for now

nightfurys (commented):
Need to ensure schema definitions here are inline with grype definitions https://github.com/anchore/grype-db/tree/main/pkg/db/model. Noticed a foreign key association that's not in the original model

nightfurys:
> Need to ensure schema definitions here are inline with grype definitions https://github.com/anchore/grype-db/tree/main/pkg/db/model. Noticed a foreign key association that's not in the original model

Link to grype-db models is an older version. Refer to db schema v3 models here https://github.com/anchore/grype-db/tree/v3/pkg/db/model

anchore_engine/clients/grype_wrapper.py
dspalmer99: Ack, will do, thanks.

dspalmer99:
Per a slack conversation with @nightfurys , I'm going to merge this pr once the CI passes, since it unblocks several things. And in the follow-up PR I will plan to revisit how the grype wrapper joins those the two relevant tables from grype db in sql alchemy.

PR #1094, opened 2021-06-03
Rescan images for vulnerabilities after a successful sync

nightfurys:
Implements the refresh mechanism after a successful grype sync
- updated the queuing logic to batch add images to avoid queue overheads
- policy-engine handler to retrieve refresh tasks from the queue and process
- move the old logic that refreshes images loaded during feed sync into the legacy provider and add a no-op for the grype provider

### DIFF
--- a/anchore_engine/common/models/schemas.py
+++ b/anchore_engine/common/models/schemas.py
@@ -645,3 +645,39 @@ class LocalFeedDataRepoMetadata(JsonSerializable):
         self.download_configuration = download_configuration
         self.download_result = download_result
         self.data_write_dir = data_write_dir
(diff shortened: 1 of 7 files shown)


Vijay-P (approved):
LGTM. I see that the name of the queue is sort of scattered around. Would be nice if we had a central registry for queue names, so there was a single source of truth.

nightfurys:
> LGTM. I see that the name of the queue is sort of scattered around. Would be nice if we had a central registry for queue names, so there was a single source of truth.

Yep, I agree. There needs to be some queue utils in the commons, accessible to all services

PR #1096, opened 2021-06-04
query images_by_vulnerability with grype provider

nightfurys:
- Implements the query using a jsonb query to look up all images that contain the vulnerability ID first. Loads the results into python space and applies the remaining query filters
- Fixes a typo/super-class function invocation
- Transfer fix observed-at timestamp from the source
- Lots of unit tests


dspalmer99 (approved):
LGTM

PR #1097, opened 2021-06-04
Updates to grype rwlock context managers.

dspalmer99:
Testing some changes to the way the rwlock in the grype wrapper is accessed in the read and write context managers.

I have not tested this yet due to time constraints. Sharing here for group debugging purposes, as we were just looking at this on a call.

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -224,7 +224,7 @@ class GrypeWrapperSingleton(object):
         read_lock = self._grype_db_lock.gen_rlock()
 
         try:
-            if read_lock.acquire(blocking=True, timeout=self.LOCK_READ_ACCESS_TIMEOUT):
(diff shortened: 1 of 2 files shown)


nightfurys:
`.acquire` is expecting the timeout to be in millis. Set the `LOCK_READ_ACCESS_TIMEOUT` and `LOCK_READ_ACCESS_TIMEOUT` to 60000. Also there's no need to specify the blocking behaviour. RWLockWrite allows for concurrent reads and serialized write by default, no need to set the `blocking` flag on `.acquire()`

dspalmer99:
@nightfurys I've made those changes, will test them in the AM.

nightfurys (approved):

dspalmer99:
Per the group's discussion, this is going to go in as-is as as to get the immediate issue resolved. Moving lock initialization pre-thread pool creation, as discussed with @zhill in Slack, will happen afterwards in a separate PR.

PR #1098, opened 2021-06-04
Grype provider db session leakage fix.

nightfurys:
Noticed db transactions that were not committed/closed in the report store code (lifted from policy-eval cache manager). Fixed the issue by passing the original db session to all utility functions


dakaneye (approved):
LGTM, as discussed, we should pass the session to each method in ImageVulnerabilitiesStore instead of setting the class var.

PR #1102, opened 2021-06-07
Bootstrap locks for grype wrapper and grype DB sync manager

Vijay-P:
These changes are done to initialize shared locks before the twistd thread pool is started. That way we can ensure that the locks are in shared memory.

- Adds a post_bootstrap lifecycle handler that initializes the GrypeWrapperSingleton.
- Adds an import statement for grypedb_sync.py so that lock in grypedb_sync.GrypeDBSyncLock gets initialized.
  - Class variables are initialized on import, so this should be sufficient.

**Special notes**:

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -129,6 +129,7 @@ class GrypeWrapperSingleton(object):
     def __new__(cls):
         # If the singleton has not been initialized yet, do so with the instance variables below
         if cls._grype_wrapper_instance is None:
+            logger.debug("Initializing Grype wrapper instance.")
             # The singleton instance, only instantiated once outside of testing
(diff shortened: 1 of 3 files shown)


anchore_engine/clients/grype_wrapper.py
```diff
@@ -125,6 +125,7 @@ class GrypeWrapperSingleton(object):
     def __new__(cls):
         # If the singleton has not been initialized yet, do so with the instance variables below
```
dspalmer99, line 130: Could you add a comment here for future maintainers noting the where and why of this having to be created at bootstrap?

nightfurys (approved):
lgtm

dspalmer99 (approved):
One minor comment requested, but otherwise LGTM

PR #1103, opened 2021-06-08
Use api model for grype groups and use provider for toggle feed enabled endpoint

zburstein:
Abandons the use of the legacy FeedGroupMetadata class for persisting groups in grype. Also uses the provider to toggle the feed

### DIFF
--- a/anchore_engine/services/policy_engine/api/controllers/feeds.py
+++ b/anchore_engine/services/policy_engine/api/controllers/feeds.py
@@ -1,4 +1,3 @@
-import typing
 from dataclasses import asdict
 
 from flask import jsonify
@@ -26,7 +25,6 @@ from anchore_engine.services.policy_engine.engine.feeds.sync_utils import (
 )
 from anchore_engine.services.policy_engine.engine.tasks import FeedsUpdateTask
 from anchore_engine.services.policy_engine.engine.vulns.providers import (
-    VulnerabilitiesProvider,
     get_vulnerabilities_provider,
 )
 from anchore_engine.subsys import logger as log
@@ -51,12 +49,9 @@ def list_feeds(refresh_counts=False):
     return jsonify([feed.to_json() for feed in provider.get_feeds()])
 
 
-def _marshall_feed_response(
-    provider: VulnerabilitiesProvider, feed: DbFeedMetadata
-) -> FeedMetadata:
+def _marshall_feed_response(feed: DbFeedMetadata) -> FeedMetadata:
     """
-    Marshalls FeedMetadata record obtained from the db into the api model
-    Calls _marshall_group_response to build the groups for the specified vuln provider
+    Old method for marshaling a feed. Currently being replaced by workflows driven by providers
     """
     if not feed:
         return ValueError(feed)
@@ -69,9 +64,7 @@ def _marshall_feed_response(
     i.enabled = feed.enabled
     i.groups = []
 
-    groups = provider.get_feed_groups_detached(feed)
-
-    for group in groups:
+    for group in feed.groups:
         i.groups.append(_marshall_group_response(group))
 
     return i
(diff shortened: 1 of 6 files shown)


anchore_engine/services/policy_engine/api/controllers/feeds.py
```diff
@@ -153,23 +130,23 @@ def sync_feeds(sync=True, force_flush=False):
...
+        return jsonify(feed.to_json()), 200
 
-        return jsonify(updated), 200
-    except AnchoreApiError:
-        session.rollback()
-        raise
+    except InvalidFeed:
+        raise BadRequest(
+            message="Feed not enabled on configured vulnerability provider",
+            detail={"feed": feed, "configured_provider": provider.__config__name__},
```
nightfurys, line 146: use `provider.get_config_name()` The error message is a bit confusing. It should be something to the effet of "Feed not supported"

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -984,45 +987,60 @@ def _set_group_counts(self, db, source_counts: List[RecordSource]) -> None:
...
     def _update_group_timestamps(self, db):
         """
         Updates the tiemstamps
         """
-        groups = [FeedGroupMetadata(**group) for group in self.grypedb_meta.groups]
 
-        for group in groups:
-            group.last_sync = self.grypedb_meta.synced_at
-            group.last_update = self.grypedb_meta.last_updated
+        if self.grypedb_meta.groups and isinstance(self.grypedb_meta.groups, List):
```
nightfurys, line 1022: instance type check should be for `list`, not `List`

anchore_engine/services/policy_engine/engine/feeds/feeds.py
```diff
@@ -1255,6 +1259,21 @@ def _sync_grype_groups(
...
+
+            results.append(
+                GroupSyncResult(
+                    group=source.group,
+                    updated_record_count=updated_count,
+                    total_time_seconds=total_time_seconds,
+                    status="success",
+                )
+            )
+
```
nightfurys, line 1276: If this block is executed in the happy path, it should be moved into try possibly before the db.commit()

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -67,13 +64,17 @@
...
 from anchore_engine.subsys import logger, metrics
-from anchore_engine.utils import rfc3339str_to_datetime, timer
+from anchore_engine.utils import timer
 
 from .dedup import get_image_vulnerabilities_deduper, transfer_vulnerability_timestamps
 from .scanners import GrypeScanner, LegacyScanner
 from .stores import ImageVulnerabilitiesStore, Status
 
 
+class InvalidFeed(Exception):
```
nightfurys, line 74: (nit) Not necessary for now but these exceptions should be defined in a common place so others can import them without importing the entire provider module. We should also have a base class with common behaviour and extend it. Otheriwse there's little point in raising custom exception

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -180,6 +223,30 @@ def update_feed_group_counts(self) -> None:
         """
         ...
 
+    def toggle_feed_enabled(
```
nightfurys, line 226: (nit) This is not the most descriptive name. I know it's copied over. How about `update_feed_status` or something like it

nightfurys (commented):
Nice refactoring all around, feed responses from providers look much better!

Vijay-P (approved):
Nice work. I agree with Swathi's comments.

anchore_engine/services/policy_engine/engine/vulns/providers.py
zburstein: Agreed. One benefit to this right now is that I can distinguish it from a ValueError in the controller and raise a more specific BadRequest error

nightfurys (approved):
lgtm

PR #1104, opened 2021-06-08
Policy evaluation cache manager updates

nightfurys:
PR refactors the logic that determines if the image vulnerabilities have been updated to the providers. The factors contributing to the decision are provider dependent
- legacy provider uses feeds last synced timestamp 
- grype provider uses the stored report and it's current state

to decide whether image vulnerabilities have been modified since the last policy evaluation

### DIFF
--- a/anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
+++ b/anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
@@ -501,23 +501,15 @@ class EvaluationCacheManager(object):
     def _inputs_changed(self, cache_timestamp):
         # A feed sync has occurred since the eval was done or the image has been updated/reloaded, so inputs can have changed. Must be stale
         db = get_session()
-        # TODO: zhill - test more
-        feed_group_updated_list = [
-            group.last_sync
-            if group.last_sync is not None
-            else datetime.datetime.utcfromtimestamp(0)
-            for feed in get_all_feeds(db)
-            for group in feed.groups
-        ]
-        feed_synced = (
-            max(feed_group_updated_list) > cache_timestamp
-            if feed_group_updated_list
-            else False
-        )
 
         image_updated = self.image.last_modified > cache_timestamp
 
-        return feed_synced or image_updated
+        return (
+            image_updated
+            or get_vulnerabilities_provider().is_image_vulnerabilities_updated(
+                image=self.image, db_session=db, since=cache_timestamp
+            )
+        )
 
     def _should_evaluate(self, cache_entry: CachedPolicyEvaluation):
         if cache_entry is None:
--- a/anchore_engine/services/policy_engine/engine/vulns/providers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/providers.py
@@ -50,6 +50,7 @@ from anchore_engine.services.policy_engine.engine.feeds.config import (
 from anchore_engine.services.policy_engine.engine.feeds.db import (
     get_all_feeds_detached,
     get_feed_detached,
+    get_all_feeds,
 )
 from anchore_engine.services.policy_engine.engine.feeds.feeds import (
     GrypeDBFeed,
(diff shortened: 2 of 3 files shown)


anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
```diff
@@ -501,23 +501,15 @@ def save(self, result):
...
-            max(feed_group_updated_list) > cache_timestamp
-            if feed_group_updated_list
-            else False
-        )
 
         image_updated = self.image.last_modified > cache_timestamp
 
-        return feed_synced or image_updated
+        return (
+            image_updated
```
dakaneye, line 508: it looks like the order of the returned values was changed, was that on purpose?

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -914,6 +924,27 @@ def update_feed_group_counts(self) -> None:
...
+        """
+        # TODO: zhill - test more
+        feed_group_updated_list = [
+            group.last_sync
+            if group.last_sync is not None
+            else datetime.datetime.utcfromtimestamp(0)
+            for feed in get_all_feeds(db_session)
+            for group in feed.groups
+        ]
+        return (
```
dakaneye, line 944: nit (feel free to ignore): my opinion here is that this is a little hard to read, could be improved with some code structure:
```
if feed_group_updated_list:
  return max(feed_group_updated_list) > since
else:
  return False
```
Furthermore, I feel the `since` variable brings into some question about the method naming. Instead of determining whether the image vulnerabilities _are_ updated, you're making sure they've been updated `since` a particular time. 

Perhaps the method could be renamed to something like: `are_image_vulns_updated_recently` or something more specific.

anchore_engine/services/policy_engine/engine/vulns/providers.py
```diff
@@ -914,6 +924,27 @@ def update_feed_group_counts(self) -> None:
         """
         DataFeeds.update_counts()
 
+    def is_image_vulnerabilities_updated(
+        self, image: Image, db_session, since: datetime.datetime
+    ):
+        """
+        Image vulnerabilities *may be* updated after feeds are synced. This function determines if the image vulnerabilities
```
dakaneye, line 931: these comments are great!

anchore_engine/services/policy_engine/engine/vulns/stores.py
```diff
@@ -108,6 +109,38 @@ def fetch(self, session):
...
+        """
+        Looks up a report for the image and returns True if the report is newer than the timestamp.
+        If the report was generated before the since timestamp, the logic checks for report status and returns True
+        if the report needs a refresh, False otherwise
+
+        """
+        db_record = (
+            session.query(DbImageVulnerabilities)
+            .filter_by(account_id=self.image.user_id, image_digest=self.image.digest)
+            .one_or_none()
```
dakaneye, line 122: Is there only one DbImageVulnerabilities record for a given digest? Or should this be iterating through a list?  This question might not make sense if I'm totally out of my depth for this feature.

It seems based on the naming like there'd be more records to check. Perhaps more sorting or filtering is required? Perhaps not...

dakaneye (approved):
Looks generally ok, but I'm not totally caught up on the feature itself. See comments.

anchore_engine/services/policy_engine/api/controllers/synchronous_operations.py
nightfurys: Yep, both the checks are relatively light weight. That said, checking if the image was updated may be slightly faster relative to the other check which is now delegated to the provider. Re-ordered the checks in the perf order so that the logic can bail out quickly if the 1st check fails

anchore_engine/services/policy_engine/engine/vulns/providers.py
nightfurys: Ack. This is the legacy code that was copied over. As for the naming, I was going for the s3 style get-if-modified-since format. I might go ahead with this merge but will give the naming more thought

anchore_engine/services/policy_engine/engine/vulns/stores.py
nightfurys: Good question, there is only one report per digest per account. That is enforced in the persistence model with a compound primary key consisting of account_id and image_digest

PR #1108, opened 2021-06-08
Grype query vulnerability updates

dspalmer99:
This PR contains updates @nightfurys and myself made to the /query/vulnerabilities grype mapping logic. It could use a little bit of refactoring still, but the response object from the grype provider backing for that endpoint is now more like that returned by the legacy provider.


nightfurys (approved):
lgtm, let's add a ticket for 3.2 to rip out the dicts and replace them with model objects

PR #1110, opened 2021-06-09
Add grype policy engine functional tests

zburstein:
Sets up exisitng policy engine functional tests to be able to test grype as the vuln provider in addition to legacy


Toure (approved):
LGTM, just one nit, can you squash a couple of the commits down where they make sense.

PR #1113, opened 2021-06-11
Use static string for feed in vuln mappers

zburstein:
Per @nightfurys suggestion, changing the feed associated with vuln in mapper to static string `grypedb`


nightfurys (approved):
lgtm. Intent is to avoid import loops here. May want to look into refactoring the name into commons if we want to keep this fee type around

PR #1116, opened 2021-06-14
Documentation for grype integration tech preview

dspalmer99:
Engine docs updates for grype integration tech preview

### DIFF
--- a/docs/content/docs/grype/_index.md
+++ b/docs/content/docs/grype/_index.md
@@ -1,0 +1,46 @@
+---
+title: "Tech Preview - Grype Integration"
+linkTitle: "Access"
+weight: 1
+---
+
+Anchore Engine 0.10 includes a limited-functionality tech preview of an integration with [Grype](https://github.com/anchore/grype)
+for vulnerability scanning. This integration will replace the legacy vulnerability scanner in a future version of Anchore Engine,
+but is provided in this release in a preview capacity so users can try it out.
+
+***Note:*** This tech preview is not intended for use in production environments. It should be installed in sandbox exnvironments,
+and is strictly provided here so users can get start to use it and get a sense for the functionality. It may not include all of the
+functionality of the legacy scanner. Please report any issues found with it on the
+[anchore-engine Github repo](https://github.com/anchore/anchore-engine/issues).
+
+### Installing
+As of 0.10.0, Anchore Engine can be configured to use either the legacy or grype vulnerability scanner. It is not possible to run
+both vulnerability scanners at the same time. This configuration is picked up at bootstrap, and cannot be changed ona running system. 
+
+#### Running with docker-compose
+1. Install or update to Anchore Engine 0.10.0.
+2. Add the following environment variable to the policy engine container section of the docker compose file:
+
+
+    policy-engine:
+      ...
+      environment:
+      ...
+      - ANCHORE_VULNERABILITIES_PROVIDER=grype
+
+3. Redeploy the services.
+
+#### Running with helm
+1. Install or update to Anchore Engine 0.10.0.
+2. Update the following value in `stable/anchore-engine/values.yaml`:
+
+
+    anchorePolicyEngine
+      ...
+      vulnerabilityProvider: grype
+3. Redeploy the services
+
+After making the relevant change above and redeploying, the system will start up with the grype vulnerability scanner enabled and will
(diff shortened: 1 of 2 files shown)


docs/content/docs/grype/_index.md
```diff
@@ -0,0 +1,46 @@
...
+and is strictly provided here so users can get start to use it and get a sense for the functionality. It may not include all of the
+functionality of the legacy scanner. Please report any issues found with it on the
+[anchore-engine Github repo](https://github.com/anchore/anchore-engine/issues).
+
+### Installing
+As of 0.10.0, Anchore Engine can be configured to use either the legacy or grype vulnerability scanner. It is not possible to run
+both vulnerability scanners at the same time. This configuration is picked up at bootstrap, and cannot be changed ona running system. 
+
+#### Running with docker-compose
+1. Install or update to Anchore Engine 0.10.0.
```
dspalmer99, line 21: I'd like to get confirmation that these configuration options are correct, particularly for helm as I've not run it that way yet. (I based this off of what I saw in your charts PR, @Btodhunter

docs/content/docs/grype/_index.md
```diff
@@ -0,0 +1,46 @@
...
+      ...
+      environment:
+      ...
+      - ANCHORE_VULNERABILITIES_PROVIDER=grype
+
+3. Redeploy the services.
+
+#### Running with helm
+1. Install or update to Anchore Engine 0.10.0.
+2. Update the following value in `stable/anchore-engine/values.yaml`:
```
zhill, line 35: Users won't use the git repo to get the chart. They'll create their own values.yaml. I think here we should link to the chart README on how to set this since the specifics are part of the chart

docs/content/docs/grype/_index.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+    anchorePolicyEngine
+      ...
+      vulnerabilityProvider: grype
+3. Redeploy the services
```
zhill, line 41: `helm upgrade`

zhill (changes requested):
Suggestions for chart/helm instructions

docs/content/docs/grype/_index.md
Vijay-P: this is right. the other variable is ANCHORE_GRYPE_DB_URL, but that only needs to be configured differently for enterprise

docs/content/docs/grype/_index.md
```diff
@@ -0,0 +1,46 @@
...
+
+After making the relevant change above and redeploying, the system will start up with the grype vulnerability scanner enabled and will
+sync the latest version of grype db. Note that legacy feeds will no longer be synced while grype is configured. All vulnerability data
```
Vijay-P, line 44: We should provide an example configuration here.

docs/content/docs/install/configuration/feed_configuration.md
```diff
@@ -43,6 +43,11 @@ feeds:
   read_timeout_seconds: 60
 ```
 
+***Note:*** Engine now has a new feed for Grype. This feed will only be synced if the
+[tech preview Grype vulnerability scanner]({{< ref "/docs/grype/_index.md" >}}) is enabled. Furthermore, if it is
```
Vijay-P, line 47: hyperlink is broken. check that you're using the right MD syntax

Vijay-P: the example that i mentioned before can also be put here.

docs/content/docs/install/configuration/feed_configuration.md
```diff
@@ -43,6 +43,11 @@ feeds:
   read_timeout_seconds: 60
 ```
 
+***Note:*** Engine now has a new feed for Grype. This feed will only be synced if the
```
Vijay-P, line 46: the old style config was moved to a new location. need to update the examples here to reflect that location. `services.policy_engine.vulnerabilities.sync.data`

Vijay-P (changes requested):
great work! Config changes introduced with grype integration need to be documented.

dspalmer99:
@zhill @Vijay-P I've addressed all feedback, and fixed some additional typos. Can I get another look from you both on this. Swathi's PR has merge priority, but I'd like to get this one in today as well.

zhill (approved):
There are some caveats we need to add for unsupported functionality in this mode, but I think that can be added as a follow-up to this either in release notes directly or embedded in this page.

PR #1117, opened 2021-06-14
Revert "[wip] Adds cpe-based matching for alpine to support upstream vulns"

Toure:
This reverts anchore/anchore-engine#1064 to undo a mistaken push to the main branch of a pull request that is currently under review. The pull request which will replace this is #1118.


Vijay-P (approved):

zhill (commented):
Just please add a comment on why this revert is done--accidental. Rather than some inherent aspect of the code or a problem (though there may be some :) ) Otherwise will be confusing to understand in the future.

PR #1118, opened 2021-06-14
Adds cpe-based matching for alpine apks to support upstream vulns

zhill:
Creates a new CpeMatcher abstraction with 2 implementations: one for non-os only, and one that supports selective distros. The current distro selector is only for "alpine". These are implemented in the LegacyScanner only, as they are not needed in the GrypeScanner

### DIFF
--- a/anchore_engine/db/entities/policy_engine.py
+++ b/anchore_engine/db/entities/policy_engine.py
@@ -4,6 +4,7 @@ import json
 import re
 import zlib
 from collections import namedtuple
+import typing
 from typing import List
 
 from sqlalchemy import (
@@ -1123,6 +1124,28 @@ class VulnDBMetadata(Base):
                 return "CVE-" + cve_id_col[0]
 
         return self.name
+
+    @property
+    def referenced_cves(self) -> typing.List[str]:
+        """
+        Returns all referenced CVE IDs from the record, may be zero, one, or many.
+
+        :return: List[Str]
+        """
+        ids = set()
+
+        # References is a list of objects each with a potential "source" key
+        for reference in [
+            ref for ref in self.references if ref.get("source") == "CVE ID"
+        ]:
+            url = reference.get("url")
+            if url:
+                # findall should return a single id list ['2020-11989']
+                cve_id_col = re.findall(r"\=(\d+\-\d+)", url)
+                if cve_id_col:
+                    ids.add("CVE-" + cve_id_col[0])
+
+        return list(ids)
 
     def _get_max_cvss_v3_metric_nvd(self):
         cvss_v3 = None
@@ -2413,8 +2436,17 @@ class ImageCpe(Base):
     )
 
     def __repr__(self):
-        return "<{} user_id={}, img_id={}, name={}>".format(
-            self.__class__, self.image_user_id, self.image_id, self.name
+        return (
+            "<{} user_id={}, img_id={}, name={}, version={}, type={}, path={}>".format(
+                self.__class__,
+                self.image_user_id,
+                self.image_id,
+                self.name,
+                self.name,
+                self.version,
+                self.pkg_type,
+                self.pkg_path,
+            )
         )
 
     def fixed_in(self):
@@ -2742,6 +2774,9 @@ class Image(Base):
             return self.distro_name + ":" + self.distro_version
         else:
             return None
+
+    def distro_namespace_obj(self):
+        return DistroNamespace.for_obj(self)
 
     def get_packages_by_type(self, pkg_type):
         db = get_thread_scoped_session()
--- a/anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
@@ -1,0 +1,333 @@
+from typing import List, Tuple
+
+from sqlalchemy.orm import Session
+from collections import namedtuple
+
+from anchore_engine.db import (
+    NvdV2Metadata,
+    CpeV2Vulnerability,
+    Image,
+    get_thread_scoped_session,
+    ImageCpe,
+    VulnDBCpe,
+    Vulnerability,
+    DistroNamespace,
+    ImagePackage,
+)
+from anchore_engine.services.policy_engine.engine.vulns.cpes import (
+    FuzzyCandidateCpeGenerator,
+    dedup_cpe_vulnerabilities,
+)
+from anchore_engine.services.policy_engine.engine.vulns.db import (
+    CpeDBQueryManager,
+    db_result_tuples_to_list,
+)
+from anchore_engine.subsys import logger
+from anchore_engine.utils import timer
+
+# Match class to bind the image and vuln sides and ease comparisons and hashing
+CpeMatch = namedtuple("CpeMatch", ["image_cpe", "vuln_cpe"])
+
+
+def cve_ids_for_vuln_record(vuln: Vulnerability) -> List[str]:
+    """
+    Get CVE Id for vuln from alpine record
+
+    :param vuln:
+    :return: list of CVE ids associated with the vulnerability record
+    """
+
+    if vuln.id.startswith("CVE-"):
+        ids = [vuln.id]
+    else:
+        ids = []
+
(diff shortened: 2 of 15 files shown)


anchore_engine/db/entities/policy_engine.py
```diff
@@ -1124,6 +1125,26 @@ def normalized_id(self):
...
+    def referenced_cves(self) -> typing.List[str]:
+        """
+        Returns all referenced CVE IDs from the record, may be zero, one, or many.
+
+        :return: List[Str]
+        """
+        ids = set()
+
+        # References is a list of objects each with a potential "source" key
+        for entry in [k for k in self.references if k.get("source") == "CVE ID"]:
```
dakaneye, line 1138: nit: `k` variable could have a more descriptive name.

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
```diff
@@ -0,0 +1,330 @@
...
+    db_result_tuples_to_list,
+)
+from anchore_engine.subsys import logger
+from anchore_engine.utils import timer
+
+# Match class to bind the image and vuln sides and ease comparisons and hashing
+CpeMatch = namedtuple("CpeMatch", ["image_cpe", "vuln_cpe"])
+
+
+def cve_ids_for_vuln_record(vuln: Vulnerability) -> List[str]:
```
dakaneye, line 32: is this type hint for the method return correct?

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
```diff
@@ -0,0 +1,330 @@
...
+    Execute the filtering functionality itself on the sets
+
+    :param image_distro:
+    :param matches: match list to filter
+    :return: filtered match list
+    """
+
+    secdb_matched_cves = db_manager.matched_records_for_namespace(image_distro, matches)
+
+    logger.spew("Secdb matched cves %s", secdb_matched_cves)
```
dakaneye, line 73: side note: spew is something that my JSON logging PR removes in favor of debug. Not saying that was the right call or anything, but food for thought.

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
dakaneye: seems like it should be List[set(str)]

anchore_engine/services/policy_engine/engine/vulns/db.py
```diff
@@ -0,0 +1,259 @@
...
+            )
+
+            result = _db_query_wrapper(qry, get_all=True)
+            logger.spew("Raw result = %s", str(result))
+            vuln_ids.extend(result)
+
+        logger.spew("Found cve id matches for %s", vuln_ids)
+        return vuln_ids
+
+    def query_nvd_cpe_matches(
```
dakaneye, line 161: This method looks like a good opportunity for an integration test.

dakaneye (approved):
I didn't see any blocking issues, and was glad to see unit test coverage. This also strikes me as an opportunity for a functional/end-to-end test to ensure this feature works as expected.

dspalmer99 (commented):
A few questions, but overall LGTM. Putting this as "Comment" so as not to block the merge.

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
```diff
@@ -0,0 +1,330 @@
...
+
+        logger.spew("distro cpes: %s", os_pkg_cpe_mappings)
+
+        os_cpes = [cpe for pkg, cpe in os_pkg_cpe_mappings]
+        # Get the matches
+        matches = self._query_cpe_matches(os_cpes)
+
+        logger.spew(
+            "pre-filter cpe distro findings: %s",
+            [(x.image_cpe, x.vuln_cpe.vulnerability_id) for x in matches],
```
dspalmer99, line 306: Similar to Sam's comment above, single-letter variables are harder to read.

dspalmer99, line 306: Similar to Sam's comment above, single-letter variables are harder to read.

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
zhill: that is a constructor call to create a list from a set from the list (basically does a dedup pass), but the result is a list of strings.

zhill:
```
l1 = [1,2,3,3,4]
list(set(l1))
[1, 2, 3, 4]
```

anchore_engine/services/policy_engine/engine/vulns/cpes.py
```diff
@@ -0,0 +1,181 @@
...
+    # all avenues of comparison have been depleted, the two cpes are same for all practical purposes
+    return 0
+
+
+class FuzzyCandidateCpeGenerator:
+    """
+    A generator for CPEs from package metadata that generates candidates predictively
+    """
+
+    embedded_semver_regex = re.compile(r".*([0-9]+\.[0-9]+\.[0-9]+).*")
```
dspalmer99, line 69: Do we need to worry about version numbers with more than three places?

anchore_engine/services/policy_engine/engine/vulns/cpes.py
```diff
@@ -0,0 +1,181 @@
...
+        else:
+            pkg_path = package.pkg_path
+
+        for name in names:
+            # for vendor in vendors:
+            c = ImageCpe()
+            c.name = name
+            c.version = package.version
+            c.pkg_type = package.pkg_type
+            c.cpetype = "a"
```
dspalmer99, line 93: Just curious, what does the "a" mean?

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
zhill: Ack. I think we need a level of verbosity for development. We wouldn't ask a user to enable that log level due to the amount of data it outputs, but provides enough detail for dev usage. But duly noted and we can update consistently when we make the broader change.

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    Key generator that uses only product and version, expects either an ImageCpe or a CpeVuln class
+
+    :param cpe:
+    :return:
+    """
+    # Name is an alias for product
+    try:
+        return (cpe.product, cpe.version)
+    except AttributeError:
```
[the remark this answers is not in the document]
zhill: If neither exists, it is ok for the exception to escape because a bigger problem is present

anchore_engine/db/entities/policy_engine.py
zhill: fixed

anchore_engine/services/policy_engine/engine/vulns/cpe_matchers.py
zhill: fixed

anchore_engine/services/policy_engine/engine/vulns/cpes.py
zhill: In cpes there are "h" - hardware, "o" - OS, and "a" - applications

anchore_engine/services/policy_engine/engine/vulns/cpes.py
zhill: This is logic re-used from the other cpe generation code in loaders.py for fuzzy_go. This is used to add a version to the candidate list, not remove the exact version. So we can improve it to support 4-part versions, but the intent is to handle the case where the raw version has extra stuff and we want to extract a simple semver as a match candidate since sometimes the CVE is against a simplified version (e.g. no alpha, or build info). I'll update it to include any number of numeric fields.

PR #1121, opened 2021-06-14
Catch no active grype db when listing feeds

zburstein:
The feed record for grype db is created whether it is successful or not so the provider needs to handle a situation in which the FeedMetadata record exists but there is no active GrypeDB


nightfurys (approved):
lgtm

PR #1123, opened 2021-06-15
Fix logging bug in VulnerabilityMatchTrigger

Vijay-P:

Log statement bug was introduced on May 5 and may cause whole gate to fail in the case that you're filtering on vendor cvss v3 impact score in the vuln match trigger and the filter criteria are not met.

**Special notes**:


zhill (changes requested):
Can you just change to use the logger.debug("some stuff %s", arg, arg, arg) flow instead? Replace each {} with the %s and remove the .format() call and pass the formatting vars as arguments directly. That is generally safer.

nightfurys (approved):
lgtm, unit tests ftw!

nightfurys:
Approving this fix for 3.1 since it's a regression caused by refactoring

zhill (approved):
lgtm

PR #1125, opened 2021-06-17
Grype stage db fix

dspalmer99:
This PR updates the stage and promote flow in the grype wrapper (and the code that calls it upstream) to only stage dbs from the feed sync, and downstream of that only promote them (without restaging them) from the grype db sync.

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -557,26 +557,37 @@ class GrypeWrapperSingleton(object):
             )
         return
 
-    def stage_grype_db_update(
+    def update_grype_db(
         self,
         grype_db_archive_local_file_location: str,
         archive_checksum: str,
         grype_db_version: str,
+        use_staging: bool = False,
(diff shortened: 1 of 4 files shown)


anchore_engine/clients/grype_wrapper.py
```diff
@@ -585,12 +596,27 @@ def stage_grype_db_update(
...
+                self._grype_db_version = grype_db_version
+                self._grype_db_session_maker = latest_grype_db_session_maker
+
+                logger.info(
+                    "Production grype_db updated to archive checksum %s",
+                    archive_checksum,
+                )
+
+            # Return the engine metadata as a data object
+            return self.get_grype_db_engine_metadata(use_staging=use_staging)
```
Vijay-P, line 619: I think you can get rid of STAGED_GRYPE_DB_NOT_FOUND_ERROR_MESSAGE

Vijay-P (approved):
Have not tested, but changes LGTM.

zburstein (approved):
lgtm

PR #1130, opened 2021-06-20
Support nvd matching for alpine os packages in grype integration

nightfurys:
This PR 
- conditionally saves os package cpes to policy-engine when grype is the provider
- updates sbom mappers to include cpes for apkg type packages
- updates sbom mappers for remaining non-os package types and default to cpe mapper if no pre-defined mappers are found 

Issue with saving os package cpes in legacy mode
Policy engine loaders don't save cpe data for os packages. While that's an easy fix, it will impact the legacy provider. The persisted cpes may create additional matches against feeds that use cpe matching. Dedup logic would filter out the ones that are matched against os or non-nvd feeds. For the cpe matches that don't get deduped, there may be false positives as the legacy provider's matching logic is not comprehensive

### DIFF
--- a/anchore_engine/services/policy_engine/engine/loaders.py
+++ b/anchore_engine/services/policy_engine/engine/loaders.py
@@ -18,7 +18,10 @@ from anchore_engine.db import (
 from anchore_engine.subsys import logger
 from anchore_engine.util.rpm import split_rpm_filename
 from anchore_engine.common.helpers import safe_extract_json_value
-
+from anchore_engine.services.policy_engine.engine.feeds.config import (
+    get_provider_name,
+    get_section_for_vulnerabilities,
+)
 
 # this is a static mapping of known package names (keys) to official cpe names for each package
 nomatch_inclusions = {
@@ -527,7 +530,10 @@ class ImageLoader(object):
         if pkgs:
             return pkgs, handled_pkgtypes
         else:
+            # todo is this even correct?
             logger.warn("Pkg Allinfo not found, reverting to using pkgs.all")
+
+        # below logic doesn't do anything, ImagePackage instances are created and dropped
 
         all_pkgs = package_analysis_json["pkgs.all"]["base"]
         all_pkgs_src = package_analysis_json["pkgs_plus_source.all"]["base"]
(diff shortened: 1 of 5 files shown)


nightfurys:
Tagging along a fix in database lookup for /query/vulnerabilities path in the PR.

@dspalmer99  can you take a look https://github.com/anchore/anchore-engine/pull/1130/commits/e2468d078ae426ac17e848fd44d384d9a0bc7ad5

dspalmer99:
@nightfurys That commit looks good to me, thanks for fixing that. Will look at the rest of the PR too.

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -383,6 +400,9 @@ def to_engine(
     "centos": DistroMapper(
         engine_distro="centos", grype_os="centos", grype_like_os="fedora"
     ),
+    "busybox": DistroMapper(
+        engine_distro="busybox", grype_os="busybox", grype_like_os=""
```
dspalmer99, line 404: Double-checking, is the empty string for `grype_like_os` here and on line 432 intentional?

anchore_engine/services/policy_engine/engine/loaders.py
```diff
@@ -1177,31 +1183,69 @@ def load_cpes_from_syft_output_with_fallback(self, analysis_json, image):
...
+    def extract_syft_cpes_for_os_packages(self, allcpes, package_dict, image):
+        """
+        Utility function for parsing cpes for os packages only. Mostly a duplicate of extract_syft_cpes with the exception of pkg_type and pkg_path.
+        For os packages pkg_path is always pkgdb. pkg_type is picked up from the data instead of passing it as a function argument
+
+        """
+        cpes = []
+        for pkg_key, pkg_json_str in package_dict.items():
+            pkg = safe_extract_json_value(pkg_json_str)
+            pkg_path = f"pkgdb/{pkg_key}"  # os package location is always pkgdb. this is a deviation to tie a package with it's cpes
```
dspalmer99, line 1241: Is there ever a scenario where `pkg_key` could be None?

anchore_engine/services/policy_engine/engine/loaders.py
```diff
@@ -527,8 +530,11 @@ def load_and_normalize_packages(self, package_analysis_json, image_obj):
         if pkgs:
             return pkgs, handled_pkgtypes
         else:
+            # todo is this even correct?
```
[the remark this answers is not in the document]
nightfurys: Nope, it's dead code but I didn't want to introduce changes other than strictly necessary

anchore_engine/services/policy_engine/engine/loaders.py
nightfurys: That should never be the case, pkg_key and pkg_json_str are the key-value pairs in a dictionary. Granted that doesn't prevent it from a null pointer reference, but worst case scenario only one such reference will exist in the dictionary. And even in that case, the logic won't bail

dspalmer99 (approved):
Ack, thanks @nightfurys . LGTM

PR #1132, opened 2021-06-23
add flag to allow disable package filtering by ownership

### LINKED ISSUE #1122: Analyzer config flag to control package-ownership filtering
In 0.9.4 the system started filtering out packages that were detected to be owned by other packages (e.g. a python package installed by an RPM using yum or dnf). This works well at removing false positive matches due to distro vendors updating the RPM version but not updating the "owned" library version when doing backports. However, if a user is building and installing their own RPMs not maintained by the vendor then this can cause issues since the RPM is not managing the backports and the "owned" packages should be scanned for vulnerabilities as independent packages using upstream NVD or other non-distro security data.

Proposed solution is to include a flag in the configuration to allow the operator to set the behavior of the system at deployment time to disable the filtering. It will be enabled by default.

dakaneye:
fixes #1122


zhill (approved):
lgtm

dspalmer99 (approved):
LGTM

dspalmer99:
Merging so we can get into an RC

PR #1140, opened 2021-07-01
Refactor object_store.drivers.swift.SwiftObjectStorageDriver._check_creds

0xDEC0DE:
It is possible on errors for the error response object to not contain an `http_status` attribute, so refactor the error handler so that this attribute is not required.

There appear to be no unit tests for these, so no tests are provided.


zhill (approved):
lgtm. thanks!

zburstein:
@0xDEC0DE Thanks for submitting the PR! The pipeline is failing but its not because of your change. There was an issue with a deadlock in our integration tests taht has since been fixed https://github.com/anchore/anchore-engine/pull/1152. The workflow should pass once the PR is synced up with master

zburstein (approved):

PR #1147, opened 2021-07-19
General cleanup of imports.

Toure:
This change contains:
* import sorting for tests 
* removal of unused imports
* along with a version bump to psycopg2-binary
* enabled new pylint rules that check for circular imports, unused imports, import shadowing, reimports
* removal of circular imports
* removal of reimports

Signed-off-by: Toure Dunnon <toure.dunnon@anchore.com>

**Special notes**:


dspalmer99 (approved):
As far as I can tell, this looks good. There are too many files changed here to thoroughly manually review, but I skimmed through and nothing jumped out at me. More importantly I built this locally, ran the unit tests, and ran project for a while Everything seems to be working correctly, feeds are syncing, images are analyzed, etc.

It's probably a good idea to get numerous sets of eyes on this one though, given the scope of the change.

zburstein (approved):
While I was unable to go over every single change because there are so many, everything I went over looks good. I specifically focused on changes to calls in the actual code. Thanks for doing this! There is a lot of really great stuff in here that makes it a lot easier to follow what is occurring in our code

PR #1149, opened 2021-07-21
factor in release for RPM if present

found-it:
Signed-off-by: James Petersen <jpetersenames@gmail.com>

**What this PR does / why we need it**: Fixes the RPM hints file not accounting for release when the field is present and source is not.

**Special notes**:

The old code doesn't take the `release` into consideration for RPMHints when parsing out the rpm fields. When the `release` field is present but the `source` field is missing, the logic will call `resolve_rpm_fields()` which didn't make use of the `release` field.

So a hints file with the following contents would fail with the `HintsTypeError`

```json
{
  "packages": [
    {
      "name": "tzdata",
      "version": "2021a",
      "release": "1.fc34",
      "arch": "noarch",
      "type": "rpm"
    }
  ]
}
```

And the following hints file won't throw any errors but it will incorrectly parse `container` out as the version

```json
{
  "packages": [
    {
      "name": "fedora-release-identity-container",
      "version": "35",
      "release": "0.7",
      "arch": "noarch",
      "type": "rpm"
    }
  ]
}
```

### DIFF
--- a/anchore_engine/analyzers/hints.py
+++ b/anchore_engine/analyzers/hints.py
@@ -64,12 +64,12 @@ class RPMHint(BaseHint):
     def resolve_rpm_fields(self):
         from anchore_engine.util.rpm import split_rpm_filename
 
-        # if self.release:
-        #     rpm = "{}-{}-{}.{}.rpm".format(
-        #         self.name, self.version, self.release, self.arch
-        #     )
-        # else:
-        rpm = "{}-{}.{}.rpm".format(self.name, self.version, self.arch)
+        if self.release:
+            rpm = "{}-{}-{}.{}.rpm".format(
+                self.name, self.version, self.release, self.arch
+            )
+        else:
+            rpm = "{}-{}.{}.rpm".format(self.name, self.version, self.arch)
         (
             parsed_name,
             parsed_version,
--- a/anchore_engine/analyzers/hints.py
+++ b/anchore_engine/analyzers/hints.py
@@ -64,12 +64,12 @@ class RPMHint(BaseHint):
     def resolve_rpm_fields(self):
         from anchore_engine.util.rpm import split_rpm_filename
 
-        if self.release:
-            rpm = "{}-{}-{}.{}.rpm".format(
-                self.name, self.version, self.release, self.arch
-            )
-        else:
-            rpm = "{}-{}.{}.rpm".format(self.name, self.version, self.arch)
+        # if self.release:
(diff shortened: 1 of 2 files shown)


anchore_engine/analyzers/hints.py
```diff
@@ -64,15 +64,19 @@ def __init__(self, pkg):
     def resolve_rpm_fields(self):
         from anchore_engine.util.rpm import split_rpm_filename
 
+        if self.release:
```
dspalmer99, line 67: Do we need to check for the existence of the other three class members used here? Are they always guaranteed to exist?

anchore_engine/analyzers/hints.py
```diff
@@ -64,15 +64,19 @@ def __init__(self, pkg):
...
+            rpm = "{}-{}-{}.{}.rpm".format(
```
dspalmer99, line 68: Are these the only two rpm filename formats we could hit?

dspalmer99 (commented):
Thanks for the PR! I left a few questions but it generally looks good to me.

anchore_engine/analyzers/hints.py
found-it: @dspalmer99 `self.arch` has a default of `x86_64` so I don't think that needs to be checked for. `self.name` and `self.version` are required fields per [the docs](https://docs.anchore.com/current/docs/overview/concepts/images/analysis/content_hints) and they are confirmed to exist [earlier in the code](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/analyzers/hints.py#L23).

anchore_engine/analyzers/hints.py
found-it: Those should be the only formats. See [RPM File Naming Convention](http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html). The old code assumed the version and release were in one string. Hints allow version and release to be broken out into separate fields which wasn't accounted for.

Vijay-P (approved):

anchore_engine/analyzers/hints.py
dspalmer99: Ack, thanks

anchore_engine/analyzers/hints.py
dspalmer99: Thanks, gotcha

dspalmer99 (approved):

PR #1154, opened 2021-07-28
Certifi path corrections

### LINKED ISSUE #1073: /home/anchore/certs Handling Broken due to Python version mismatch

---

**Is this a BUG REPORT or a FEATURE REQUEST?** (choose one): BUG REPORT

docker.io/anchore/anchore-engine:v0.9.4
anchore-cli, version 0.9.1

**What happened**:
Mounted certs at /home/anchore/certs/, deployed using helm chart version 1.12.15 . Instead of using certifi cacert.pem _and_ the mounted custom certs, the original certs are disregarded. This is because the docker image uses python 3.8, but the docker-entrypoint.sh script seems to expect python 3.6, so an incorrect file path is used.

The result is that any external SSL connections are untrusted and fail, such as the request needed to sync feeds.

Certs to be added to both OS and certfi cert stores.

From pod logs:
```
cp: cannot stat '/usr/local/lib/python3.6/site-packages/certifi/cacert.pem': No such file or directory
```

**What docker images are you using**:
docker.io/anchore/anchore-engine:v0.9.4 (sha256:f9f5aecc19fa813e7c98dd647af1ef1c87be799883de0da2096ab4e3d5e51bd3)

Mount a certificate to /home/anchore/certs/, start container normally.

Documentation at `https://docs.anchore.com/current/docs/installation/configuration/custom_certs/` also erroneously lists the default certificate file at the incorrect location.

A potential workaround, extracting cacert.pem from image, and re-mounting back into /home/anchore/certs/, is not possible with the helm chart, as the entire cert must be loaded into a single secret, and the combined cert is too large to fit into a Kubernetes Secret. The workaround for the workaround is to only mount in the particular root CA that is needed for the https://ancho.re/v1/service/feeds call, but this feels wrong somehow.

`python3 -c 'import site; print(site.getsitepackages())'` can be used to determine path to site packages dynamically, which could be used to future-proof the docker-entrypoint.sh script.

If this element of the docker-entrypoint.sh script is supported, it should be documented and maintained. If it is not supported, it should be removed and the helm chart updated accordingly.

I'm happy to submit a PR if it would be accepted.

Vijay-P:
Correct path of certifi's cacert.pem file in docker-entrypoint.sh. The broken path was preventing users from adding custom CAs.

Fixes #1073 

**Special notes**:
As suggested by user who reported the issue, we might dynamically determine which python version is being used in the dockerfile rather than hardcoding the path.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -10,7 +10,7 @@ if [[ -d "/home/anchore/certs" ]] && [[ ! -z "$(ls -A /home/anchore/certs)" ]]; 
     mkdir -p /home/anchore/certs_override/python
     mkdir -p /home/anchore/certs_override/os
     ### for python
-    cp /usr/local/lib/python3.6/site-packages/certifi/cacert.pem /home/anchore/certs_override/python/cacert.pem
+    cp /usr/local/lib/python3.8/site-packages/certifi/cacert.pem /home/anchore/certs_override/python/cacert.pem
     for file in /home/anchore/certs/*; do
         if grep -q 'BEGIN CERTIFICATE' "${file}"; then
             cat "${file}" >> /home/anchore/certs_override/python/cacert.pem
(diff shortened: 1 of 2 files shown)


Btodhunter (approved):
lgtm!

dakaneye (approved):
LGTM

PR #1156, opened 2021-07-29
remove commented out methods

dakaneye:
During code investigation, found these methods that were commented out and unused.


nightfurys (approved):
lgtm. Not strictly necessary for the upcoming patch release. This can go into master after the minor release branches are created

zhill (approved):
lgtm

PR #1163, opened 2021-08-11
Iterate analysis over all available tags for a given image digest

nightfurys:
This PR addresses the issue of failing analysis for an image that has multiple tag references. If one of the tags in accessible, the analysis for that specific tag fails as expected. However analysis of a different tag that is accessible may also fail. This happens because 
1. the tag reference is not maintained in the analysis request
2. analyzers simply try the 1st tag in the list of tags to pull the image

For instance image x was analyzed with tag y first followed by tag z. After deleting tag y from the registry, force analyzing tag y fails as expected. Force analyzing tag z also fails as the analyzers attempt to pull the image with tag y 

Solution: analyzers should iterate through the list of all available tags until they can successfully pull the image and analyze it

### DIFF
--- a/anchore_engine/clients/localanchore_standalone.py
+++ b/anchore_engine/clients/localanchore_standalone.py
@@ -1011,143 +1011,160 @@ def analyze_image(
     event = None
     pullstring = None
     fulltag = None
+    analysis_error = None
 
-    try:
-        imageDigest = image_record["imageDigest"]
+    imageDigest = image_record["imageDigest"]
 
-        dockerfile_mode = image_record.get("dockerfile_mode", "")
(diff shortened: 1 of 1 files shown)


dspalmer99 (approved):
LGTM, it would be great to have some unit tests on this, but I know the complexity of these methods makes that difficult until we can refactor.

nightfurys:
> LGTM, it would be great to have some unit tests on this, but I know the complexity of these methods makes that difficult until we can refactor.

Yep, I tried to split things up but that means a high potential for regression :( I want an automated end-to-end test for this specific use case

zburstein (approved):
lgtm but ditto setting up tests of some kind in e2e

PR #1164, opened 2021-08-11
Generates UserAnalyzeImageCompleted events for tags where the image (with a different tag) has already been analyzed.

### LINKED ISSUE #1142: Events not generated correctly if an image digest has multiple tags
Adding the same image with multiple tags generates the `user.image.analysis.completed` event for just one of the tags. Events for remaining tags are missing and since events drive notifications, the related notifications are also missing for the other tags. This is tangentially related to #1139 because of how tags and digests are maintained in engine

dspalmer99:
**What this PR does / why we need it**: This PR generates events for already generated images, when a new tag is submitted. The previous behavior was that when multiple tags were submitted, only the first one through the analyzer had a UserAnalyzeImageCompleted event generated. This changes that to generate events for each tag on the image.

There are two paths for this:

1. In the analyzer, when the image is actually being analyzed, the code will now check for an updated image from catalog just before generating events, and then generate an event for each tag. This accommodates the case where tags are added in close proximity to each other, while the analyzer is still running.
2. For the case where a tag is added after the image is already analyzed, the catalog will now generate an event when updating the tag on the image.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #1142 

**Special notes**:

### DIFF
--- a/anchore_engine/services/analyzer/analysis.py
+++ b/anchore_engine/services/analyzer/analysis.py
@@ -19,13 +19,15 @@ from anchore_engine.services.analyzer.errors import (
 from anchore_engine.services.analyzer.tasks import WorkerTask
 from anchore_engine.services.analyzer.utils import (
     emit_events,
-    fulltag_from_detail,
     get_tempdir,
     update_analysis_complete,
     update_analysis_failed,
     update_analysis_started,
 )
 from anchore_engine.subsys import events, logger
+from anchore_engine.subsys.events.factories import (
+    analysis_complete_notification_factory,
+)
 from anchore_engine.utils import AnchoreException
 
 ANALYSIS_TIME_SECONDS_BUCKETS = [
@@ -40,48 +42,6 @@ ANALYSIS_TIME_SECONDS_BUCKETS = [
     1800.0,
     3600.0,
 ]
-
-
-def analysis_complete_notification_factory(
-    account,
-    image_digest: str,
-    last_analysis_status: str,
-    analysis_status: str,
-    image_detail: dict,
-    annotations: dict,
-) -> events.UserAnalyzeImageCompleted:
-    """
-    Return a constructed UserAnalysImageCompleted event from the input data
-
-    :param account:
-    :param image_digest:
-    :param last_analysis_status:
-    :param analysis_status:
-    :param image_detail:
-    :param annotations:
-    :return:
-    """
-
-    payload = {
-        "last_eval": {
-            "imageDigest": image_digest,
-            "analysis_status": last_analysis_status,
-            "annotations": annotations,
-        },
-        "curr_eval": {
-            "imageDigest": image_digest,
-            "analysis_status": analysis_status,
-            "annotations": annotations,
-        },
-        "subscription_type": "analysis_update",
-        "annotations": annotations or {},
-    }
-
-    fulltag = fulltag_from_detail(image_detail)
-
-    return events.UserAnalyzeImageCompleted(
-        user_id=account, full_tag=fulltag, data=payload
-    )
 
 
 def notify_analysis_complete(
@@ -103,6 +63,17 @@ def notify_analysis_complete(
             annotations = json.loads(image_record.get("annotations", "{}"))
     except Exception as err:
         logger.warn("could not marshal annotations from json - exception: " + str(err))
+
+    # Re-pull the image record before sending notifications
+    # For the case where new tags for the image were added
+    catalog_client = internal_client_for(CatalogClient, account)
+    try:
+        image_record = get_image_record(catalog_client, image_digest)
+    except Exception as err:
+        logger.warn(
+            "Cannot re-get image from catalog for image digest %s, generating notifications for already-known tags"
+            % str(image_digest)
+        )
 
     for image_detail in image_record["image_detail"]:
         event = analysis_complete_notification_factory(
@@ -305,9 +276,7 @@ def process_analyzer_job(
         # check to make sure image is still in DB
         catalog_client = internal_client_for(CatalogClient, account)
         try:
-            image_record = catalog_client.get_image(image_digest)
-            if not image_record:
-                raise Exception("empty image record from catalog")
+            image_record = get_image_record(catalog_client, image_digest)
         except Exception as err:
             logger.warn(
                 "dequeued image cannot be fetched from catalog - skipping analysis ("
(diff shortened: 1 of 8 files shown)


anchore_engine/subsys/events/factories.py
```diff
@@ -0,0 +1,57 @@
+from anchore_engine.subsys.events import UserAnalyzeImageCompleted
```
dspalmer99, line 1: @nightfurys I pulled these out so they'd 1) be in a common location between the services and 2) be more testable. I am not at all married to the module name if you have a better suggestion.

tests/unit/anchore_engine/subsys/events/test_factories.py
```diff
@@ -0,0 +1,70 @@
+import pytest
```
dspalmer99, line 1: I would have liked to add more unit tests to this PR, but being able to do so on the analyzer and catalog code that has been changed I think requires them to be refactored into more-atomic methods, which I do not consider in scope for this change/release.

anchore_engine/subsys/events/factories.py
nightfurys: Refactoring this for testability is fine by me, can we rename the file to utils?

anchore_engine/services/analyzer/analysis.py
```diff
@@ -104,6 +64,17 @@ def notify_analysis_complete(
...
 
+    # Re-pull the image record before sending notifications
+    # For the case where new tags for the image were added
+    catalog_client = internal_client_for(CatalogClient, account)
+    try:
+        image_record = get_image_record(catalog_client, image_digest)
+    except Exception as err:
+        logger.warn(
+            "Cannot re-get image from catalog for image digest %s, generating notifications for already-known tags"
+            % str(image_digest)
```
nightfurys, line 75: Update this to use the safe log formatting here. no need to stringify image_digest

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -1964,7 +1968,9 @@ def add_or_update_image(
             dockerfile = None
             dockerfile_mode = None
 
-    # logger.debug("rationalized input for imageId ("+str(imageId)+"): " + json.dumps(image_ids, indent=4))
+    # Get a catalog client, for any events that need to be emitted
+    catalog_client = internal_client_for(CatalogClient, userId)
```
nightfurys, line 1972: This logic is already executing in the catalog service, so the client is not needed for invoking ops with in the service. Service clients are used for communicating between services via internal APIs. Using the catalog client may work but essentially the request leaves the service, gets routed over the wire and comes back to the same service (not instance) which is unnecessary. Use add_event() in this module to create a new event

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2162,6 +2168,24 @@ def add_or_update_image(
...
+                                for new_id in new_image_detail:
+                                    event = analysis_complete_notification_factory(
+                                        userId,
+                                        image_record["imageDigest"],
+                                        image_record["analysis_status"],
+                                        image_record["analysis_status"],
+                                        new_id,
+                                        annotations,
+                                    )
+                                    catalog_client.add_event(event)
```
nightfurys, line 2187: See above comment regarding client usage. use add_event() to create a new event

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2162,6 +2168,24 @@ def add_or_update_image(
...
                             if not parent_manifest:
                                 parent_manifest = json.dumps({})
+
+                            # If image status is already analyzed, generate a UserAnalyzeImageCompleted event
+                            if image_record[
+                                "analysis_status"
+                            ] == anchore_engine.subsys.taskstate.complete_state(
+                                "analyze"
+                            ):
+                                for new_id in new_image_detail:
```
nightfurys, line 2178: This may not be accurate. We should only be generating events for the tags in this request, not all tags. Otherwise the same event will be generated for a tag multiple times. The loop should probably iterate over `tags` instead

nightfurys (changes requested):

dspalmer99:
@nightfurys Thanks for the feedback, I have updated all of those items.

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2177,6 +2199,37 @@ def add_or_update_image(
...
 
 
+def is_new_tag(image_record: dict, registry: str, repo: str, tag: str):
+    """
+    Returns false if the provided registry, repo, tag tuple exactly match those fields in one of the image_detail objects in
+    image_record. If no such match is possible, return true.
+    """
+
+    # If we are missing an image_details block entirely or it is None, assume this is a new tag
+    if "image_detail" not in image_record or not image_record["image_detail"]:
```
nightfurys, line 2209: (nit) slightly optimized check for image_record and image_detail in a single condition
```
if image_record and image_record.get("image_detail")
```

anchore_engine/services/catalog/catalog_impl.py
```diff
@@ -2177,6 +2199,37 @@ def add_or_update_image(
...
+    else:
+        # Iterate through each existing image_detail
+        for image_detail in image_record["image_detail"]:
+            # We shouldn't be missing one of these fields, but if we are, assume this is not a match
+            if (
+                "registry" not in image_detail
+                or "repo" not in image_detail
+                or "tag" not in image_detail
+            ):
+                return True
```
nightfurys, line 2220: I think it should continue iterating through the list here rather than returning True if registry repo or tag are not in the image_detail. To make this simpler I would recommend the loop look like this 
```
for image_detail in image_record.get("image_detail", []):
  existing_registry = image_detail.get("registry")
  existing_repository = image_detail.get("repo")
  existing_tag = image_detail.get("tag")

   if existing_registry == registry and existing_repository == repository and existing_tag == tag:
     return False  
else:
  return  True
```

nightfurys (commented):
minor feedback in "is-tag-new" function

anchore_engine/services/catalog/catalog_impl.py
dspalmer99: I've made that change

anchore_engine/services/catalog/catalog_impl.py
dspalmer99: I made that change as well. This is better

dspalmer99: As now if one of those parameters is missing in both tags, it will match correctly (albeit that would be very unexpected)

nightfurys (approved):
lgtm, thanks for incorporating the feedback @dspalmer99

PR #1165, opened 2021-08-11
Fix hints key error

zburstein:
Fixes a bug where key error exceptions were raised when syft did not find an artifact type that was later applied through hints. Hints application was not declaring nested attributes in the same way that syft would. This change does thi by:
1. Creates a common utility for all handlers for saving an entry to the report that will define the necessary attributes in the findings if not already present
2. This same handler adds a condition to check whether the pkg is already present in the report. If so it logs a message and does not overwrite the existing entry
3. Adds unit tests and updates existing functional tests


anchore_engine/analyzers/syft/handlers/common.py
```diff
@@ -0,0 +1,45 @@
+from collections import defaultdict
+from typing import Union
+
+from anchore_engine.analyzers.utils import merge_nested_dict
+from anchore_engine.subsys import logger
+
+
+def save_entry_to_findings(
+    findings: Union[dict, defaultdict], entry: dict, pkg_type: str, pkg_key: str
```
Vijay-P, line 9: https://docs.python.org/3/library/typing.html#typing.DefaultDict

docs/content/docs/general/concepts/images/analysis/content_hints.md
```diff
@@ -5,11 +5,11 @@ weight: 5
...
-discovered by Anchore during its image analysis process.  The hints file, if present, contains records that describe a software package characteristics explicitly, 
+Anchore Engine includes the ability to read a user-supplied 'hints' file to allow users to add software artifacts to Anchore's analysis report.  The hints file, if present, contains records that describe a software package characteristics explicitly, 
 and are then added to the software bill of materials (SBOM).  For example, if the owner of a CI/CD container build process knows that there are some 
-software packages installed explicitly in a container image, but Anchore's regular analyzers are either not discovering the software or the analysis of the software 
-package is incomplete, this mechanism can be used to include that information in the image's SBOM, exactly as if the package were discovered normally.
+software packages installed explicitly in a container image, but Anchore's regular analyzers fail to them, this mechanism can be used to include that information in the image's SBOM, exactly as if the packages were discovered normally.
```
Vijay-P, line 10: "fail to them" seems like a typo. I think you need a verb after "fail to" like "find".

Vijay-P: also nit: line lengths seem off

tests/functional/clients/standalone/test_hints.py
```diff
@@ -18,52 +18,87 @@ def test_npm_hints(self, hints_image):
...
+                },
             ]
         }
         result = hints_image(hints, "npm")
         pkgs = result["image"]["imagedata"]["analysis_report"]["package_list"][
             "pkgs.npms"
         ]["base"]
+
+        # Verify message logged that package cannot be overwritten
+        assert "package already present under" in caplog.text
```
Vijay-P, line 36: Would be nice if you could add a flag to the parametrization that tells the test whether or not a given package should generate a log message. It seems that these tests might be annoying to update for someone who isn't familiar with the hints.

Vijay-P (changes requested):
Overall looks good. Nice work on the tests. Some minor changes requested, but most are NIT.

anchore_engine/analyzers/syft/handlers/common.py
```diff
@@ -0,0 +1,44 @@
+from typing import DefaultDict, Union
+
+from anchore_engine.analyzers.utils import merge_nested_dict
+from anchore_engine.subsys import logger
+
+
+def save_entry_to_findings(
```
Toure, line 7: I like this consolidation.

Toure (approved):

tests/functional/clients/standalone/test_hints.py
```diff
@@ -1,8 +1,12 @@
 import json
 
 
+def overwrite_msg_in_log(caplog, pkg, pkg_type):
```
Vijay-P, line 4: nit: user verb like "is" -> "is overwrite msg in log"

tests/unit/anchore_engine/analyzers/hints/test_apply_hints.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Reads and returns the analyzer report in this directory. Has every type of package in it
+        """
+        f = open(os.path.join(module_path, "analyzer_report.json"))
+        return json.load(f)
+
+    @pytest.fixture
+    def hints_dirpath(self, module_path):
+        """
+        Direcotry path of hints files
```
Vijay-P, line 45: nit: typo

Vijay-P (approved):
some comments but not blocking

anchore_engine/analyzers/syft/handlers/common.py
```diff
@@ -0,0 +1,44 @@
+from typing import DefaultDict, Union
```
dspalmer99, line 1: Nit: Do we have a consensus on whether these types of modules should be called `common`, `util`, or `utils`? We have all three in the codebase but I think they are all more or less the same thing.

(This is not at all a review-blocker comment, just interested in folks thoughts.)

dspalmer99 (approved):
One more-general question, but its not a blocker and overall this LGTM

anchore_engine/analyzers/syft/handlers/common.py
zburstein: Yeah we definitely do not have a consensus on this. Maybe something we should bring up in the platform team meeting?

PR #1174, opened 2021-08-17
Update docker entrypoint with certifi path and set

found-it:

# Proposed enhancement

Use the `python3 -m certifi` command which returns the path of the certifi cacert.pem file. This means the script doesn't need to be changed whenever the python version for the container is changed.

Also updated line 13 with [SC2236](https://github.com/koalaman/shellcheck/wiki/SC2236) change to remove double negative.

Use `set` inside the entrypoint so it hard fails whenever there are errors.

**Special notes**:

##### To test

```sh
mkdir -p /tmp/certs
openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/certs/anchore.key -out /tmp/certs/anchore.crt
# fill out all info with junk, fields don't matter for test

make build
docker run -v /tmp/certs/:/home/anchore/certs/ anchore-engine:dev cat /home/anchore/certs_override/python/cacert.pem
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 if [[ "${SET_HOSTID_TO_HOSTNAME}" == "true" ]]; then
     echo "Setting ANCHORE_HOST_ID to ${HOSTNAME}"
     export ANCHORE_HOST_ID=${HOSTNAME}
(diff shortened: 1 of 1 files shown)


Vijay-P:
@found-it pretty sure this partially conflicts with what's currently in master. I think we already have the certifi fix in. Can you rebase please?

rbrady (approved):
The options set here are a good common addition. Thanks @found-it .

zhill:
Can this be squash-merged to only have 1 commit instead of 7 since the changeset is fairly small?

PR #1169, opened 2021-08-13
Update Syft Version

rbrady:
An issue was fixed in Syft to address cleaning up the stereoscope
cache on exit and reduce the usage growth inside the analyzer
container.

This patch bumps the Syft version up in our Dockerfile to v0.19.1
to include the version that fixes the issue.

Signed-off-by: Ryan Brady <ryan.brady@anchore.com>

**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>)(, fixes #<issue_number, ...)` format, will close the issue when PR is merged*: fixes #:

**Special notes**:

PR #1175, opened 2021-08-17
Automate upgrading Anchore tools

DaneWeber:
I would like to see us move toward always having the latest versions of Syft and Grype in Engine. Manually updating the Dockerfile is a minimal step (see https://github.com/anchore/anchore-engine/pull/1169), but I'd like to automate this so that we can get early warning when a Syft or Grype upgrade breaks our automated tests.

This PR adds a Makefile script for updating the Dockerfile with the following outline:
1. `curl` latest release data of Syft & Grype from GitHub
2. Get version via `jq` (install if necessary in CI)
3. Modify `Dockerfile` via `sed`

The full script can be executed with `make latest-anchore-tools`.

I tested from both the Mac command line and within a `circleci/python:3.9` Docker container to ensure that it works with both GNU and BSD versions of `sed`.

A follow-up PR is intended that defines a GitHub Action to auto-open a PR when this script finds a newer version of Syft or Grype. The CircleCI checks will show if tests fail based on the upgrade, so this will have value even if the PRs are not merged rapidly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,15 +251,15 @@ endif
 
 SYFT_LATEST_VERSION = $(shell curl "https://api.github.com/repos/anchore/syft/releases/latest" 2>/dev/null | jq -r '.tag_name')
 # Regex note: double-dollarsign for Makefile escaping; % instead of / for Sed to match anchore/syft
-latest-syft: jq-installed
+upgrade-syft: jq-installed ## Upgrade Syft to the latest release
 	# Setting Syft to ${SYFT_LATEST_VERSION}
 	$(SEDI) 's%^(.+anchore/syft.+)v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$$%\1${SYFT_LATEST_VERSION}%' Dockerfile
 
 GRYPE_LATEST_VERSION = $(shell curl "https://api.github.com/repos/anchore/grype/releases/latest" 2>/dev/null | jq -r '.tag_name')
 # Regex note: double-dollarsign for Makefile escaping; % instead of / for Sed to match anchore/syft
-latest-grype: jq-installed
+upgrade-grype: jq-installed ## Upgrade Grype to the latest release
 	# Setting Grype to ${GRYPE_LATEST_VERSION}
 	$(SEDI) 's%^(.+anchore/grype.+)v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$$%\1${GRYPE_LATEST_VERSION}%' Dockerfile
 
 # TODO: Intent is to create a weekly/daily/continuous GitHub Action that runs the following and auto-opens a PR
-latest-anchore-tools: latest-syft latest-grype
+upgrade-anchore-tools: upgrade-syft upgrade-grype ## Upgrade Syft and Grype to the latest release
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ GIT_TAG := $(shell echo $${CIRCLE_TAG:=null})
 .PHONY: push-dev push-nightly push-rc push-prod push-rebuild push-redhat
 .PHONY: compose-up compose-down cluster-up cluster-down
 .PHONY: setup-test-infra venv printvars help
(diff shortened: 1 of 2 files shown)


DaneWeber:
There is a patch release in progress, so I'll hold off on requesting a review until that's complete.

DaneWeber:
I have the GitHub action mostly drafted, but I'd like to merge this in first before tackling that.

zburstein (approved):
Dont have too much experience with sed but  just tested and it appears to work.

Makefile
```diff
@@ -221,3 +222,44 @@ help:
...
+SEDI = sed -E -i''
+endif
+
+
+# Code change targets
+#######################
+
+SYFT_LATEST_VERSION = $(shell curl "https://api.github.com/repos/anchore/syft/releases/latest" 2>/dev/null | jq -r '.tag_name')
+# Regex note: double-dollarsign for Makefile escaping; % instead of / for Sed to match anchore/syft
+latest-syft: jq-installed
```
rbrady, line 254: Would it be better to describe the task with a verb and noun, to match more closely what's going on?

update-grype
update-syft
update-anchore-tools

Makefile
```diff
@@ -221,3 +222,44 @@ help:
...
+
+# Utility targets
+#######################
+
+jq-installed:
+ifeq ($(OS),Darwin)
+	# Skipping installation of jq for local dev on Mac.
+	# You can install via 'brew install jq' the following command if needed.
+else
+	which jq ; if [ $$? -eq 1 ] ; then sudo apt-get install -y jq ; fi
```
Btodhunter, line 234: Nit: I think this is a little easier to read.

```
if which jq; then sudo apt-get install -y jq ; fi
```

Btodhunter (approved):
Nit: I'd like to see syft/grype versions declared in variables at the top of the Dockerfile like we do for skopeo & kai. That would simplify your sed regex a little, but would also make it much easier to see what version we're installing.

Other than that, LGTM!

rbrady (approved):
I tested and it works.  It would be nice to figure out how to add the targets to the printout when you run bare make for discovery purposes, but maybe that's a good follow up when we have spare time.

Makefile
Btodhunter: > I tested and it works. It would be nice to figure out how to add the targets to the printout when you run bare make for discovery purposes, but maybe that's a good follow up when we have spare time.

Good catch @rbrady! To add this to the help output, you just add two `##` after the target declaration. Like this:
```
latest-syf: jq-installed ## Update syft to the latest version
```

Makefile
DaneWeber: I'm AFK, but wouldn't your rewrite do the opposite? I'm looking to install jq if it isn't found (error code 1).

Makefile
DaneWeber: Nice. I'll make these improvements.

DaneWeber:
> Nit: I'd like to see syft/grype versions declared in variables at the top of the Dockerfile like we do for skopeo & kai. That would simplify your sed regex a little, but would also make it much easier to see what version we're installing.
> 
> Other than that, LGTM!

This is a good call. I plowed ahead with automating our existing process without stepping back to consider a refactor like this. Even if this doesn't simplify the regex, I imagine it will make this script more resilient to future changes to the Dockerfile.

Makefile
DaneWeber: Would you prefer

```
if which jq ; then ; else sudo apt-get install -y jq ; fi
```

rbrady (approved):
I like the change to "upgrade" from "latest".  It's now more descriptive with its action.  Thanks!

Makefile
Btodhunter: whoops, you're right! This is what I _meant_ to say :)
```
if ! which jq; then sudo apt-get install -y jq ; fi
```
However, it's not a big deal. This just feels cleaner to _me_

DaneWeber: Done!

Btodhunter (approved):

PR #1183, opened 2021-08-24
Quote all of the $VARs in the Dockerfile

DaneWeber:
BASH variables should be enclosed in double-quotes. Ref: https://tldp.org/LDP/abs/html/quotingvar.html

Because I made changes to the Dockerfile and @Btodhunter pointed out this good practice I thought I'd open a PR to apply this to the rest of the Dockerfile.

`make build` succeeds.


Btodhunter (approved):
lgtm!

PR #1196, opened 2021-08-30
Set default provider to grype

zburstein:
This PR does 2 things:

1. Changes the default vulnerability provider to be grype
2. Updates the testing infrastructure to support this change


scripts/ci/docker-compose-ci.yaml
```diff
@@ -122,7 +122,7 @@ services:
       - ANCHORE_FEEDS_ENABLED=true
       - ANCHORE_FEEDS_CLIENT_URL=null
       - ANCHORE_FEEDS_TOKEN_URL=null
-      - ANCHORE_VULNERABILITIES_PROVIDER=${ANCHORE_VULNERABILITIES_PROVIDER:-legacy}
```
Vijay-P, line 125: can you remind me what the dash does here in front of grype?

zburstein: Its for use in functional tests in ci https://github.com/anchore/anchore-engine/blob/master/.circleci/config.yml#L226. It used to be that testing legacy with the makefile command was just `setup-and-test-functional` while grype was `setup-and-test-functional`. Now that I renamed them though we can feasibly move the dash to ci command and out fo the docker compose

zburstein: Scratch that last comment. The dash that I was referring to was the one in the circle ci config. The das here in the docker compose is the syntax for conditionally setting a variable within a docker compose file if it is not already set

zhill (approved):
lgtm

PR #1197, opened 2021-09-01
Set will_not_fix in Vulnerability swagger def to boolean type

zburstein:
`will_not_fix` is set to type string when it should be boolean so this PR fixes that


dspalmer99 (approved):

Toure (approved):
LGTM

PR #1198, opened 2021-09-02
convert FeedRegistry from global to singleton pattern

Vijay-P:

**Special notes**:


zhill:
I like this direction. I propose a generalization into a 'RegistrySingleton' that can be used for other kinds of registries that are singletons since I think we'll use this pattern more than once.

PR #1210, opened 2021-09-10
Add flag to get method of ImageContentGetter to allow content retriev…

Vijay-P:
This PR is to support some functionality needed in Anchore Enterprise.

The Bug discovered in Enterprise:
On image ingress in the policy engine (after the image is loaded from manifest, while loading the vulnerabilities for the image), the catalog image content API is called to retrieve the image content with CPE corrections. The catalog ImageCorrectedContentGetter expects the task state of the image to be "analyzed", but because the image ingress is incomplete, the state at this point is still "analyzing". This bug was causing image analysis to fail in enterprise for the GrypeCorrectingScanner in the policy engine.

The fix in engine:
Update the ImageContentGetter's .get method to accept a flag, which will allow content to be retrieved while images are still in "analyzing" state.

The fix in enterprise:
In enterprise: Added a query parameter flag to the catalog's internal API for image content retrieval. This flag is called allow_analyzing_state and defaults to false. When the flag is set to true, it allows for image content retrieval when the image is in analyzing state. Under the hood, this uses the new flag in the ImageContentGetter's .get method.

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/image_content/get_image_content.py
+++ b/anchore_engine/services/catalog/image_content/get_image_content.py
@@ -27,17 +27,27 @@ class ImageContentGetter:
             "image_digest": self.image_digest,
         }
 
-    def verify_analysis_status(self, image_report):
-        if image_report and image_report["analysis_status"] != taskstate.complete_state(
-            "analyze"
-        ):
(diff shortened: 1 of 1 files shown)


zhill (approved):
lgtm, this gives some nice flexibility later.

PR #1211, opened 2021-09-10
Update Grype to v0.17.0

Vijay-P:
Grype matching logic has changed, so the expected content for one of the vuln tests has changed.


zhill (approved):
lgtm

PR #1212, opened 2021-09-10
Uncomment yum install line

robertp:
Instruct yum to retain cache across yum invocations; note that this is
NOT required today but if more yum invocations are added to stage 1 it
will be required, so it makes sense to add it preemptively

Signed-off-by: Robert Prince <robert.prince@anchore.com>


Btodhunter (approved):
lgtm

zhill (approved):
lgtm

PR #1214, opened 2021-09-13
Syft v0.23.0 update

Vijay-P:
Part of VulnDB support in grype includes a change in CPE generation from Syft. So i'm trying to update Syft. Updating the test output should pave the way for the next version bump as well.


zhill:
Updated the branch to get the CI fixes in.

zhill (approved):
nice update. lgtm

zhill:
For the test failures, the code here: https://github.com/anchore/anchore-engine/commit/be20f3dd627bd91312075eb202c72877f78eb2dc may help with results sorting if that's the issue.

Vijay-P:
> For the test failures, the code here: [be20f3d](https://github.com/anchore/anchore-engine/commit/be20f3dd627bd91312075eb202c72877f78eb2dc) may help with results sorting if that's the issue.

Turns out my test runner container was caching analysis manifests in a different directory from my local deployment and was updating to the wrong expected content as a result.

PR #1218, opened 2021-09-15
Feed index rename

Vijay-P:
This PR adds new common db utility to truncate index names in a reproducible way that matches sqlalchemy's default naming convention but is FIPS compliant. All uses of sqlalchemy.Index were updated to use new truncate_index_name function.

Sources used to verify approach:
https://docs.sqlalchemy.org/en/14/core/constraints.html#the-default-naming-convention
https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.MetaData.params.naming_convention

Query used to check index names:

```
SELECT
    tablename,
    indexname
FROM
    pg_indexes
WHERE indexname LIKE 'ix_%'
ORDER BY
    tablename,
    indexname;
```

### DIFF
--- a/anchore_engine/db/entities/catalog.py
+++ b/anchore_engine/db/entities/catalog.py
@@ -21,16 +21,16 @@ from sqlalchemy import (
 )
 from sqlalchemy.orm import relationship
 
-from anchore_engine.util.time import datetime_to_rfc3339
-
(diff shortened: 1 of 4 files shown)


zhill (approved):
lgtm. nice update.

PR #1220, opened 2021-09-16
Fix debian source package name

nightfurys:
Save debian source package name without version to analysis manifest and image content documents

Downstream needs the source package name and tries to split using the package version but it's error prone since source and package versions may be different. An improper source package may result in missed vulnerability matches

Example of not-working case: following is a snippet from syft output of bsdutils package
```
{
   "name": "bsdutils",
   "version": "1:2.20.1-5.3",
   "type": "deb",
   "foundBy": "dpkgdb-cataloger",
   "metadataType": "DpkgMetadata",
   "metadata": {
    "package": "bsdutils",
    "source": "util-linux",
    "version": "1:2.20.1-5.3",
    "sourceVersion": "2.20.1-5.3",
```

Notice version and sourceVersion are different because of the epoch. Before the change the analyzer populated  `util-linux-2.20.1-5.3` as the source package. policy-engine tries to split this string using the package version in the loaders. But that does not work because of the epoch difference in the versions. This resulted in the generated grype sbom with source package set to mangled package name version combo, util-linux-2.20.1-5.3. That's an issue because grype tries to match util-linux-2.20.1-5.3 instead of just util-linux


Vijay-P (approved):
LGTM

PR #1222, opened 2021-09-16
CI change

Vijay-P:
CI change: update .pylintrc file to use disable=all and use rc file for enabling warnings only

Pylint got an update and a new warning got added.

This change will prevent new warnings from getting included in pylint's checks during CI runs.

Signed-off-by: Vijay Pillai vijay.pillai@anchore.com

PR #1225, opened 2021-09-17
Enable all passing pylint rules

DaneWeber:
There are many Pylint rules that currently pass. By enabling all passing rules, we can guard against regressing in these areas.

I'm including a TODO comment with the Pylint rules that must be disabled at present. https://github.com/anchore/anchore-engine/pull/1222 removed these in order to avoid surprise failures due to new Pylint rules. As an improvement, as suggested by @Vijay-P, let's pin Pylint and create a GitHub Action to upgrade Pylint and check for new findings, that way we can intentionally decide what to do about new rules.


Vijay-P:
This PR disables unused-import. After trying different variations of the pylintrc, we found that enabling some rules causes that rule not to pass. Since we had a big import ordering and sorting PR in a few months ago that cleaned up that issue, I don't want to merge something that will stop catching violations of that issue. So i'm thinking after this week, i'll get a PR in to fix the failures for that specific warning. Then when that's merged, we can reenable the rule and merge this PR.

.pylintrc
```diff
@@ -76,22 +76,401 @@ confidence=
...
+       logging-unsupported-format,
+       long-builtin,
+       long-suffix,
+       lost-exception,
+       metaclass-assignment,
+       method-check-failed,
+       method-hidden,
+       misplaced-format-function,
        misplaced-future,
-       unused-import
```
DaneWeber, line 93: See @Vijay-P 's comment

Vijay-P (approved):
This is ready.

PR #1227, opened 2021-09-17
Add nvd to the feed group rank map in policy engine.

dspalmer99:
Signed-off-by: Dan Palmer <dan.palmer@anchore.com>

**What this PR does / why we need it**: Adds nvd to the feed rank map in policy engine. This is needed because, for the legacy provider we put nvd data into the `nvdv2` feed group, but with the v2 provider grype db call that group `nvd`.

This also adds some relevant test cases to the existing unit tests.

**Special notes**:

### DIFF
--- a/tests/unit/anchore_engine/services/policy_engine/engine/vulns/test_deduper.py
+++ b/tests/unit/anchore_engine/services/policy_engine/engine/vulns/test_deduper.py
@@ -25,6 +25,7 @@ class TestFeedGroupRank:
         "test_group, expected_rank",
         [
             pytest.param("nvdv2:cves", 1, id="nvdv2"),
+            pytest.param("nvd:cves", 2, id="nvd"),
             pytest.param("github:java", 10, id="github"),
             pytest.param("alpine:3.9", 100, id="os-distro"),
             pytest.param("foobar", 100, id="random"),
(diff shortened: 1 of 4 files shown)


tests/unit/anchore_engine/services/policy_engine/engine/vulns/test_deduper.py
```diff
@@ -550,6 +551,28 @@ class TestImageVulnerabilitiesDeduplicator:
...
+                            feed="vulnerabilities",
+                            feed_group="nvd:cves",
+                            vulnerability_id="CVE-2019-12904",
+                        ),
+                        nvd=[NVDReference(vulnerability_id="CVE-2019-12904")],
+                    ),
+                ],
+                1,
+                id="nvdv2-vs-nvd",
+            ),
```
nightfurys, line 575: Neat! Can you add a similar tests for nvd vs ghsa and ghsa vs random identifier to mimic 3rd party sources

nightfurys (commented):
lgtm! just a suggestion for another couple of unit tests

tests/unit/anchore_engine/services/policy_engine/engine/vulns/test_deduper.py
dspalmer99: Yes, wil do

dspalmer99: I've added those.

nightfurys (approved):

dspalmer99:
Looks like a few of the functional tests for grype are failing now, because of this change. I'll look into updating them next.

PR #1229, opened 2021-09-20
Grype provider sbom generation update.

nightfurys:
- Switch the content source for sbom generation in grype provider from policy engine persistence to catalog image content API. The catalog API is also used to serve user image content requests. This change ensures that all content related requests operate over the same source data and it improves result coherenence and transperency
- Add unit tests for image-content to sbom mappers

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -4,6 +4,7 @@ import uuid
 from collections import defaultdict
 from typing import Dict, List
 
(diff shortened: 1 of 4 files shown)


Vijay-P:
Looks good, but i think that you should remove deprecated methods because having them lying around is more confusing than not. We can always grab them out of git history if necessary.

nightfurys:
> Looks good, but i think that you should remove deprecated methods because having them lying around is more confusing than not. We can always grab them out of git history if necessary.

Yep, ideally I would like it all in a single commit so it can be reverted back

PR #1230, opened 2021-09-21
Fix return in kb mapper function

nightfurys:
https://github.com/anchore/anchore-engine/pull/1229 updated the mapper functionality to account for sbom generation using image content API response. It introduced a bug - a missing return statement in kb mapper which resulted in a very empty sbom. This PR adds the return statement and adds a unit test that was missed in the previous PR


dspalmer99 (approved):
LGTM

Vijay-P (approved):
Not sure why the "cpes" key was added/removed but LGTM.

PR #1233, opened 2021-09-21
Fix fixed_artifact issues

zburstein:
PR makes 2 changes:

1. Adds safe navigator to fixed_artifact reference in providers. Resolves error where fixed_artifact is none
2. Ensures that fix_has_no_advisory returns a bool. If fixed_artifact was none in that function, it would return none instead of False


nightfurys (approved):
lgtm

zhill (approved):
lgtm

PR #1236, opened 2021-09-22
Updates grype-to-engine image vulnerabilities mapper

nightfurys:
- Build url for vulnerability if grype doesn't return any and point it to the query vulnerabilities API.
- Transform third party vulnerability identifier into CVE identifier when possible


dspalmer99 (approved):
LGTM

PR #1239, opened 2021-09-22
Set default provider to grype when incorrect config

zburstein:
Sets the fdefault provider to grype when config is neither legacy or grype. Also changes the integration tests to continue to default to legacy because they are not designed for functionality with grype.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -46,3 +46,14 @@ def bundle():
             return json.load(f)
 
     return find
+
+
+@pytest.fixture(autouse=True)
+def set_legacy_provider(monkeysession):
+    def _provider_name(section):
+        return "legacy"
+
+    monkeysession.setattr(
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -46,3 +46,14 @@ def find(bundle_name="bundle-large_whitelist.json"):
             return json.load(f)
 
     return find
+
+
+@pytest.fixture(autouse=True)
+def set_legacy_provider(monkeysession):
+    def _provider_name(section):
+        return "legacy"
```
dspalmer99, line 54: @zburstein Can you open a ticket to investigate changing this to grype? Whether we switch over to doing that completely, or do parallel runs with both providers for now, either way we need to be running these tests with the v2 provider eventually (and probably sooner rather than later).

dspalmer99 (approved):
LGTM

PR #1241, opened 2021-09-22
Updated docs 1.0.0 Engine and Engine Release Notes

anchoredeanlappi:
Dean made Anchore Engine documentation changes, including the 1.0.0 release notes.

Signed-off-by: anchoredeanlappi <88210304+anchoredeanlappi@users.noreply.github.com>

**Special notes**:

### DIFF
--- a/docs/content/docs/grype/_index.md
+++ b/docs/content/docs/grype/_index.md
@@ -4,7 +4,7 @@ linkTitle: "Grype Vulnerability Scanner"
 weight: 1
 ---
 
-As of Anchore Engine 1.0.0, Anchore Engine is fully integrated with Grype by default for vulnerability scanning. The V2 vulnerability scanner, based on Grype, replaces the legacy vulnerability scanner in previous versions of Anchore Engine.
+As of Anchore Engine 1.0.0, Anchore Engine is fully integrated with Grype by default for vulnerability scanning. The V2 vulnerability scanner, based on Grype, replaces the legacy vulnerability scanner in previous versions of Anchore Engine. The legacy vulnerability scanner will still be used for anyone running older versions of Anchore engine.
 You can keep the legacy vulnerability scanner when installing Anchore Engine 1.0.0, but you will have to explicitly configure 1.0.0 to use the legacy vulnerability scanner.
 
 If you are upgrading to Anchore Engine 1.0.0 from an earlier version, you will retain your previous vulnerability scanner setting. You will need to follow the linked instructions for upgrading to the new vulnerability scanner. 
@@ -14,7 +14,7 @@ If you are upgrading to Anchore Engine 1.0.0 from an earlier version, you will r
 ### Installing
 As of Anchore Engine 1.0.0, the V2 vulnerability scanner, based on Grype, is included with Anchore Engine by default. 
 
-### Installing and keeping the legacy feed
+### Installing and keeping the legacy feed and vulnerability scanner
 Anchore Engine can be configured to use the legacy vulnerability scanner. It is not possible to run both legacy and V2 vulnerability scanners at the same time. This configuration is picked up at bootstrap, and cannot be changed on a running system.
 Downgrading from the V2 vulnerability scanner back to the legacy scanner is not supported and will cause data issues.
 
@@ -50,7 +50,7 @@ Downgrading from the V2 vulnerability scanner back to the legacy scanner is not 
     helm upgrade
 ```
 
-After making the relevant change above and redeploying, the system will start up with the legacy vulnerability scanner enabled and will sync the latest version of legacy database. Note that Grype feeds will no longer be synced while legacy is configured. All vulnerability data and scanning will now come from the legacy feed.
+After making the relevant change above and redeploying, the system will start up with the legacy vulnerability scanner enabled and will sync the latest version of legacy database. Note that Grype feeds will not sync while legacy is configured. All vulnerability data and scanning will now come from the legacy scanner and feed.
 
 ### Vulnerability Feed Data and Syncs
 
--- a/docs/content/docs/grype/_index.md
+++ b/docs/content/docs/grype/_index.md
@@ -1,27 +1,25 @@
 ---
-title: "Beta - Grype Integration"
+title: "Grype Integration"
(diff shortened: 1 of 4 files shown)


docs/content/docs/grype/_index.md
```diff
@@ -1,49 +1,47 @@
 ---
-title: "Beta - Grype Integration"
+title: "Grype Integration"
 linkTitle: "Grype Vulnerability Scanner"
 weight: 1
 ---
 
-Anchore Engine 0.10 includes a limited-functionality beta release of an integration with [Grype](https://github.com/anchore/grype)
```
dspalmer99, line 7: Nit: "replaces the legacy vulnerability scanner in previous versions" -> "replaces the legacy vulnerability scanner used in previous versions". The legacy scanner will still be used for anyone running an older version of engine.

docs/content/docs/grype/_index.md
```diff
@@ -1,49 +1,47 @@
...
-for vulnerability scanning. This integration will replace the legacy vulnerability scanner in a future version of Anchore Engine,
-but is provided in this release in a preview capacity so users can try it out.
+As of Anchore Engine 1.0.0, Anchore Engine is fully integrated with Grype by default for vulnerability scanning. The V2 vulnerability scanner, based on Grype, replaces the legacy vulnerability scanner in previous versions of Anchore Engine.
+You can keep the legacy vulnerability scanner when installing Anchore Engine 1.0.0, but you will have to explicitly configure 1.0.0 to use the legacy vulnerability scanner.
```
dspalmer99, line 8: "you will have to explicitly configure 1.0.0 to use the legacy vulnerability scanner" is this correct? I thought the plan (this may have changed over time) was that existing users could upgrade to 1.0.0 and keep the legacy scanner without having to change their config. I.e. upgrading and changing to the grype provider is what requires updating your helm chart.

docs/content/docs/grype/_index.md
```diff
@@ -52,17 +50,13 @@ grype scanner back to the legacy scanner is not supported and will cause data is
     helm upgrade
 ```
 
-After making the relevant change above and redeploying, the system will start up with the grype vulnerability scanner enabled and will
-sync the latest version of grype db. Note that legacy feeds will no longer be synced while grype is configured. All vulnerability data
-and scanning will now come from the grype feed.
+After making the relevant change above and redeploying, the system will start up with the legacy vulnerability scanner enabled and will sync the latest version of legacy database. Note that Grype feeds will no longer be synced while legacy is configured. All vulnerability data and scanning will now come from the legacy feed.
```
dspalmer99, line 53: "will no longer be synced" It would be more accurate to say that they will not be synced, as a customer using (and continuing to use) the legacy scanner will not have previously synced grype feed data.

dspalmer99: "legacy feed" -> "legacy scanner and feed". The configuration for these two things is controlled by the same bit here, but the functionalities in the system are distinct.

docs/content/docs/grype/_index.md
```diff
@@ -1,49 +1,47 @@
...
+***Note:*** The legacy vulnerability scanner will be removed in a future release.
 
 ### Installing
-As of 0.10.0, Anchore Engine can be configured to use either the legacy or grype vulnerability scanner. It is not possible to run
-both vulnerability scanners at the same time. This configuration is picked up at bootstrap, and cannot be changed on a running system.
+As of Anchore Engine 1.0.0, the V2 vulnerability scanner, based on Grype, is included with Anchore Engine by default. 
 
-The grype scanner is intended for use in sandbox or staging environments in the current release. Downgrading from the
-grype scanner back to the legacy scanner is not supported and will cause data issues.
+### Installing and keeping the legacy feed
```
dspalmer99, line 17: "the legacy feed" -> "the legacy feed and vulnerability scanner"

docs/content/docs/install/configuration/feed_configuration.md
```diff
@@ -147,13 +137,11 @@ feeds:
...
-      vulnerabilities: true
-      nvdv2: true
-      github: true
-      packages: false
+      grypedb: true
+      packages: true
 ```
 
-Those boolean values can be used to enable/disable the feeds. Note that changes will require a restart of the policy engine to take effect and settng
+Those boolean values can be used to enable/disable the feeds. Note that changes will require a restart of the policy engine to take effect and settnig
```
dspalmer99, line 144: *setting

dspalmer99 (changes requested):

docs/content/docs/grype/_index.md
dspalmer99: See this thread in slack:  https://anchore.slack.com/archives/C02D6U7KS4W/p1632356329335500 for more context.

docs/content/docs/grype/_index.md
anchoredeanlappi: Hi Dan. I made the changes. Thank you. Dean

docs/content/docs/grype/_index.md
anchoredeanlappi: Hi @dspalmer99. I think this is confusing. From my understanding from @DaneWeber, it differs depending between a new install and an existing customer upgrading.
A new install comes with Grype set as the default. If a customer wants to use the legacy feed/scanner, then they will have to configure the install to use legacy.
For an existing custoer who is upgrading, they will retain their previous vulnerability scanner settings (legacy). They can also follow linked instructions to upgrade to the new Grype vulnerability scanner. 
At least this is my understanding.

docs/content/docs/grype/_index.md
anchoredeanlappi: Good catch!

docs/content/docs/grype/_index.md
anchoredeanlappi: I made this change, Thank you.

docs/content/docs/install/configuration/feed_configuration.md
anchoredeanlappi: Good catch!

PR #1245, opened 2021-09-23
fix affected version in query vulnerabilities for grype provider

Vijay-P:
[Grype provider] 
Query vulnerabilities was using the fixed in versions instead of the affected versions. This change corrects that behavior.

### DIFF
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -50,6 +50,12 @@ class GrypeVulnerability(Base, UtilMixin):
     def deserialized_fixed_in_versions(self):
         return json.loads(self.fixed_in_versions)
 
+    @property
+    def deserialized_version_constraint(self):
+        return [
+            version.strip(" '\"") for version in self.version_constraint.split("||")
+        ]
+
 
 class GrypeVulnerabilityMetadata(Base, UtilMixin):
     __tablename__ = VULNERABILITY_METADATA_TABLE_NAME
--- a/anchore_engine/clients/grype_wrapper.py
+++ b/anchore_engine/clients/grype_wrapper.py
@@ -50,12 +50,6 @@ class GrypeVulnerability(Base, UtilMixin):
     def deserialized_fixed_in_versions(self):
         return json.loads(self.fixed_in_versions)
 
-    @property
-    def deserialized_version_constraint(self):
-        return [
-            version.strip(" '\"") for version in self.version_constraint.split("||")
-        ]
-
 
 class GrypeVulnerabilityMetadata(Base, UtilMixin):
     __tablename__ = VULNERABILITY_METADATA_TABLE_NAME
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -19,6 +19,7 @@ from anchore_engine.db import Image, ImageCpe, ImagePackage
 from anchore_engine.services.policy_engine.engine.vulns.utils import get_api_endpoint
 from anchore_engine.subsys import logger as log
 from anchore_engine.util.rpm import split_fullversion
+from anchore_engine.utils import dedup_list_of_dicts
 
 
 class DistroMapper:
@@ -938,5 +939,10 @@ class EngineGrypeDBMapper:
                         "will_not_fix": grype_vulnerability.fix_state == "wont-fix",
                     }
                 )
+
+        for _, vulnerability in intermediate_tuple_list.items():
+            vulnerability["affected_packages"] = dedup_list_of_dicts(
+                vulnerability["affected_packages"]
+            )
 
         return list(intermediate_tuple_list.values())
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -923,10 +923,17 @@ class EngineGrypeDBMapper:
             if grype_vulnerability:
 
                 # Transform the versions block
-                if grype_vulnerability.deserialized_version_constraint:
-                    version = ",".join(
-                        grype_vulnerability.deserialized_version_constraint
-                    )
+                if grype_vulnerability.version_constraint:
+                    if "unspecified" in grype_vulnerability.version_constraint.lower():
+                        version = "*"
+                    else:
+                        version_strings = [
+                            version.strip(" '\"")
+                            for version in grype_vulnerability.version_constraint.split(
+                                "||"
+                            )
+                        ]
+                        version = ",".join(version_strings)
                 else:
                     version = "*"
 
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -923,16 +923,13 @@ class EngineGrypeDBMapper:
 
                 # Transform the versions block
                 if grype_vulnerability.version_constraint:
-                    if "unspecified" in grype_vulnerability.version_constraint.lower():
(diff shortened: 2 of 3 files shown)


anchore_engine/clients/grype_wrapper.py
```diff
@@ -50,6 +50,12 @@ def deserialized_related_vulnerabilities(self):
     def deserialized_fixed_in_versions(self):
         return json.loads(self.fixed_in_versions)
 
+    @property
+    def deserialized_version_constraint(self):
+        return [
```
nightfurys, line 55: version_constraint could be empty/blank. It means the vulnerability has not been fixed implying all versions of the package are vulnerable (for the matching process). We see this often in both os and non-os packages. For visual representation [legacy mode](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/engine/vulns/providers.py#L744) uses wild character `*`.   
I would recommend returning the version_constraint string as it is here and moving the parsing/interpreting logic to the usage. That could check if the version_constraint is blank and use a wild char  or parse the versions and populate them accordingly 

```
sqlite> SELECT count(*) from vulnerability where version_constraint='';
count(*)
82354

sqlite> SELECT namespace, count(*) from vulnerability where version_constraint='' group by namespace order by namespace;
namespace|count(*)
debian:10|3084
debian:11|1513
debian:12|1387
debian:9|4754
debian:unstable|1558
msrc:10379|56
msrc:10543|44
msrc:10729|59
msrc:10735|60
msrc:10788|49
msrc:10789|49
msrc:10816|58
msrc:10852|56
msrc:10853|57
msrc:10855|59
msrc:10951|16
msrc:10952|25
msrc:11453|14
msrc:11454|14
msrc:11466|23
msrc:11497|32
msrc:11498|35
msrc:11499|22
msrc:11563|17
msrc:11568|34
msrc:11569|36
msrc:11570|16
msrc:11571|31
msrc:11572|31
msrc:11583|14
msrc:11644|40
msrc:11645|42
msrc:11646|40
msrc:11647|36
msrc:11712|14
msrc:11713|16
msrc:11714|15
msrc:11715|8
msrc:11766|105
msrc:11767|104
msrc:11768|105
msrc:11769|100
msrc:products|37
nvd|13088
rhel:5|6195
rhel:6|4462
rhel:7|4360
rhel:8|1710
ubuntu:12.04|3351
ubuntu:12.10|1320
ubuntu:13.04|1023
ubuntu:14.04|3931
ubuntu:14.10|2207
ubuntu:15.04|2392
ubuntu:15.10|3005
ubuntu:16.04|737
ubuntu:16.10|2461
ubuntu:17.04|2643
ubuntu:17.10|2048
ubuntu:18.04|3860
ubuntu:18.10|1485
ubuntu:19.04|2311
ubuntu:19.10|1559
ubuntu:20.04|1857
ubuntu:20.10|1289
ubuntu:21.04|1195
```
```

Comment (Vijay-P) [reply anchore_engine/clients/grype_wrapper.py:55]: Hey swathi, the empty case is already handled by Dan's code in mappers.py line 890. If you want me to move things around I can, but the edge case is already covered.

**Review by dspalmer99 — COMMENTED** [review]
Logic looks good, ok by me pending resolution of @nightfurys ask. +1 on adding some extra null protection, out of an abundance of caution and to protect against upstream changes.

File: anchore_engine/clients/grype_wrapper.py (hunk 1 above)
Comment (nightfurys) [reply anchore_engine/clients/grype_wrapper.py:55]: Would be good to add null check to protect from unexpected db changes. Just to be clear, missing fix version is *not* an edge case. There are over 80K vuln records in the database

Comment (Vijay-P) [reply anchore_engine/clients/grype_wrapper.py]: As requested, I've moved all of the parsing to the caller (EngineGrypeDBMapper.to_engine_vulnerabilities()). I've also identified another case where affected version is "Unspecified". Also i've included some reusable dedup logic for the affected packages.

File: anchore_engine/utils.py [hunk 2]
```diff
@@ -858,3 +859,53 @@ def bytes_to_mb(value, round_to=None):
...
+
+    :param list_of_dicts: list of dictionaries to deduplicate
+    :type list_of_dicts: List[Dict[str, Any]]
+    :return: deduplicated list of dictionaries
+    :rtype: List[Dict[str, Any]]
+    """
+    unique_items = set(_recursive_freeze(child_dict) for child_dict in list_of_dicts)
+    return [
+        _recursive_unfreeze(child_frozenset) for child_frozenset in sorted(unique_items)
+    ]
```
nightfurys, line 878: This is a generic approach that works with any data shape. Are there any other solutions that might make it simpler and or more performant? One solution that comes to my mind is using named tuples as they are hashable and have faster creation (not read) times than dataclasses or objects. They might also be better suited for this data since it's not nested.  Granted, it's tied to the data shape and not as generic as the approach here. 

Pseudo-code for named tupel approach: create an output map/dict where the key is the named tuple and value is the affected package dictionary. Iterate over the input list and create a named tuple for each item, check if it's present in the map before adding it. Return the list-ified values of the output. 
```
AffectedPackageTuple = namedTuple("AffectedPackage", ["name", "type", "version", "wont_fix"])
output_map = dict()
for item in list_of_dicts:
  # create the named tuple using item
  # check if the named tuple is present in the output map and
  # add the named tuple-dict key-value pair to the output map
return list(output_map.values())
```

Vijay-P: tried some different things and went with simplified frozenset cast. this approach has the best average time for dedup. see latest commit

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -900,4 +907,15 @@ def to_engine_vulnerabilities(self, grype_vulnerabilities):
                     }
                 )
 
+        for _, vulnerability in intermediate_tuple_list.items():
+            unique_affected_packages = set(
+                [
```
nightfurys, line 912: (nit) is the list comprehension necessary here? set can operate over an iterable

Vijay-P: you are correct. i keep forgetting that i can just put the expression in there

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -922,10 +922,17 @@ def to_engine_vulnerabilities(self, grype_vulnerabilities):
             if grype_vulnerability:
 
                 # Transform the versions block
-                if grype_vulnerability.deserialized_fixed_in_versions:
-                    version = ",".join(
-                        grype_vulnerability.deserialized_fixed_in_versions
-                    )
+                if grype_vulnerability.version_constraint:
+                    if "unspecified" in grype_vulnerability.version_constraint.lower():
+                        version = "*"
```
nightfurys, line 927: @Vijay-P The inference behind using `*` is that all versions of the package are vulnerable i.e. the grype would match any version of this package to the vulnerability. I'm not aware of any such grype matching logic that special cases "Unspecified" version. It's possible I could be missing something. Otherwise the PR looks good to me

nightfurys: Sorry, I missed this before

Vijay-P: No problem. I only implemented this because i found data in the grype DB that follows this pattern. After taking a second look at the data with this query:
```
SELECT id, namespace, version_constraint, fixed_in_versions, fix_state FROM vulnerability WHERE vulnerability.version_constraint LIKE '%unspecified%' OR vulnerability.version_constraint LIKE '%unknown%';
```
it has become apparent that the "unspecified" check that I used cannot be broadly applied, as while most of them are not fixed, a couple of records have a fixed-in version.

PR #1246, opened 2021-09-23
Added active criteria in determining if grypedb being synced is new

zburstein:
We currently store 2 grype db meta records in the db at max. This is done so the old one is preserved as a fallback in the event that the new sync fails. The grypedb sync was failing when attempting to sync a grypedb that was in the GrypeDBFeedMetadata, but not the active record. This was because the logic used to determine if the grypedb currently being synced was new or not was not using the active criteria in determining if it was present. This change resolves that. There is a potentially more performant fix here that switches to the db already on the system rather than reprocessing it, but in the interest of time for upcoming release this resolves the issue


nightfurys (approved):
Neat!

PR #1248, opened 2021-09-25
Fix broken policy engine tests for grype provider

zburstein:
Fixes broken vulnerability tests for grype provider by adding the test images via the api instead of adding data to catalog and ingressing into the policy engine directly. Since the images need to analyze as opposed to having the content already available for vulnerability scanner, this changes does add some overhead to the workflow.


zburstein:
This failed because of the fixes made for alpine source matching and rpm epoch matching. I've updated the expected content with these changes now

PR #1249, opened 2021-09-27
Adds epoch to rpm metadata sbom generation for syft to fix version comparison for source pkg

zhill:
Adds an 'epoch' field to the package metadata in the SBoM being sent to Grype for scan. Without the epoch, Grype is incorrectly comparing the version fields and resulting in false-positives for already-fixed issues against RPMs.


zhill:
I have a unit test in my local branch that I'll push up here before merge.

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -170,6 +183,20 @@ def image_content_to_grype_sbom(self, record: Dict):
...
+            # Without this step Grype will use the wrong version string for the vulnerability match
+            full_version = record.get("version")
+            if full_version:
+                epoch_str, version, release = split_fullversion(full_version)
+                if epoch_str:
+                    try:
+                        artifact["metadata"]["epoch"] = int(epoch_str)
+                    except ValueError:
+                        # not a valid epoch, something went wrong
+                        pass
```
nightfurys, line 198: This epoch processing block will get executed only if a valid `sourcepkg` is found in the image content. Is that intended? I don't think those two things are related i.e. there may be cases where sourcepkg is invalid and epoch may still need to be processed

zhill: My read of: https://github.com/anchore/grype/blob/main/grype/matcher/rpmdb/matcher.go#L106 indicates it's only used for the sourcePkg match and when I modified the SboM to include it in the sourceRrpm field directly it also had the correct matching behavior, so I believe this is limited to the source package since main package version does contain the epoch. But to your point, I agree this block is safe to run regardless of source pkg presence. I'll update it.

nightfurys: Ah, I didn't realize the issue was with the epoch comparison in source package. After reading through the (Jira) ticket,  my only concern is that the epoch used here corresponds to the package, not source. It looks like the analyzers don't pick up the epoch from syft sbom. 

Not strictly necessary for the PR - ideally I would like the image content to be the source for this information, this allows the the user to view/access the image data that's being matched. rpm handler in the analyzer can be updated to preserve this value making it available to the catalog image content API. The mapper can then be updated to simply lookup that value

zhill: Agreed. Let's file an issue for that to track the change.

nightfurys (approved):
lgtm!

PR #1250, opened 2021-09-27
Changed sbom alpine metadata type to ApkMetadata

zburstein:
Changes sbom metadata type to `ApkMetadata` instead of `ApkgMetadata` because that is what grype uses to build the metadata information


nightfurys (approved):
ltgm, good catch Zane!

PR #1257, opened 2021-09-28
Fixes incorrect grypedb namespace record query

zhill:
Uses the metadata table instead of the constraint table to ensure the counts are per vulnerability, not per package constraint.


Vijay-P:
FYI before merge, this will conflict with changes in https://github.com/anchore/anchore-engine/pull/1258, so if it's merged before that PR then @zburstein will need to update #1258. Otherwise, f92607144b700ee4e76bd2f961e861c323a2b70a will need to be dropped from this branch.

PR #1258, opened 2021-09-28
Fix unsupported distro trigger when no grypedb is present

zburstein:
Changes unsupported distro trigger in grype to use `get_most_recent_active_grypedb` instead of using the grype wrapper to query the record source counts. This resolves a number of issues surfaced by @nightfurys  as well as fixes errors ocurring when grypedb is not fully synced

### DIFF
--- a/anchore_engine/services/policy_engine/engine/policy/gate_util_provider.py
+++ b/anchore_engine/services/policy_engine/engine/policy/gate_util_provider.py
@@ -2,7 +2,6 @@ import datetime
 from abc import ABC, abstractmethod
 from typing import Optional
 
-from anchore_engine.clients.grype_wrapper import GrypeWrapperSingleton
 from anchore_engine.db import DistroNamespace, session_scope
 from anchore_engine.db.db_grype_db_feed_metadata import (
     NoActiveGrypeDB,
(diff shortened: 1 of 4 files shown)


anchore_engine/services/policy_engine/engine/policy/gate_util_provider.py
```diff
@@ -119,9 +118,19 @@ def oldest_namespace_feed_sync(
...
+            try:
+                grypedb = get_most_recent_active_grypedb(session)
+            except NoActiveGrypeDB:
+                logger.info(
+                    "No vulnerabilities for image distro found because no active grype db found"
+                )
+                return False
+
+            groups = [
+                group["name"] for group in grypedb.groups if group["record_count"] > 0
```
nightfurys, line 131: Can grypedb.groups be an invalid reference? Might be worth a check before accessing it

zburstein: @nightfurys  Do you mean in the case that grypedb is potentially None? That should be handled by `get_most_recent_active_grypedb` which raises NoActiveGrypeDB if the query does not return any results. I can add that check for safety here as well though, but want to confirm that is what you mean by invalid reference.

zhill: 2 cases I can see: grypedb = None, or grypdb.groups = None.

zhill (approved):
lgtm

anchore_engine/services/policy_engine/engine/policy/gate_util_provider.py
zhill: I think yes, grypedb=None or grypedb.groups=None. If those are handled in the contract of that function call that should cover it.

zburstein: Oh i just pushed the changes to add that right before I saw this response @zhill

zburstein: `get_most_recent_active_grypedb` does not check if groups is None so the addition of that check here in this function is a good one

nightfurys: @zburstein I meant grypedb.groups != None should be checked. Was that check been added?

PR #1259, opened 2021-09-28
Adds support for vuln scanning binary and go content types with grype

zhill:
Adds cpe generation where needed in the sbom generation step prior to
handoff to Grype.

Also updates to latest Grype release that supports go-module vuln scans
and a default matcher that will handle 'binary' types.


tests/unit/anchore_engine/services/policy_engine/engine/vulns/test_mappers.py
```diff
@@ -152,427 +152,508 @@ def test_image_content_to_grype_metadata(self, input_metadata, expected_output):
...
-                    "name": "openssl-libs",
-                    "type": "rpm",
-                    "locations": [{"path": "pkgdb"}],
-                    "version": "1:1.1.1g-15.el8_3",
-                    "metadataType": "RpmdbMetadata",
-                    "metadata": {
-                        "sourceRpm": None,
-                        "epoch": 1,
-                    },
+    pass
```
nightfurys, line 155: Any reason why this was added and the tests are out of the class?

zhill: hmm, this doesn't look right. I'll fix this

zhill:
The cpe generation code is lifted from the policy engine loader. I moved it and cleaned it up a bit to ensure no weird import dependencies and since the call parameters were different. The only generator with any real complexity is the java generator which I only very mildly refactored (moved the tokenizer phase to its own function).

nightfurys (approved):
lgtm

PR #1260, opened 2021-09-29
Updates cpe generators to be fuzzy generally to support the lookup tables

zhill:
This is for name substitutions and

Updates java cpe generator to use the loaders.py version verbatim for
consistency with legacy scanner for hints input and pre-syft-integration analyzed images.


nightfurys (commented):
(nit) Are [these](https://github.com/anchore/anchore-engine/blob/87f7b2ea24c49e76512b3ffdb3185f7ff5444204/anchore_engine/util/cpe_generators.py#L115-L124) functions used anywhere besides testing? Can they be safely removed?

zhill:
Yes, they can be removed.

PR #1261, opened 2021-09-30
Disable default to store all sboms, and bump grype version

dspalmer99:
Disable default to store all sboms, and bump grype version


Vijay-P (approved):

dspalmer99:
Thanks @Vijay-P !

PR #1263, opened 2021-09-30
Updates to query vulnerabilities response with grype provider

nightfurys:
- Change the shape of references to be inline with legacy results - a list of objects where each object contains source and url attributes
- Construct a link to the vulnerability when one is not available
- Change the shape of cvss data to be inline with with legacy mode results
- Assign the cvss data to nvd_data or vendor_data depending on vulnerability namespace


dspalmer99 (approved):
LGTM, +1 on some unit tests if there is time.

nightfurys:
> LGTM, +1 on some unit tests if there is time.

Thanks @dspalmer99 pushed a few unit tests that verify the changes

PR #1266, opened 2021-10-04
Fix unused imports

Vijay-P:
This PR removes some unused imports in the policy engine and the db upgrade script. It's possible that the unused imports in the upgrade script were doing something, but I can't tell what. Would appreciate a sanity check.


Toure (approved):
LGTM, nice clean-up of the imports.

rbrady (approved):
Thanks for reducing tech debt.

PR #1284, opened 2021-10-19
Adds upgrade task to remove 'github:os' and 'github:unkown' entries

zburstein:
Removes 'github:os' and 'github:unkown' records from policy engine

### DIFF
--- a/anchore_engine/db/entities/upgrade.py
+++ b/anchore_engine/db/entities/upgrade.py
@@ -26,7 +26,6 @@ from anchore_engine.configuration.localconfig import (
 )
 from anchore_engine.db import (
     FeedGroupMetadata,
-    FixedArtifact,
     Image,
     ImageGem,
     ImageNpm,
@@ -35,7 +34,6 @@ from anchore_engine.db import (
     LegacyArchiveDocument,
     ObjectStorageMetadata,
     Vulnerability,
-    VulnerableArtifact,
     db_anchore,
     db_archivedocument,
     db_catalog_image,
@@ -1974,8 +1972,6 @@ def remove_incorrect_github_vuln_data():
             ImagePackageVulnerability,
             ImagePackageVulnerability.vulnerability_namespace_name,
         ),
-        (VulnerableArtifact, VulnerableArtifact.namespace_name),
-        (FixedArtifact, FixedArtifact.namespace_name),
         (Vulnerability, Vulnerability.namespace_name),
         (FeedGroupMetadata, FeedGroupMetadata.name),
     )
@@ -1990,6 +1986,7 @@ def remove_incorrect_github_vuln_data():
         try:
             with session_scope() as session:
                 results = session.query(type).filter(column.in_(namespace_names)).all()
+                log.err(f"Deleting {len(results)} records from {type.__tablename__}")
                 for result in results:
                     session.delete(result)
         except Exception as e:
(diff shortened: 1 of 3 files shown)


anchore_engine/db/entities/upgrade.py
```diff
@@ -1955,6 +1960,49 @@ def db_upgrade_013_014():
...
+        try:
+            engine.execute(
+                f"DELETE from {table_name} WHERE {col_name} IN {namespace_names}"
+            )
+        except Exception as e:
+            err_string = (
+                f"Failed to perform deletion on {table_name} - exception: {str(e)}"
+            )
+            log.err(err_string)
+            raise Exception(err_string)
```
nightfurys, line 1995: I would recommend using SQLAlchemy ORM to delete the records - refer to archive_data_upgrade_005_006 or other functions that use session object to perform db ops. Direct SQL execution should be sparingly used only when the functionality is not available via SQLAlchemy such as altering tables - adding, dropping, updating columns, creating indices etc. Using the ORM would also help with cleaning up foreign key relationships with cascading deletes (delete records from Vulnerability) and prevent typos in column names

tests/integration/db/upgrade/test_db_upgrade_015_016.py
```diff
@@ -0,0 +1,87 @@
...
+
+def add_records(records):
+    with session_scope() as session:
+        # Add some feed and group records
+        for record in records:
+            session.add(record)
+        session.flush()
+
+
+def test_remove_incorrect_github_vuln_data(anchore_db):
```
nightfurys, line 20: Good work on the unit tests, love it!

nightfurys (commented):
Looks good functionally. I just have a recommendation for db access

anchore_engine/db/entities/upgrade.py
```diff
@@ -1955,6 +1960,52 @@ def db_upgrade_013_014():
...
+    for type, column in to_delete:
+        log.err(
+            "Removing records from table {} where {} is {}".format(
+                type, column, namespace_names
+            ),
+        )
+
+        try:
+            with session_scope() as session:
+                results = session.query(type).filter(column.in_(namespace_names)).all()
```
nightfurys, line 1992: (nit) it would be nice to have log statement for how many records are getting deleted. It has to be at an error level to show up in the logs since it's executed prior to bootstrap. Something like 
```
log.err(f"Deleting {len(results)} records from {type.__tablename__}")
```

anchore_engine/db/entities/upgrade.py
```diff
@@ -1955,6 +1960,52 @@ def db_upgrade_013_014():
...
+
+    namespace_names = ["github:os", "github:unknown"]
+
+    to_delete = (
+        (
+            ImagePackageVulnerability,
+            ImagePackageVulnerability.vulnerability_namespace_name,
+        ),
+        (VulnerableArtifact, VulnerableArtifact.namespace_name),
+        (FixedArtifact, FixedArtifact.namespace_name),
```
nightfurys, line 1978: (nit) VulnerableArtifact and FixedArtifact deletion is not necessary. Vulnerability is related to those entities via a 1-many relationship. The [foreign key relationship](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/db/entities/policy_engine.py#L287-L292) also defines cascade operations. This would mean that deleting the parent Vulnerability record would delete all the related children in both tables. The end result is same regardless with a minor/insignificant performance gain as the data set is very small. IMO delegating cascade deletion to ORM is more precise and reduces the possibility of orphaning records.

nightfurys (approved):
lgtm. Minor suggestion around logging and cascade deletion

dspalmer99:
Thanks @nightfurys , I've addressed those comments.

PR #1288, opened 2021-10-25
use threadpoolexecutor in image archival

Vijay-P:
Related to: https://github.com/anchore/anchore-engine/pull/1273
This PR attempts to parallelize image archival with the use of a ThreadPoolExecutor. The hope that it may result in some performance improvements.

**Special notes**:

### DIFF
--- a/anchore_engine/services/catalog/api/controllers/archives.py
+++ b/anchore_engine/services/catalog/api/controllers/archives.py
@@ -2,8 +2,10 @@
 API controller for /archives routes
 
 """
+import concurrent
 import json
 import uuid
+from typing import List, Tuple
 
 from sqlalchemy import or_
 
@@ -330,12 +332,16 @@ def list_analysis_archive():
 
 
 @authorizer.requires_account(with_types=INTERNAL_SERVICE_ALLOWED)
-def archive_image_analysis(imageReferences):
+def archive_image_analysis(imageReferences: List[str]):
     """
     POST /archives/images
 
     body = [digest1, digest2, ... ]
 
+    :param imageReferences: list of image digests to archive
+    :type imageReferences: List[str]
+    :return: response body and status code
+    :rtype: Tuple[Any, int]
     """
 
     try:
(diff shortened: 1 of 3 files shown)


anchore_engine/services/catalog/api/controllers/archives.py
```diff
@@ -350,32 +356,58 @@ def archive_image_analysis(imageReferences):
...
-                        "detail": result_detail,
-                    }
-                )
-            except Exception as ex:
-                logger.exception(
-                    "Unexpected an uncaught exception from the archive task execution"
-                )
-                results.append({"digest": digest, "status": "error", "detail": str(ex)})
-
+        logger.info("Images to archive: %d", len(imageReferences))
```
zhill, line 359: I understand the reason for this change and it makes sense in terms of implementation, but it has a side-effect of enabling users to put quite a bit of load on the system in ways the operator might not expect. So, I'd like to see this stay serial so that user API requests cannot cause thread growth in unexpected ways for the admin. The thread workers should be limited to controlled async processes. If a user wants to speed this up they can send parallel requests instead of a single request and those will get queued/throttled by the api thread logic appropriately.

zhill (commented):
I know it's draft, but just providing some feedback now on the approach etc.

anchore_engine/services/catalog/api/controllers/archives.py
```diff
@@ -350,32 +356,58 @@ def archive_image_analysis(imageReferences):
...
+
+        with concurrent.futures.ThreadPoolExecutor(
```
zhill, line 361: I understand the reason for this change and it makes sense in terms of implementation, but it has a side-effect of enabling users to put quite a bit of load on the system in ways the operator might not expect. So, I'd like to see this stay serial so that user API requests cannot cause thread growth in unexpected ways for the admin. The thread workers should be limited to controlled async processes. If a user wants to speed this up they can send parallel requests instead of a single request and those will get queued/throttled by the api thread logic appropriately.

PR #1295, opened 2021-10-29
Merge v1.0.1 into master

zburstein:
Merge v1.0.1 into master


zburstein:
I believe that the PR is showing the rc_image_workflow because the head to this branch has rc tags. Just a note that it has failed because there exists a high severity vuln in engine that is a part of the base rhel image. It has been determined that the presence of the vuln is not actually an issue nor does it have a fix available. However, the rc_image_workflow runs a default policy bundle on the image which fails on the presence of any high severity vulns. This happens after the workflow has already tested and pushed the ec tagged image to DH

Btodhunter (approved):
lgtm

zhill (approved):
lgtm

PR #1302, opened 2021-11-04
Btodhunter/reformat dockerfile

Btodhunter:
The main purpose of the PR is to clean up our Dockerfile to make it easier to read. It also has some changes to the ordering and creation of layers to optimize build caching when building the image locally. 

By moving our COPY commands directly above the layers that are affected by changes to the local filesystem, we can utilize build caching to speed up image builds during development.

In the final stage, by consolidating to a single layer instead of multiple layers we can reduce the image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,10 @@ FROM ${BASE_REGISTRY}/${BASE_IMAGE}:${BASE_TAG} as anchore-engine-builder
 
 ARG CLI_COMMIT
 
-ENV LANG=en_US.UTF-8 LC_ALL=C.UTF-8
-ENV GOPATH=/go
+ENV LANG=en_US.UTF-8 
+ENV LC_ALL=C.UTF-8
+
+# environment variables for dependent binary versions
 ENV SYFT_VERSION=v0.26.0
 ENV GRYPE_VERSION=v0.22.0
 ENV PIP_VERSION=21.0.1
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,196 +1,249 @@
...
     yum install -y /build_output/build_deps/*.rpm && \
-    rm -rf /build_output /root/.cache
+    # Copy default application configuration files
+    echo "copying default application config files" && \
+    cp /build_output/LICENSE /licenses/ && \
+    cp /build_output/configs/default_config.yaml /config/config.yaml && \
+    md5sum /config/config.yaml > /config/build_installed && \
+    cp /build_output/configs/docker-entrypoint.sh /docker-entrypoint.sh && \
+    chmod +x /docker-entrypoint.sh && \
+    cp /build_output/configs/clamav/freshclam.conf /home/anchore/clamav/ && \
```
Vijay-P, line 236: This is my fault, but we could potentially install this the same way that we do in enterprise. Meaning, we wouldn't need to copy the configuration file over to the config directory since it's already inside the anchore engine wheel.

Vijay-P:
This LGTM, but let's chat about the freshclam file if you have some time.

subecho (approved):
Looks good to me! Thanks for cleaning this up, Brady! 👍

PR #1303, opened 2021-11-05
Adds Go and Binary Support For Semantic Version Comparison

rbrady:
A bug was reported where the go language was not supported
for version comparison.  This patch adds go and binary types
to the language version comparison checking.

closes: #ENT-870

Signed-off-by: Ryan Brady <ryan.brady@anchore.com>

**Special notes**:

### DIFF
--- a/anchore_engine/util/langpack.py
+++ b/anchore_engine/util/langpack.py
@@ -21,7 +21,7 @@ def language_compare(a, op, b, language="python"):
     if language in ["java", "maven"]:
         aoptions = [MavenVersion(a)]
         boptions = [MavenVersion(b)]
-    elif language in ["js", "npm", "ruby", "gem", "nuget"]:
+    elif language in ["js", "npm", "ruby", "gem", "nuget", "go", "binary"]:
         try:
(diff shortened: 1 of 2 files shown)


nightfurys (commented):
@rbrady the code change looks good for enabling matching for go and binary packages. Nice extension to the unit test as well! 

A 2nd issue in the ticket is error handling. Currently an exception is raised when a language matcher is not found. It's not handled and propagates up in the stack causing analysis to fail. This is not consistent with the vulnerability matching logic i.e missing/lack of matching logic should not cause the image analysis to fail. I would suggest raising a ValueError instead of an Exception and handling it in the calling function to log the error

rbrady:
@nightfurys thanks for the review and feedback.  I have added the prescribed ValueError with handling/logging and updated the test.

nightfurys (approved):
LGTM. Thanks for adding the exception handling!

nightfurys:
@rbrady Since this is targeted for 1.1.0 release, can you edit the base branch of the PR to v1.1.0-dev?

PR #1310, opened 2021-11-10
Initial implementation of unit test coverage prioritization tool

dakaneye:
We have had a hard time prioritizing how to improve our test coverage at Anchore, and that was a problem I had set out to try and make some progress on in this PR. 

At first, I was cataloging our test coverage in a spreadsheet (which was super painful), and then we decided to pivot and switch to building a script that can analyze the code and spit out a list of functions that do not have unit tests and sort the list of functions in order of how much they're used across the project.

I have a v1 script completed but it is a little bit of a frankenstein. It's not great at determining usages of methods across the application (it's just doing a file keyword search right now) or in the unit test directory.....but it does help a little!

I'll be using this to start making a dent in our unit test coverage and wanted to commit it to the project so that we can continue iterating on it!

### DIFF
--- a/scripts/tools/reporting/README.md
+++ b/scripts/tools/reporting/README.md
@@ -25,3 +25,9 @@ Working Directory: anchore-engine/scripts/tools/reporting
 
 Python 3.8
 Engine already installed and development environment set up
+
+
+## Improvement Notes
+- Ignore "virtual" functions in an ABC that cannot directly be called.
+- If a function doesn't belong to a class, it looks like it will just give you the directory that the function is in rather than the file (e.g. anchore_engine.clients::unpack)
+- Determine if ther'es any additional multithreading/processing we can implement that  would help lower the runtime of the script.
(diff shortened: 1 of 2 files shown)


subecho (approved):
This is a great little tool, thanks Sam! I think the only thing that I'd want to see before we merge is to add a note to the README that this will take about five or so minutes to finish running (I thought the script had hung for a while 😅 )

Some things that I'd like to see in future passes:
- Ignore "virtual" functions in an ABC that cannot directly be called.
- If a function doesn't belong to a class, it looks like it will just give you the directory that the function is in rather than the file (e.g. `anchore_engine.clients::unpack`) 
- I feel like some multithreading/processing would help lower the runtime of the script.

dakaneye:
Thanks for the review @subecho !

Agree with your points! I have multithreaded a bit of the execution (you should've seen it before haha). but in general the points are valid. I'll add them and merge!

dakaneye:
@DaneWeber this is ready to merge for now, but not sure where you all are at regarding release timing, will let you do the actual merging! This should NOT affect any runtime engine code

PR #1313, opened 2021-11-12
Add nvd data to query vuln

zburstein:
Adds nvd data to response of query vulnerabilities for grype provider

### DIFF
--- a/anchore_engine/services/policy_engine/engine/vulns/mappers.py
+++ b/anchore_engine/services/policy_engine/engine/vulns/mappers.py
@@ -4,6 +4,7 @@ import re
 import uuid
 from typing import Dict, List, Optional
 
+from anchore_engine.clients.grype_wrapper import GrypeVulnerabilityMetadata
 from anchore_engine.common import nonos_package_types
 from anchore_engine.common.models.policy_engine import (
     CVSS,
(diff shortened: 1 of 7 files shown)


dspalmer99 (approved):
LGTM

anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -929,6 +930,91 @@ def _transform_cvss(self, vulnerability_id: str, cvss: Dict) -> Dict:
...
+
+                    if nvd_metadata_record:
+                        related_vuln_nvd_data = self._cvss_from_grype_raw_result(
+                            nvd_metadata_record
+                        )
+                        nvd_cvss_cache[nvd_metadata_record.id] = related_vuln_nvd_data
+                        nvd_data += related_vuln_nvd_data
+                    else:
+                        log.debug("Missing related record for %s", related_vuln["ID"])
+
```
nightfurys, line 1015: This seems a little complicated. Suggestions for simplification:
- build a map of nvd-id->cvss score map prior to this call and use that for look ups. It might be more performant/efficient than the cache built here
- the logic is a bit convoluted for nvd and non-nvd vulnerabilities, that's another thing that could be handled in a higher order layer

nightfurys (commented):
Good job on the changes and tests! Chatted with @zburstein, will follow up with a commit that addresses some of the issues

PR #1314, opened 2021-11-15
Fix analyzer log messages with wrong format

zhill:
Incorrect log message format use in the analyzer manager. Not a functional issue but creates logging confusion due to logging the error.


luhring (approved):

dspalmer99 (approved):

Toure (approved):
Looks good, nice clean-up that will remove one more warning from the linter.

PR #1315, opened 2021-11-15
Support for Rocky Linux distro

nightfurys:
This PR adds mappings to recognize Rocky Linux distro for both grype and legacy vulnerabilities provider


dspalmer99 (approved):
LGTM

PR #1316, opened 2021-11-16
Tools update for Rocky Linux support

nightfurys:
- Bump Syft to v0.30.1
- Bump Grype to v0.25.1
- Add schema section in policy-engine generated sbom, necessary due to recently introduced Syft sbom validation


anchore_engine/services/policy_engine/engine/vulns/mappers.py
```diff
@@ -769,6 +769,13 @@ def image_content_to_grype_sbom(image: Image, image_content_map: Dict) -> Dict:
     # initialize the sbom
     sbom = dict()
 
+    # set the schema version, this whole fn could be refactored as a versioned transformer
+    # not refactoring since use of source sbom will make this transformation moot
+    sbom["schema"] = {
```
dspalmer99, line 774: @nightfurys This is to keep the engine-generated sboms consistent with the schema changes in syft that grype expects?

nightfurys: Syft recently introduced sbom validation on schema section. I think it was always a part of the syft schema but it was left out as an optional element. Without this newer versions of grype fail with validation error

dspalmer99: Ack, thanks

dspalmer99 (approved):
LGTM

PR #1318, opened 2021-11-17
Add Fix for `get_input_string` when a Regsitry is Specified with a Port

subecho:
Previously, we would make the improper assumption in the `get_input_string` function within `catalog_impl.py` that the `split()` function on the image_key's tag field would always return two items: the registry and image name and then the tag for the image. However when a port number was explicitly used in the registry, this caused the
`split()` function to return three things: the registry name, port number and image name, and then the image tag.

This fix changes the `split()` function to `rsplit()` and will only split the string once, ensuring that the first item is always the registry (with port if specified) and image name, and the second is the tag for the image.

New unit test parameters were added to `test_get_input_string` to ensure that the updated function still works as intended and that the observed bad behavior has been fixed.

Signed-off-by: Dustin Schoenbrun <dustin.schoenbrun@anchore.com>

### DIFF
--- a/anchore_engine/services/catalog/catalog_impl.py
+++ b/anchore_engine/services/catalog/catalog_impl.py
@@ -1870,7 +1870,9 @@ def get_input_string(image_key: ImageKey) -> str:
         if image_key.digest == "unknown":
             return image_key.tag
         else:
-            return "{}@{}".format(image_key.tag.split(":")[0], image_key.digest)
+            return "{}@{}".format(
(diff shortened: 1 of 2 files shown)


zhill (approved):
lgtm! Nice work!

nightfurys (approved):
Nice work with the unit test!

subecho:
> Nice work with the unit test!

Thanks! Lots of help from Zach with getting it just right 👍

PR #1322, opened 2021-11-27
Add unit test for syft handlers save_entry

dakaneye:

**What this PR does / why we need it**: These methods are highly used across engine and don't have unit tests currently, so let's add some!

### DIFF
--- a/tests/unit/anchore_engine/analyzers/syft/handlers/test_alpine.py
+++ b/tests/unit/anchore_engine/analyzers/syft/handlers/test_alpine.py
@@ -1,0 +1,42 @@
+import pytest
+
+from anchore_engine.analyzers.syft.handlers.alpine import save_entry
+
+
+class TestAlpine:
+    @pytest.mark.parametrize(
+        "param",
+        [
+            pytest.param(
+                {
+                    "findings": {},
+                    "engine_entry": {"name": "test"},
+                    "pkg_key": "basic-test",
+                    "expected": {"name": "test"},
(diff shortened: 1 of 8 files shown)


emily-lopez (commented):
Overall looks good

Should we avoid using the same code/logic the function uses in the unit test itself? Instead we could use another _expected_ parameter and verify the correct pkg_key is generated as part of the test? 

Not super critical for test like alpine & debian that are pretty simple, but for tests like gem and java there are some more complex cases to generate the pkg_key if not provided. 

```
   pytest.param(
   {
         "findings": {},
         "engine_entry": {"name": "test", "latest": "1.0.1"},
         "pkg_key": None,
         "expected": {"name": "test", "latest": "1.0.1"},
         "expected_key": "/virtual/gempkg/test-1.0.1"
   },
id="no-pkgkey-no-location-no-version-case",

...

    def test_save_entry(self, param):
        findings = param["findings"]
        save_entry(findings, param["engine_entry"], param["pkg_key"])
        findings_key = param.get("expected_key", None)
        assert (
            findings.get("package_list", {})
            .get("pkgs.gems", {})
            .get("base", {})
            .get(findings_key, {})
            == param["expected"]
        )
```

dakaneye:
@emily-lopez great feedback! FYI the broken unit tests are externally related, and fixed in https://github.com/anchore/anchore-engine/pull/1365

emily-lopez (approved):
LGTM

PR #1325, opened 2021-11-30
add unit tests for CPE method

dakaneye:

**What this PR does / why we need it**: Adds some unit tests for the CPE class most used methods

### DIFF
--- a/tests/unit/anchore_engine/test_utils.py
+++ b/tests/unit/anchore_engine/test_utils.py
@@ -7,6 +7,7 @@ from anchore_engine.utils import (
     run_check,
     run_command_list_with_piped_input,
     run_sanitize,
+    CPE,
 )
 
 images = [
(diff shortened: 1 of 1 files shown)


emily-lopez (changes requested):
Overall looks good to me. 

For bind_for_cpe22_URI, parameter unfound-encoding-endofelement, should this just be "\\" ? Looking at the function, it looks like this case covers the "\\" being the last element? Correct me if my understanding is not right.

dakaneye:
> Overall looks good to me.
> 
> For bind_for_cpe22_URI, parameter unfound-encoding-endofelement, should this just be "\" ? Looking at the function, it looks like this case covers the "\" being the last element? Correct me if my understanding is not right.

Hi @emily-lopez ! Thanks for reviewing. I see how this can be a little confusing based on the code. I'm actually testing an inner block of that part when "\\" is the **current** character, and the next character is not found in the list of the `CPE_SPECIAL_CHAR_ENCODER` dict. Currently, the "." character is not in that dict's keys so it forces the unit test to enter the else block in this if statement: 

```
if encoded:
    result += encoded
else:  # no encoding found, let it go through as it is
    logger.warn(
        "No encoding found for {}{}".format(char, n_char)
    )
    result += char + n_char
```

I double checked my understanding here as well to make sure, but I think it looks correct as is.

emily-lopez (approved):
Got it, make sense. Everything looks good.

PR #1332, opened 2021-12-04
add unit tests for file system object storage driver

dakaneye:

**What this PR does / why we need it**: Adds a unit test for the file system object storage driver

**Special notes**:

### DIFF
--- a/anchore_engine/subsys/object_store/drivers/s3.py
+++ b/anchore_engine/subsys/object_store/drivers/s3.py
@@ -36,6 +36,11 @@ class S3ObjectStorageDriver(ObjectStorageDriver):
         self.s3_client = None
         self.session = None
 
+        if "unittest" in self.config:
+            self.prefix = self.config.get("prefix", "")
+            self.bucket_name = self.config.get("bucket")
+            return
+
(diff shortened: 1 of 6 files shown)


emily-lopez (commented):
Overall looks good. 

For test_filesystem - test_uri_for, should the param also include an expected key with the path instead of using the driver function, similar to the other tests in rdbms, s3?  Otherwise, it's calling the same function the function under test uses. Or is similar to test_get_archive_filepath where expected values should be created dynamically ?

dakaneye:
Hi @emily-lopez thanks for the comment! For this one I just had figured that since the archive file path includes a hex digest and an md5 hash, I'd basically just be re-writing the underlying method in order to provide a realistic input to the `uri_for` method. 

I can definitely compute and add a string as an input instead if that's what you mean, just let me know

emily-lopez:
That make sense. After reviewing the uri_for function, it looks like the test just needs to verify that the return of  get_archive_filepath is being formatted correctly. Since that function has its own dedicated tests, I think we can keep the test as is.

emily-lopez (approved):

PR #1333, opened 2021-12-04
Add unit test for FilesystemAnalysis database object files json method

dakaneye:

**What this PR does / why we need it**: This method is called ~16 times across engine and isn't tested, so I've added a unit test for it now


emily-lopez (approved):
LGTM

PR #1336, opened 2021-12-07
Adds save and retrieve image sboms

zhill:
Saves raw syft analysis output and makes it available from GET /images/<digest>/sboms/native.

Still needs to add analysis archive/restore logic to include this data.

### DIFF
--- a/anchore_engine/analyzers/syft/__init__.py
+++ b/anchore_engine/analyzers/syft/__init__.py
@@ -1,90 +1,25 @@
-import collections
+import typing
 
 from anchore_engine.analyzers.syft.handlers import (
     modules_by_artifact_type,
     modules_by_engine_type,
 )
-from anchore_engine.analyzers.utils import defaultdict_to_dict, dig
 from anchore_engine.clients.syft_wrapper import run_syft
-from anchore_engine.subsys import logger
+
+from .adapters import FilteringEngineAdapter
 
 
-def filter_relationships(relationships, **kwargs):
-    def filter_fn(relationship):
-        for key, expected in kwargs.items():
-            if relationship[key] != expected:
-                return False
-        return True
-
-    return [r for r in relationships if filter_fn(r)]
-
-
-def filter_artifacts(artifacts, relationships):
-    def filter_fn(artifact):
-        # some packages are owned by other packages (e.g. a python package that was installed
-        # from an RPM instead of with pip), filter out any packages that are not "root" packages.
-        if filter_relationships(
-            relationships, child=dig(artifact, "id"), type="ownership-by-file-overlap"
-        ):
-            return False
-
-        return True
-
-    return [a for a in artifacts if filter_fn(a)]
-
-
-def catalog_image(unpackdir, imagedir, package_filtering_enabled=True):
+def catalog_image(
+    tmp_dir: str, image_oci_dir: str, package_filtering_enabled=True
+) -> typing.Tuple[dict, dict]:
     """
-    Catalog the given image with syft, keeping only select artifacts in the returned results.
+    Catalog the given image with syft, keeping only select artifacts in the returned results
+
+    :param tmp_dir: path to directory where the image data resides
+    :param image_oci_dir: path to the directory for temp file construction
+    :return: tuple of engine formatted result and raw syft output to allow it to be used downstream if needed
     """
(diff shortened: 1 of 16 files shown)


zhill:
The FilteringAdapter and the adapters.py module is just a movement of code from elsewhere, not a change in logic for format of the actual artifacts. My intent was to create a place we can start swapping or enhancing implementations for those mappings as we move from the legacy analysis report and start relying on the full advanced features of syft itself.

anchore_engine/services/apiext/api/controllers/images.py
```diff
@@ -1570,3 +1573,42 @@ def list_secret_search_results(imageDigest):
...
+    account = ApiRequestContextProxy.namespace()
+    try:
+        _get_image_ok(account, imageDigest)
+
+        client = internal_client_for(CatalogClient, account)
+        resp = client.get_document("syft_sbom", imageDigest)
+
+        # Do
+        compressed_bytes = json_content_gzipper(resp)
+        resp = flask.Response(
```
Vijay-P, line 1596: Can you set the name in the headers like so: https://github.com/anchore/enterprise/blob/960db726453d72b0043dba7bc48de8d2489d3fd6/anchore_enterprise/services/api/controllers/sources.py#L113

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -1116,6 +1116,30 @@ paths:
...
+        - application/gzip
+      parameters:
+        - name: imageDigest
+          in: path
+          type: string
+          required: true
+        - $ref: "#/parameters/AsAccountParameter"
+      responses:
+        200:
+          description: Image lookup success
```
Vijay-P, line 1137:
```
         schema:
            type: file
```

anchore_engine/services/apiext/api/controllers/images.py
zhill: Yes I can!

zhill: I will also move this into the catalog api in a follow-up PR so that the API controller isn't hitting the object store directly.

anchore_engine/services/apiext/swagger/swagger.yaml
```diff
@@ -5132,4 +5164,6 @@ definitions:
         additionalProperties: true
         type: object
     additionalProperties: true
+    type: object
+  NativeSbom:
```
Vijay-P, line 5168: i think you can get rid of this? appears unused

anchore_engine/analyzers/syft/__init__.py
```diff
@@ -1,90 +1,25 @@
...
 
 from anchore_engine.analyzers.syft.handlers import (
     modules_by_artifact_type,
     modules_by_engine_type,
 )
-from anchore_engine.analyzers.utils import defaultdict_to_dict, dig
 from anchore_engine.clients.syft_wrapper import run_syft
-from anchore_engine.subsys import logger
 
+from .adapters import FilteringEngineAdapter
```
Vijay-P, line 9: please use absolute path for this import

anchore_engine/analyzers/syft/adapters.py
```diff
@@ -0,0 +1,122 @@
+"""
+Module for adapters that convert syft to other formats
+"""
+import collections
+
+from anchore_engine.analyzers.utils import defaultdict_to_dict, dig
+from anchore_engine.subsys import logger
+
+from .handlers import modules_by_artifact_type
```
Vijay-P, line 9: please use absolute path

Vijay-P:
This looks great! A few nits about imports. Note that adapters refactor will break the source analyzer in enterprise, so we will need a follow-up PR to change import paths in enterprise.

anchore_engine/services/apiext/api/controllers/images.py
Vijay-P: i'm doing the same thing in enterprise, so no complaints for now, haha. i'll make sure to make the same change

Toure (approved):
Looks good, I like the fact that __init__ is getting thinner and overall general cleanup.

Vijay-P (approved):

PR #1352, opened 2021-12-24
Add unit test for the write_kvfile_fomdict method

dakaneye:

**What this PR does / why we need it**: Adds a unit test for the write_kvfile_fromdict method

**Special notes**:

### DIFF
--- a/tests/unit/anchore_engine/analyzers/test_utils.py
+++ b/tests/unit/anchore_engine/analyzers/test_utils.py
@@ -81,11 +81,19 @@ class TestDig:
         ),
         pytest.param(
             {
-                "filename": "./nonstring-key-test.json",
-                "dict": {2: ""},
+                "filename": "./nonstring-value-test.json",
+                "dict": {"two": 2},
                 "expected": TypeError(
-                    "Expected value of key 2 to be a string, found int"
+                    "Expected value of key two to be a string, found int"
                 ),
+            },
+            id="nonstring-value",
+        ),
+        pytest.param(
+            {
+                "filename": "./nonstring-key-test.json",
+                "dict": {2: "two"},
+                "expected": TypeError("expected string or bytes-like object"),
             },
             id="nonstring-key",
         ),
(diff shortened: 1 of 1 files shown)


emily-lopez (commented):
For case non-string key:
This code is checking the value should be string type versus the key type being string. 
```
 # Expects the value of each key in the dict to be a string
 type(dict[k])
```
For cleank it also looks like its expecting the key be a string as well
```
cleank = re.sub(r"\s+", "____", k)
```
This fixture address the second case, where key is non-string, though the expected error message mentions value vs key is incorrect type.  Should we add a case to cover value being non string? Does the code need to handle case for non-string key before cleaning? 

Everything else look good.

dakaneye:
> For case non-string key: This code is checking the value should be string type versus the key type being string.
> 
> ```
>  # Expects the value of each key in the dict to be a string
>  type(dict[k])
> ```
Hi @emily-lopez I'm a little confused about what you mean here! Feel free to double check my understanding because I haven't written anything in a month, but my interpretation of this line in the actual method (not the unit test):
```
if type(dict[k]) != str:
    raise TypeError(
```
Is that the `dict[k]` expression is actually accessing the _value_ of the key, not the key itself. So the conditional is really asking if the value of that key within the dict is a string or not. Therefore, my interpretation is that the unit test is correct for this part.

> For cleank it also looks like its expecting the key be a string as well
>
> cleank = re.sub(r"\s+", "____", k)

This section is a little confusing as well. The sub method does say that the 3rd argument _should_ be a string, but in actual execution of the test, no error was returned, so I assume it is able to do some formatting of the integer into the string. I can add a comment for this part to clarify!

Let me know also if you'd like to chat in slack or something to make sure we're on the same page

dakaneye:
@emily-lopez fixed the issues you called upon! Thanks for pointing that out. By the way, the current unit tests failures are an external issue, which I've also fixed in: https://github.com/anchore/anchore-engine/pull/1365

emily-lopez (approved):
Changes look good.

PR #1354, opened 2022-01-11
Adds events gc monitor function in catalog. Fixes #335

### LINKED ISSUE #335: Feature: Automatically delete events older than X days
As a system admin, I want to be able to set a max age for events to be retained and any event in the system older than that age will be automatically flushed from the db so that my db doesn't grow in size forever or until I manually flush events in all accounts.

zhill:
Adds new services.catalog.event_log.max_retention_age_days config value
to control the number of days that event records are kept. A value >0 is
used for a gc process to cleanse old records. The default is 0 so there
is no upgrade behavior change for existing users.

### DIFF
--- a/anchore_engine/db/db_events.py
+++ b/anchore_engine/db/db_events.py
@@ -151,6 +151,30 @@ def delete_byfilter(userId, session=None, since=None, before=None, **dbfilter):
     return ret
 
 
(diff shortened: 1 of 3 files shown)


natefive:
Hi @zhill, why was this PR closed without merging?

PR #1359, opened 2022-01-25
add libpq-devel to Dockerfile.functional so that psycopg2 can be compiled on ARM

Vijay-P:
There is no precompiled binary for psycopg2-binary on ARM. When running the functional tests on an ARM machine, psycopg2 fails to compile because pg_config is not available. This PR adds libpq-devel to the Dockerfile for the job runner so that psycopg2 may be compiled from source.

### DIFF
--- a/scripts/ci/Dockerfile.functional
+++ b/scripts/ci/Dockerfile.functional
@@ -5,7 +5,7 @@ USER root
 
 RUN set -ex && \
     echo "installing OS dependencies" && \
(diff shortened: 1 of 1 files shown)


subecho (approved):
Saw it work on Zoom with Weston and Vijay and we were able to successfully build and run the functional tests on both x86_64 and ARM64. Thanks for helping us out, Vijay! 👍

westonsteimel (approved):

PR #1361, opened 2022-02-01
integrate syft go-module detection with engine 

westonsteimel:
The version of syft bundled with anchore-engine already supports extracting Go module dependencies from go binaries; however, this information is currently filtered out by anchore-engine.  This PR provides the necessary mappings to support go-modules within anchore-engine.


westonsteimel:
We might also be able to remove https://github.com/anchore/anchore-engine/blob/master/anchore_engine/analyzers/modules/32_golang_packages.py, but I wasn't completely sure

subecho (approved):
Looks good to me! Thanks, Weston! 👍

subecho:
> We might also be able to remove https://github.com/anchore/anchore-engine/blob/master/anchore_engine/analyzers/modules/32_golang_packages.py, but I wasn't completely sure

We can try to remove it in another PR and see if anything breaks!

PR #1365, opened 2022-02-14
Change msg to reason within the skip annotation

dakaneye:
Noticed that some old unit tests were causing issues in pytest, due to the fact that the "msg" keyword was used instead of the "reason" keyword. Maybe this was a new update or something?


emily-lopez (commented):
LGTM

Vijay-P (approved):
already made the same changes in a different branch, LGTM
go ahead and merge

PR #1366, opened 2022-02-14
Add unit test cases for the make_anchore_exception method

dakaneye:
This method is called numerous times across the application, and is untested. Adding tests here!


emily-lopez (approved):
LGTM

PR #1367, opened 2022-02-16
devops-138: gate pushing prod on test-container

hunguyen1:
Signed-off-by: Hung Nguyen <hung.nguyen@anchore.com>

- Remove test-container dependency on prod push (using latest image)
- Pulls the RC candidate image to be pushed to prod to test the container to make sure it contains all expected binaries before pushing to prod

**Special notes**:


.circleci/config.yml
```diff
@@ -538,6 +538,11 @@ workflows:
           py_version: "3.8"
           context: e2e-testing
           filters: *filter_semver_tags
+      - container_tests:
```
Btodhunter, line 541: I think we should probably just pull `containter_tests` out of the prod workflow and stick it into the RC workflow as a gate to the 'push_image' job. The prod image workflow just retags the last RC image, so this will provide us the tests we need early on in the release pipeline.

Btodhunter (changes requested):

.circleci/config.yml
hunguyen1: sounds good. i've removed test-container-prod. Thanks

Makefile
```diff
@@ -164,11 +164,6 @@ setup-and-test-cli: setup-test-infra venv ## Set up and run end to end tests
 test-container-dev: setup-test-infra venv ## CI ONLY Run container-structure-tests on locally built image
```
Btodhunter, line 164: nit: I think this could just be `test-container`

Btodhunter (approved):
lgtm

Makefile
hunguyen1: sounds good. updated. thanks!

Btodhunter (approved):
lgtm

PR #1376, opened 2022-04-28
Update readme

zhill:
Clearer distinction with Syft, Grype and Engine.


nwl:
Can I suggest adding text which points people to the 1.0 blog after mentioning Syft/Grype so new users can see the plan.

PR #1392, opened 2022-09-28
Bump Syft/Grype dependencies

brennoo:
**What this PR does / why we need it**: Bump Syft, Grype dependencies. I understand that there is no active development on anchore-engine but these dependencies need to get updated to address issues that are affecting anchore-engine.

**Which issue this PR fixes** : some false positives/negatives, examples: https://github.com/anchore/grype/issues/504 https://github.com/anchore/grype/pull/917 as well other recent improvements they received.

**Special notes**:


tuxerrante:
Grype is already at 0.55.0
https://github.com/anchore/grype/releases/tag/v0.55.0

Dockerfile
```diff
@@ -12,8 +12,8 @@ ENV LANG=en_US.UTF-8
 ENV LC_ALL=C.UTF-8
 
 # environment variables for dependent binary versions
-ENV SYFT_VERSION=v0.33.0
-ENV GRYPE_VERSION=v0.27.3
+ENV SYFT_VERSION=v0.57.0
+ENV GRYPE_VERSION=v0.50.2
```
tuxerrante, line 16: This could be already bumped to 0.55
```suggestion
ENV GRYPE_VERSION=v0.55.0
```

